PR #5, opened 2019-02-01
Initial support for YUV4MPEG2 (.y4m) input files

mafonsoN:
Support for YUV4MPEG2 (.y4m) input files, all bit depths, standard input mode.

Reads the y4m header and saves the video's parameters (width, height, bitdepth, etc) to the config structure, so there is not need to specify them explicitly. 

In addition, when reading each frame, a delimited "FRAME\n" is removed.

More info about the YUV4MPEG2 format can be found here: https://wiki.multimedia.cx/index.php/YUV4MPEG2

### DIFF
--- a/Source/App/EbAppInputy4m.c
+++ b/Source/App/EbAppInputy4m.c
@@ -1,5 +1,5 @@
 /*
-* Copyright(c) 2018 Intel Corporation
+* Copyright(c) 2019 Intel Corporation
 * SPDX - License - Identifier: BSD - 2 - Clause - Patent
 */
 
@@ -226,15 +226,20 @@ int32_t readY4mFrameDelimiter(EbConfig_t *cfg){
 }
 
 /* check if filename contains y4m extension */
-EbBool checkIfY4m(const char* inputFileName){
+EbBool checkIfY4m(FILE* inputFile){
 
-    size_t len;
+    unsigned char buffer[YFM_HEADER_MAX+1];
 
-    len = strlen(inputFileName);
-    if(strncmp(&inputFileName[len-4], ".y4m", 4)==0){
-        return EB_TRUE; /* .y4m file extension */
+    /* find out if this is actually a .y4m file*/
+    fread(buffer, YUV4MPEG2_IND_SIZE, 1, inputFile);
+    buffer[YUV4MPEG2_IND_SIZE] = 0;
+
+    fseek(inputFile, 0, SEEK_SET);
+
+    if (strcmp((const char*)buffer, "YUV4MPEG2") == 0) {
+        return EB_TRUE; /* YUV4MPEG2 file */
     }else{
-        return EB_FALSE; /* not .y4m file extension */
+        return EB_FALSE; /* Not a YUV4MPEG2 file */
     }
 
 }
--- a/Source/App/EbAppInputy4m.c
+++ b/Source/App/EbAppInputy4m.c
@@ -218,18 +218,22 @@ int32_t readY4mFrameDelimiter(EbConfig_t *cfg){
 }
 
 /* check if the input file is in YUV4MPEG2 (y4m) format */
-EbBool checkIfY4m(FILE* inputFile){
+EbBool checkIfY4m(EbConfig_t *cfg){
 
-    unsigned char buffer[YFM_HEADER_MAX+1];
+    unsigned char buffer[YUV4MPEG2_IND_SIZE+1];
 
     /* Parse the header for the "YUV4MPEG2" string */
-    fread(buffer, YUV4MPEG2_IND_SIZE, 1, inputFile);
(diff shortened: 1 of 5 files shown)


kylophone:
Ah, nice. I would prefer something like a `EbDemuxContext` that can demux other formats too. This is probably fine for now though. I will see how this y4m reader fairs with all our AWCY input files.

gnafuthegreat:
When piping Y4M data from ffmpeg (-f yuv4mpegpipe), it looks like I still need to specify width and height.

Source/App/EbAppInputy4m.c
```diff
@@ -0,0 +1,240 @@
...
+
+}
+
+/* check if filename contains y4m extension */
+EbBool checkIfY4m(const char* inputFileName){
+
+    size_t len;
+
+    len = strlen(inputFileName);
+    if(strncmp(&inputFileName[len-4], ".y4m", 4)==0){
```
Kagami: Out of range read if `len < 4`

mafonsoN:
> When piping Y4M data from ffmpeg (-f yuv4mpegpipe), it looks like I still need to specify width and height.
> 
> EDIT: Presumably it's not even recognizing it as Y4M in that situation. If it wouldn't be too much trouble, it would be nice if it detected Y4M from stdin.

Reading y4m from stdin is not supported yet, only from the specified input file (-i). I'll add it.

kylophone:
> EDIT: Presumably it's not even recognizing it as Y4M in that situation. If it wouldn't be too much trouble, it would be nice if it detected Y4M from stdin.

Yep. A better way of detecting y4m would probably just be to probe the first 10 bytes of the file and `strsmp()` with `YUV4MPEG2` .

gnafuthegreat:
> 
> 
> > When piping Y4M data from ffmpeg (-f yuv4mpegpipe), it looks like I still need to specify width and height.
> > EDIT: Presumably it's not even recognizing it as Y4M in that situation. If it wouldn't be too much trouble, it would be nice if it detected Y4M from stdin.
> 
> Reading y4m from stdin is not supported yet, only from the specified input file (-i). I'll add it.

Thanks!  My typical workflow is to use ffmpeg to pipe an otherwise compressed source video so I'm not eating up a ton of disk space.

Source/App/EbAppInputy4m.c
mafonsoN: Thanks for the input

mafonsoN:
> OOB

Done

mafonsoN:
> Yep. A better way of detecting y4m would probably just be to probe the first 10 bytes of the file and `strcmp()` with `YUV4MPEG2` .
> 
> EDIT: This would also avoid the OOB problem @Kagami pointed out.

Done

Source/App/EbAppInputy4m.c
```diff
@@ -0,0 +1,245 @@
...
+/* check if filename contains y4m extension */
+EbBool checkIfY4m(FILE* inputFile){
+
+    unsigned char buffer[YFM_HEADER_MAX+1];
+
+    /* find out if this is actually a .y4m file*/
+    fread(buffer, YUV4MPEG2_IND_SIZE, 1, inputFile);
+    buffer[YUV4MPEG2_IND_SIZE] = 0;
+
+    fseek(inputFile, 0, SEEK_SET);
```
Kagami: I don't think it's possible to seek inside a pipe (e.g. when passing y4m from ffmpeg)?

Source/App/EbAppInputy4m.c
```diff
@@ -0,0 +1,245 @@
...
+    if (strcmp((const char*)bufferY4Mheader, "FRAME\n") != 0) {
+        fprintf(cfg->errorLogFile, "Failed to read propor y4m frame delimeter. Read broken.\n");
+        return EB_ErrorBadParameter;
+    }
+
+    return EB_ErrorNone;
+
+}
+
+/* check if filename contains y4m extension */
```
Kagami: Also comment should be fixed.

mafonsoN: Done

Source/App/EbAppInputy4m.c
mafonsoN: Thanks. Modified this.

mafonsoN:
> > > When piping Y4M data from ffmpeg (-f yuv4mpegpipe), it looks like I still need to specify width and height.
> > > EDIT: Presumably it's not even recognizing it as Y4M in that situation. If it wouldn't be too much trouble, it would be nice if it detected Y4M from stdin.
> > 
> > 
> > Reading y4m from stdin is not supported yet, only from the specified input file (-i). I'll add it.
> 
> Thanks! My typical workflow is to use ffmpeg to pipe an otherwise compressed source video so I'm not eating up a ton of disk space.

Reading from stdin should be working now (-i stdin).

PR #8, opened 2019-02-02
Build.sh improvements

tux3:
Note that build.sh was using 8-wide hard tabs, even though STYLE.md wants 4 spaces.
I assumed this was on purpose and kept it as is, can update if needed.

PR #27, opened 2019-02-06
Move the bash shebang to the top of build.sh so that it is executable.

xorgy:
The `#!` needs to be on the first line for the script to be executable.


xorgy:
I noticed #8 does this, but there's a lot more to that PR which is unrelated.

PR #32, opened 2019-02-06
Fix compile warnings in y4m header parsing

Spacefish:
Replace warnings by error checking via assertions

### DIFF
--- a/Source/App/EbAppInputy4m.c
+++ b/Source/App/EbAppInputy4m.c
@@ -14,6 +14,7 @@ int32_t readY4mHeader(EbConfig_t *cfg){
     FILE *ptr_in;
     unsigned char buffer[YFM_HEADER_MAX];
     unsigned char *tokstart, *tokend;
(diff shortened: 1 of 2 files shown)


Source/App/EbAppInputy4m.c
```diff
@@ -36,13 +38,13 @@ int32_t readY4mHeader(EbConfig_t *cfg){
             continue;
         switch (*tokstart++) {
         case 'W': /* width, required. */
-            width = strtol(tokstart, &tokend, 10);
+            width = strtol((const char*)tokstart, (char**)&tokend, 10);
```
kylophone: To avoid all this casting, `tokstart` and `tokend` should just be initialized as `const char*`.

kylophone:
Thanks. Made one comment inline.

Spacefish:
Good point i changed the datatype ;)

kylophone:
Looks good. Thanks!

PR #34, opened 2019-02-08
build.sh: Compat with old Bash versions (macOS)

tux3:
Should fix https://github.com/OpenVisualCloud/SVT-AV1/pull/8#issuecomment-461713169


kylophone:
Fixes the MacOS Travis build, thanks!

PR #36, opened 2019-02-09
Adding unit testing framework

andrey-norkin:
1. Added Google Test framework
2. Added a unit test example


kylophone:
@hassount, Does the VS error message in the appveyor build look familiar to you? We don't have any Windows machines to reproduce.

hassount:
@kylophone 
The windows error message is due to warnings in the test project and for that file all warnings are considered errors. so we should probably try to address all warnings and enable the -Werror flag every where.

andrey-norkin:
@hassount Yes, there are warnings on Windows on the test project and in a couple of other places. The Mac build also seems to generate warnings in various places unrelated to the patch (but warnings are not considered errors there).

hassount:
@andrey-norkin 
The build is now broken for Windows, is that intended ?

Issue #4, opened 2019-02-01
Add continuous integration


EwoutH:
First of all, huge kudo's to Intel for developing this open source, and kudo's to the people who pushed it inside Intel. It really makes AV1 an amazing example of how symbiotic development is a viable business model.

Aside from that, it would be very nice to have a very solid CI/CD implementation. My current, primitive proposal is to use the following tools:

- [x] Integrate Jenkins server (@kylophone? see [this guide](https://resources.github.com/whitepapers/practical-guide-to-CI-with-Jenkins-and-GitHub/))
- Building & testing with Travis CI
   - [x] Linux (PR #16)
   - [x] macOS (PR #25)
- [x] Building, testing & delivering on Windows with AppVeyor (PR #14)
- [x] Building Linux packages with Snapcraft (PR submitted: #38)
- Coveralls for code coverage
   - [x] Implement coverage framework (PR #25)
   - [x] Add first tests (@andrey-norkin)
- [x] Bit-exact output agains master branch (PR #376)
- [x] Measure encoding quality, speed and efficiency with [AWCY](https://github.com/tdaede/awcy).

Any suggestions on how to improve the continuous integration, testing and delivery are welcome!

kylophone:
Hi,

> First of all, huge kudo's to Intel for developing this open scource, and kudo's to the people who pushed it inside Intel. It really makes AV1 an amazing example of how symbiotic development is a viable business model.
> 
> Aside from that, it would be very nice to have a very solid CI/CD implementation. My current, primitive proposal is to use the following tools:
> 
> * Building & testing on Linux & macOS with Travis CI
> * Building, testing & delivering on Windows with AppVeyor
> * Coveralls for code coverage
> * Bit-exact decodability tested against both aomdec and dav1d.
> 
> Travis and AppVeyor could I implement on monday, and coveralls probably also. But please use this issue to discuss optimal implementations that cover an as large portion of the code base and usage scenario's as possible.

Yes, all of these things are needed. A Jenkins server already exists for CI, but needs to be integrated into this new repo. You'll notice the test coverage is non-existent right now, that will be changing soon. I like the idea of adding a Coveralls and AppVeyor build as well. I'll keep this issue open until all this stuff is integrated. Other suggestions are welcome, too.

Thanks,
Kyle

kylophone:
> * [ ]  Building & testing on Linux & macOS with Travis CI
In addition to these two platforms, the hope is to add Windows to CI as well.

[...]

PR #38, opened 2019-02-09
Add snapcraft.yaml

markshuttle:
I have a personal interest in AV1 so checked out your project and made a snapcraft.yaml. With this landed upstream we can setup a daily build of an edge snap.


EwoutH:
Looks great! I added it to #4, where does it need to be activated? [snapcraft.io](https://snapcraft.io/)?

markshuttle:
Yes. I guessed that svt-av1-enc would be a reasonable command name so wrote the snap that way. To register the snap name:

snap install snapcraft --classic
snapcraft register svt-av1-enc

Then you can setup a webhook to trigger an automatic edge build on new commits via build.snapcraft.io or I think you can get Travis to do the same thing (x86 only).

kylophone:
Will likely integrate this, but for the time being I'll let this hang out for a while. Packaging and release timelines are still being debated. We consider this encoder to be pretty beta. It's possible the API could change, for example.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,27 @@
+name: svt-av1-enc
+base: core18
+version: '0.1'
```
EwoutH: Update to `version: '0.5'`

EwoutH:
Now that the first [pre-release](https://github.com/OpenVisualCloud/SVT-AV1/releases) is tagged, this can be merged when the version number is updated to 0.5 I think?

snap/snapcraft.yaml
markshuttle: Makes sense, yes :) There is a way to grab the version from a file in the source tree or from the git tags, it might be worth switching to that so that the snapcraft doesn't have to be updated at release time.

EwoutH: Could you implement this?

EwoutH:
@hassount I like lowering barriers for users. This does that. Can we merge?

EwoutH:
@markshuttle Is this PR still up to date?

If so, @hassount can you assign someone to review?

hassount:
@EwoutH @markshuttle 
I am still not sure if this is applicable yet as we don't yet have releases. If you think otherwise, let's update it and test it?

markshuttle:
I've updated the snapcraft to build the current version of SVT-AV1, and manually hardcoded the version to 0.8+dev. The snap builds for me (on the current development release of Ubuntu 20.04 LTS) using a VM of 18.04 LTS as the build environment.

There are a couple of things you will want to get, which I can't do for you but which the folks in the snapcraft.io Forum Discourse can certainly help. https://forum.snapcraft.io/categories is a good place to start.

First, figure out the magic incantation to get the version auto-set based on git tags. 'git describe' doesn't seem to be producing usable output for me in my branch, but then I don't know much about git. This should mean that anybody who checks out the branch can then just run 'snapcraft' and they will get a snap of that branch with a reasonable version number, like 0.8+git-3423423.

Second, register the snap name at snapcraft.io. That should really be done by someone upstream who is involved in the release process. Pointers at https://snapcraft.io/docs/releasing-to-the-snap-store

Third, I would try to get a daily build into the edge channel for your snap. I think Travis can stuff them there, or you can have snapcraft.io follow your github repo and automatically build snaps for you. If it's useful for you to have non-x86 architectures, snapcraft can build those for you too <duck> :)

After that, making a release into the stable channel is a single command exercise. You can test the specific revision that's in the edge channel, then release that revision in the stable channel, and everyone will get it.

If in future you will have long-lived versions that are maintained in parallel (like 1.0.1, 1.0.2, 1.0.3 in parallel with 1.1.0, 1.1.1, 1.1.2 etc) then you can do that with 'tracks'. But if you are really just going to have one version that you want everyone to use at any given time, just use the 'latest' track which was what you get out of the box.

jbkempf:
Can we please merge this?

EwoutH:
@hassount Can we P1 this and promote it to the 0.8.1 milestone?

jbkempf:
Thanks.

EwoutH:
For now we can just update the version variable manually with each tag I think.

hassount:
> For now we can just update the version variable manually with each tag I think.

it's really frustrating that we can't get the version number automatically generated especially the commit hash so that we have some form of tracking. trying to look into this.

hassount:
merging this for now, we'll work on the automated rev later

EwoutH:
@hassount Nice! Can you coordinate this part?
> Second, register the snap name at snapcraft.io. That should really be done by someone upstream who is involved in the release process. Pointers at https://snapcraft.io/docs/releasing-to-the-snap-store

jbkempf:
> @hassount Nice! Can you coordinate this part?
> 
> > Second, register the snap name at snapcraft.io. That should really be done by someone upstream who is involved in the release process. Pointers at https://snapcraft.io/docs/releasing-to-the-snap-store

We can expose it through videolan, if you want, where we expose already vlc and dav1d.

hassount:
> > @hassount Nice! Can you coordinate this part?
> > > Second, register the snap name at snapcraft.io. That should really be done by someone upstream who is involved in the release process. Pointers at https://snapcraft.io/docs/releasing-to-the-snap-store
> 
> We can expose it through videolan, if you want, where we expose already vlc and dav1d.

@EwoutH yes, going through the process now.
@jbkempf sounds good, I should still register the snap name though right ?

jbkempf:
> > > @hassount Nice! Can you coordinate this part?
> > > > Second, register the snap name at snapcraft.io. That should really be done by someone upstream who is involved in the release process. Pointers at https://snapcraft.io/docs/releasing-to-the-snap-store
> > 
> > 
> > We can expose it through videolan, if you want, where we expose already vlc and dav1d.
> 
> @EwoutH yes, going through the process now.
> @jbkempf sounds good, I should still register the snap name though right ?

I will see with @thresheek what needs to be done.

markshuttle:
Hi folks, just to say thank you for onboarding the snapcraft.yaml

You should easily be able to get automatic versions from tags (I just didn't know the correct incantation). I found [this discussion](https://forum.snapcraft.io/t/getting-the-git-commit-tag-as-a-version/14716/3) on the snapcraft forums which I think is exactly the hint you need. In short, if you can write a little shell script that can decide the version you want, then you can have that set during the build or stage phases.

https://forum.snapcraft.io/t/getting-the-git-commit-tag-as-a-version/14716/3

PR #43, opened 2019-02-11
Intra opt

okhlif:
New trade-off for Intra prediction  For M1, M2 and M3
Valgrind fix

### DIFF
--- a/Source/Lib/Codec/EbModeDecision.c
+++ b/Source/Lib/Codec/EbModeDecision.c
@@ -1779,20 +1779,17 @@ void  inject_intra_candidates(
         disable_z2_prediction       = 0;
         disable_angle_refinement    = 0;
         disable_angle_prediction    = 1;
-    }
-    else if (picture_control_set_ptr->parent_pcs_ptr->intra_pred_mode == 1) {
+    } else if (picture_control_set_ptr->parent_pcs_ptr->intra_pred_mode == 1) {
         disable_z2_prediction       = 0;
         disable_angle_refinement    = 0 ;
         disable_angle_prediction    = (context_ptr->blk_geom->sq_size > 16 ||
                                        context_ptr->blk_geom->bwidth == 4 ||
                                        context_ptr->blk_geom->bheight == 4) ? 1 : 0;
-    }
-    else if (picture_control_set_ptr->parent_pcs_ptr->intra_pred_mode == 2) {
+    } else if (picture_control_set_ptr->parent_pcs_ptr->intra_pred_mode == 2) {
         disable_z2_prediction       = 1;
         disable_angle_refinement    = 1;
         disable_angle_prediction    = 0;
-    }
-    else if (picture_control_set_ptr->parent_pcs_ptr->intra_pred_mode == 3) {
+    } else if (picture_control_set_ptr->parent_pcs_ptr->intra_pred_mode == 3) {
         disable_z2_prediction       = (context_ptr->blk_geom->sq_size > 16 ||
                                        context_ptr->blk_geom->bwidth == 4 ||
                                        context_ptr->blk_geom->bheight == 4) ? 1 : 0;
@@ -1800,8 +1797,7 @@ void  inject_intra_candidates(
                                        context_ptr->blk_geom->bwidth == 4 ||
                                        context_ptr->blk_geom->bheight == 4) ? 1 : 0;
         disable_angle_prediction    = 0;
-    }
-    else {
+    } else {
         disable_z2_prediction       = 0;
         disable_angle_refinement    = 0;
         disable_angle_prediction    = 0;
--- a/Source/Lib/Codec/EbModeDecision.c
+++ b/Source/Lib/Codec/EbModeDecision.c
@@ -2027,7 +2027,7 @@ uint8_t product_full_mode_decision(
 
 
     if (candidate_ptr->type == INTER_MODE && candidate_ptr->merge_flag == EB_TRUE) {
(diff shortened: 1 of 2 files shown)


Source/Lib/Codec/EbModeDecision.c
```diff
@@ -1771,58 +1771,96 @@ void  inject_intra_candidates(
...
+                                       context_ptr->blk_geom->bwidth == 4 ||
+                                       context_ptr->blk_geom->bheight == 4) ? 1 : 0;
+        disable_angle_prediction    = 0;
+    }
+    else {
+        disable_z2_prediction       = 0;
+        disable_angle_refinement    = 0;
+        disable_angle_prediction    = 0;
+
+    }
```
kylophone:
```
if (condition_1) {
    ...
} else if (condition_2) {
    ...
} else {
    ...
}
```

Source/Lib/Codec/EbModeDecision.c
```diff
@@ -2027,7 +2065,7 @@ uint8_t product_full_mode_decision(
 
 
     if (candidate_ptr->type == INTER_MODE && candidate_ptr->merge_flag == EB_TRUE) {
-        context_ptr->md_ep_pipe_sb[cu_ptr->leaf_index].chroma_distortion = buffer_ptr_array[lowestCostIndex]->candidate_ptr->chroma_distortion;
+        context_ptr->md_ep_pipe_sb[cu_ptr->mds_idx].chroma_distortion = buffer_ptr_array[lowestCostIndex]->candidate_ptr->chroma_distortion;
```
kylophone: unchecked deref? I see this happens also happens all over this function.

Source/Lib/Codec/EbPictureDecisionProcess.c
```diff
@@ -401,17 +401,24 @@ EbErrorType signal_derivation_multi_processes_oq(
...
+            picture_control_set_ptr->intra_pred_mode = 4;
+        else
+            picture_control_set_ptr->intra_pred_mode = 2;
+    }
+    else {
+        if (picture_control_set_ptr->temporal_layer_index == 0)
+            picture_control_set_ptr->intra_pred_mode = 3;
+        else
+            picture_control_set_ptr->intra_pred_mode = 0;
+    }
```
kylophone: Are there other enc_modes? Could make this a switch. Just a suggestion, feel free to leave as is.

Source/Lib/Codec/EbModeDecision.c
```diff
@@ -1771,58 +1771,96 @@ void  inject_intra_candidates(
...
-                            picture_control_set_ptr->parent_pcs_ptr->reduced_tx_set_used);
-                    // candidateArray[canTotalCnt].transform_type[PLANE_TYPE_UV]            = context_ptr->blk_geom->bwidth_uv >= 32 || context_ptr->blk_geom->bheight_uv >= 32  ? DCT_DCT : chroma_transform_type[candidateArray[canTotalCnt].intra_chroma_mode];
-                    candidateArray[canTotalCnt].mpm_flag = EB_FALSE;
-                    candidateArray[canTotalCnt].ref_frame_type = INTRA_FRAME;
-                    candidateArray[canTotalCnt].pred_mode = (PredictionMode)openLoopIntraCandidate;
-                    candidateArray[canTotalCnt].motion_mode = SIMPLE_TRANSLATION;
-                    ++canTotalCnt;
-                }
+
+            if (!disable_angle_prediction) {
```
kylophone: There seems to be a lot of duplicated code in the following `else` block. Wouldn't hurt to condense, so as not to repeat yourself.

Source/Lib/Codec/EbModeDecision.c
```diff
@@ -1771,58 +1771,96 @@ void  inject_intra_candidates(
...
+                        candidateArray[canTotalCnt].cfl_alpha_idx = 0;
+                        candidateArray[canTotalCnt].is_directional_chroma_mode_flag = (uint8_t)av1_is_directional_mode((PredictionMode)candidateArray[canTotalCnt].intra_chroma_mode);
+                        candidateArray[canTotalCnt].angle_delta[PLANE_TYPE_UV] = 0;
+                        candidateArray[canTotalCnt].transform_type[PLANE_TYPE_Y] = DCT_DCT;
+
+                        if (candidateArray[canTotalCnt].intra_chroma_mode == UV_CFL_PRED)
+                            candidateArray[canTotalCnt].transform_type[PLANE_TYPE_UV] = DCT_DCT;
+                        else
+                            candidateArray[canTotalCnt].transform_type[PLANE_TYPE_UV] =
+                            av1_get_tx_type(
```
kylophone: According to our style guide this is not the preferred way of writing a function call. This is really prevalent in the codebase, so leave as-is for now and we will decide if we wish to enforce the guide later. My preference would be to follow the style guide.

Source/Lib/Codec/EbModeDecision.c
okhlif: leaf_index should be removed . + chroma_distortion is not used now . To be part of future clean ups
It was just to clear Valgrind to look for other problems.

Source/Lib/Codec/EbModeDecision.c
okhlif: Makes sense let's decide how to proceed. I remember we discussed such but we did not decide on it .

Source/Lib/Codec/EbModeDecision.c
okhlif: I agree. We are not final here. We will do it after we converge on the settings.

Source/Lib/Codec/EbModeDecision.c
okhlif: Fixed.. Thanks

hassount:
@kylophone 
is this PR good to go after the changes?

PR #50, opened 2019-02-12
Warnings fixed (removing unused functions, etc)

joelsole:


kylophone:
lgtm, I'll let others chime in about the deleted functions. Can always be recovered from history if they're needed later.

hassount:
@chkngit @NaderMahdi
Are any of the functions removed needed ?

chkngit:
as a general comment, it is good to keep the intrinsic functions around in case we would need them in the future. if removed we might think we need to write them again. other than that I see bench oh hevc based functions that we definitely not need now.

ttrigui:
Correct we might need to keep a list of intrinsics kernels that have been removed as a reference for future usage. But I believe it is fine if we remove them from the code for now especially if a function has multiple optimized kernel versions (i.e SSE and AVX)

PR #55, opened 2019-02-14
Buffer optimizations to reduce the memory size

hassount:
+ update buffer configuration to reduce commit memory to by 40%  …
+ make buffer optimization applicable to low core count (e.g. 4-core machines) reducing the memory size by ~50%
+ add function to cap the look-ahead distance to what is needed only by the encoder (e.g. 2*minigop +1 for CQP)
+ remove tabs / fix coding style


hassount:
@chkngit 
Could you please take a look whenever you have a chance ?

Source/Lib/Codec/EbEncHandle.c
```diff
@@ -391,38 +381,42 @@ void LoadDefaultBufferConfigurationSettings(
...
 
     //#====================== Inter process Fifos ======================
-    sequence_control_set_ptr->resource_coordination_fifo_init_count = 300;
-    sequence_control_set_ptr->picture_analysis_fifo_init_count = 300;
-    sequence_control_set_ptr->picture_decision_fifo_init_count = 300;
-    sequence_control_set_ptr->initial_rate_control_fifo_init_count = 300;
-    sequence_control_set_ptr->picture_demux_fifo_init_count = 300;
-    sequence_control_set_ptr->rate_control_tasks_fifo_init_count = 300;
-    sequence_control_set_ptr->rate_control_fifo_init_count = 301;
+    sequence_control_set_ptr->resource_coordination_fifo_init_count       = 300;
```
kylophone: Are all these fifos actually full for all configurations, or is this a worst case? Would it make sense to start smaller and grow them?

hassount: These are only pointers between the processes and they have been already reduced from 3000 to 300 before the OS. We can keep it as is for now, and optimize it further when we start going faster with the encoder?

hassount:
@kylophone @andrey-norkin @joelsole @mafonsoN 
Can we go ahead with this ?

Issue #15, opened 2019-02-04
Error on FFmpeg patch


EwoutH:
I got this error when trying to compile the FFmpeg patch.

- Compile commands: https://github.com/EwoutH/SVT-AV1/blob/248f8486e93f8259093b57437f33bf8b3f882d85/.travis.yml
- Full log: https://travis-ci.com/EwoutH/SVT-AV1/jobs/175239496#L2588

```
CC	libavcodec/libsvt_av1.o
libavcodec/libsvt_av1.c: In function ‘config_enc_params’:
libavcodec/libsvt_av1.c:200:14: error: ‘EbSvtAv1EncConfiguration {aka struct EbSvtAv1EncConfiguration}’ has no member named ‘video_usability_info’
         param->video_usability_info = svt_enc->vui_info;
              ^
libavcodec/libsvt_av1.c: In function ‘eb_enc_init’:
libavcodec/libsvt_av1.c:303:1: warning: label ‘failed_init_enc’ defined but not used [-Wunused-label]
 failed_init_enc:
 ^
ffbuild/common.mak:60: recipe for target 'libavcodec/libsvt_av1.o' failed
make: *** [libavcodec/libsvt_av1.o] Error 1
make: *** Waiting for unfinished jobs....
The command "make -j `nproc`" exited with 2.
```

vannan-mani:
Have the same problem, since this was usability information, commented the above, plugin compiles and installs correctly. 
On initialization comes up with an arithmetic failure. Attached the debugger logs

I did try the patch against the latest ffmpeg on the trunk and the patch had no compilations. There was a problem with non-available member as below
 /*    if (svt_enc->vui_info)
    200         param->video_usability_info = svt_enc->vui_info;
    201 */

since this was regarding video usability information, i decided to comment this out and the ffmpeg code compiled and launched.

now on a simple encode, the plugin throws up a floating point exception and coredumps
 ffmpeg  -i pt.mov -c:v libsvt_av1 -y pt_av1.mp4
ffmpeg version N-93075-gb429c86 Copyright (c) 2000-2019 the FFmpeg developers
  built with gcc 7 (Ubuntu 7.3.0-27ubuntu1~18.04)
[...]

PR #57, opened 2019-02-14
Removed the out-of-date parameter video_usability_info from ffmpeg_plugin.

jsunintel:


jsunintel:
Fixed https://github.com/OpenVisualCloud/SVT-AV1/issues/15.

hassount:
@kylophone 
I think we can now remove the allowed failure for the ffmpeg patch

PR #58, opened 2019-02-14
Handle NULL Pointer in api functions

ciwenyao:
Return EB_ErrorBadParameter if the pointer is NULL


ciwenyao:
Added it since I'm adding an api unit test, and found it's crashing on invalid parameters.

kylophone:
Thanks for the fix.

PR #59, opened 2019-02-14
Convert sse4 avx2 small

spawlows:


gnafuthegreat:
Is this removing SSE2 and SSE4.1 SIMD while introducing AVX2, or am I misunderstanding?  Is the intention to only support AVX2 SIMD going forward and only have C for older/cheaper processors?  As someone with older and cheaper machines, I'd hate to lose optimizations that used to be there ;-).

EwoutH:
While I agree that new optimizations should primarily be aimed at AVX2 and AVX-512, there is really no reason to take already made optimizations for older SIMD extensions out. Also low-effort high-gain optimizations for SSE4.1 should still be made.

Spacefish:
When i first read the code i wondered if the switch between MMX/SSE Instructions and AVX comes with a performance penality as described in https://software.intel.com/en-us/articles/avoiding-avx-sse-transition-penalties is this why you replace the SSE versions with the AVX equivalent?

cdgriffith:
While I won't pretend to understand the code, I can chip in with some usage stats. Using this build (1.0.65) vs the most recent (1.0.113) this one was 15~20% faster on a i9-9900k encoding a 720p video. 

Settings used: `-enc-mode 3 -bit-depth 8 -w 1280 -h 720 -fps 25 -q 30`

Latest Build (1.0.113): 
```
time to encode: 117.37 seconds
resulting file size:  24,828 KB
max processor usage: 53% 
mac CPU temp:  79* C 
```

This PR (1.0.65):
```
time to encode: 99.10 seconds
resulting file size: 23,777 KB
max processor usage: 90%
mac CPU temp: 89* C 
```

So at least for my setup this code pegged my CPU as I would expect and had a faster result with smaller file size.

ttrigui:
looks fine to me.

singhkays:
@cdgriffith How did you get the max processor usage? And is that average across all the cores?

cdgriffith:
@singhkays Using HWMonitor and clearing Min/Max between runs, and yes was average across all cores. 

Running again with builds 1.0.118 and1.0.119 the performance was also similar to 1.0.113 (1.0.119 was slightly improved to 114 seconds from 117 seconds) so there must have been significant changes since 1.0.65 to slow down the processing speed and this one might not have been the true cause of speed improvement.

EwoutH:
@spawlows @hassount Is there a certain reason the SSE got removed in this PR? If not, can we add it back to keep older processors as fast as possible?

EwoutH:
@spawlows @hassount So?

hassount:
@EwoutH I don't see a reason why it should not be brought up, we can add it to #262 ?

Issue #39, opened 2019-02-10
Add support for static library build


tmatth:
Currently it seems that the library is hardcoded to build a .so.

bleunova:
Yes, I like to know how to build static library. Any help would be appreciated.

jstebbins:
Hi all.  I'm looking into adding static lib support.  I basically have it working, but there are duplicate symbols in the App and the lib that case final link to fail (symbols are str\*_ss, EbStartTime and EbFinishTime). They are duplicated because the shared lib doesn't export them, and they are needed in the App.  

I'm wondering if these things should become part of the library's exported API or if you would prefer to keep them separate and just de-duplicate by changing function names?  The str\*_ss functions appear to be currently only used in the App even though the functions are duplicated in the lib.  EbStartTime and EbFinishTime are used in both the App and the lib.

FYI, I was also able to coax it into building with mingw with a few small changes.

kylophone:
Hi,
> Hi all. I'm looking into adding static lib support. I basically have it working, but there are duplicate symbols in the App and the lib that cause final link to fail (symbols are str*_ss, EbStartTime and EbFinishTime). They are duplicated because the shared lib doesn't export them, and they are needed in the App.
> 
> I'm wondering if these things should become part of the library's exported API or if you would prefer to keep them separate and just de-duplicate by changing function names? The str*_ss functions appear to be currently only used in the App even though the functions are duplicated in the lib. EbStartTime and EbFinishTime are used in both the App and the lib.

 I wonder if we could somehow do without those functions is the app. They're not part of the encoding API, so they should not be exported.  I think the string functions are mostly used for parsing the command line options (iirc), which could/should be replaced with something like `getopt.h`. There are shell built-ins for time on most platforms, so that might be OK to axe as well. I'll let others chime in.

> 
> FYI, I was also able to coax it into building with mingw with a few small changes.

Ah, great. Let me know If you get that going, It would be nice to have it in Travis.

Thanks,
Kyle

jstebbins:
For now, I've stripped those function definitions from App and exported them from the lib so that App can use them.  If the decision is to also strip their usage from App, it would be a simple thing to un-export these functions.

jstebbins:
I submitted a PR for adding static library build support.  MinGW build support is also working, see https://github.com/jstebbins/SVT-AV1/tree/mingw. It's built on top of the static lib PR, so I'll hold off on submitting that until the static lib PR is reviewed and accepted.

[...]

PR #67, opened 2019-02-19
Adds static library support to build

jstebbins:
Also de-duplicates multiply defined functions.

See https://github.com/OpenVisualCloud/SVT-AV1/issues/39


hassount:
@kylophone @EwoutH 
Do you think we should update travis to build both static and shared ?

jstebbins:
By default, this PR would build both.  I have only provided a way to disable shared (-DENABLE_SHARED=OFF).  Did you also want a way to disable static?

hassount:
@jstebbins 
Thank you very much for the PR, this looks good to me, a couple of items to change though: 
- We don't need the string functions in the library and we would like to restrict their usage to the application only. Could you please not have them as APIs and make EbString.h /c an application only file. 
- The string safe functions we got from the safe string open source library, so the license header should reflect so and can be found in the EbAppFifo.c file
- The only function that would be needed in the library is the memset function which is a safe function to use so you can change that in the library without having to carry it as an API. 

Please let me know if you need help with these changes and we could help addressing some of them.

jstebbins:
@hassount sounds good.  It might be a few days before I can get back to it, but I'll update the PR with those changes.

jstebbins:
This commit a6c1f81989c6cab0f477adfa867d5ff3dad2725c just introduced use of the string functions into the library (it uses strnlen_ss).  How would you like to resolve this?

jstebbins:
For now, to resolve the conflict, I just changed EbEncHandle.c to use regular strnlen, see 1173b38. Please let me know if this is not a suitable way to handle this.

Force pushed to clean up history and resolve conflicts with master.

hassount:
@jstebbins 
No Problem, I would also wait for #135 to avoid conflicts.

jstebbins:
Sure, I'll just rebase this on whatever comes along.  You can merge whenever you think appropriate.

tmatth:
> Sure, I'll just rebase this on whatever comes along. You can merge whenever you think appropriate.

Looks like #135 was merged.

jstebbins:
Thanks for the ping.  Conflicts resolved.

I'm not sure if my changes to the time functions fits properly into the API rework that just happened.  Please let me know if that needs to be done differently.

1480c1:
@jstebbins can you confirm if the conflicts are resolved, it will still build static libs? Right now, if I build using mingw-w64, it will produce .dll and .dll.a which I personally feel is undesirable.

jstebbins:
The latest set of conflicts are not resolved as github indicates.  I'll fix that when I have time to do so.  There is a cmake veraible to disable shared libs if you don't want the dll
~~~
cmake -DENABLE_SHARED=off
~~~

bleunova:
Hi @jstebbins any update on this would be appreciated. :)

jstebbins:
Conflicts resolved.

There were also a few bugs in the generation of libSvtAv1Dec that had to be fixed in order to link the static lib version.

1480c1:
One thing I would like to mention is that cmake comes with a variable called `BUILD_SHARED_LIBS` that would select shared or static for the add_library functions that don't have a specific mode set. You could use that instead of creating a custom variable

jstebbins:
BUILD_SHARED_LIBS does one or the other.  There is no nice way that I know of using BUILD_SHARED_LIBS to say, I may want one, or the other, or both.

1480c1:
What do you mean? I'm a bit confused.

1480c1:
Do you mean some of them as static libs and some as shared?

jstebbins:
There are 2 add_library instances. One for shared and one for static libraries.  By default, both are built.

1480c1:
I see, so by default it will build both static and shared. Which will get installed/linked?

1480c1:
I'm trying to understand what is the design

jstebbins:
Both get installed by default.  If shared libs are enabled, the SvtApp's are linked against the shared lib.  If shared is disabled, the SvtApp's are linked against the static libs.  External programs that use the libraries would choose which they link to depending on the needs of the app.

1480c1:
One thing you can do is leave everything as it is but use that variable instead of the custom variable because it will only make the libs shared if shared or static is not stated explicitly. So since you have shared and static stated explicitly, it would work perfectly fine in this case

1480c1:
at least from what I have seen

1480c1:
I'm going to test it in a sec

jstebbins:
Assume I left one call to add_library unspecified so that BUILD_SHARED_LIBS could control the type of the library. What would I set in the other add_library call?  The only solution that makes any sense that I can think of would look like this:
~~~
if(BUILD_SHARED_LIBS)
    # Encoder Lib Source Files
    add_library(SvtAv1EncShared SHARED
        ${all_files}
        $<TARGET_OBJECTS:COMMON_CODEC>
        $<TARGET_OBJECTS:COMMON_C_DEFAULT>
        $<TARGET_OBJECTS:COMMON_ASM_SSE2>
        $<TARGET_OBJECTS:COMMON_ASM_SSSE3>
        $<TARGET_OBJECTS:COMMON_ASM_SSE4_1>
        $<TARGET_OBJECTS:COMMON_ASM_AVX2>
    )
    set_target_properties(SvtAv1EncShared PROPERTIES OUTPUT_NAME SvtAv1Enc)
    target_link_libraries(SvtAv1EncShared ${PLATFORM_LIBS})
    install(TARGETS SvtAv1EncShared DESTINATION "${CMAKE_INSTALL_LIBDIR}")
endif(BUILD_SHARED_LIBS)

add_library(SvtAv1EncStatic STATIC
    ${all_files}
    $<TARGET_OBJECTS:COMMON_CODEC>
    $<TARGET_OBJECTS:COMMON_C_DEFAULT>
    $<TARGET_OBJECTS:COMMON_ASM_SSE2>
    $<TARGET_OBJECTS:COMMON_ASM_SSSE3>
    $<TARGET_OBJECTS:COMMON_ASM_SSE4_1>
    $<TARGET_OBJECTS:COMMON_ASM_AVX2>
)
~~~
I could leave off the 'SHARED' in the add_library call above, but that's kind of pointless and obscures what the code is really doing.  So it would allow you to use the variable BUILD_SHARED_LIBS, but it doesn't use the built-in behaviour of add_library since it's not necessary and just obscures the meaning of the code.

Note that the only difference between this and what I've done in the PR is in the 'if' conditional.

1480c1:
That's what I mean

1480c1:
That way both static and shared can be built and assuming that the rest of the logic is fine, it should be as you written

1480c1:
The only real reason why I prefer to use `BUILD_SHARED_LIBS` is because that way I don't have to remember if a certain package used a special variable to do certain stuff.

1480c1:
Changing all of the `ENABLE_SHARED` variables to `BUILD_STATIC_LIBS` worked and both shared and static are built and with it enabled, the EncApp is linked with the shared objects

jstebbins:
OK, I'll update the PR.  Thanks.

1480c1:
When running with it off, the app is statically built, although it still does link to shared pthread, m, c, vdso, and ld which might be a bit annoying

1480c1:
One thing that can be done is <https://stackoverflow.com/a/29824424> so that way it would be a little cleaner

jstebbins:
If you are referring to object libraries, I'm already using them. Previously, it built static libs for the avx, sse, etc. dsp functions.  Then it linked those static libs into the shared lib.  You can't directly add a static lib to another static lib, so it was necessary to replace those with object libraries.

1480c1:
I was thinking more in the codec cmakelist using the `OBJECT IMPORTED` keyword for add_library, but trying it, it gave me undefined references, so nevermind

jstebbins:
Conflicts fixed again...

Source/API/EbSvtAv1Time.h
```diff
@@ -6,14 +6,24 @@
...
+
 #define NANOSECS_PER_SEC ((uint32_t)(1000000000L))
 
-void EbStartTime(uint64_t *Startseconds, uint64_t *Startuseconds);
-void EbFinishTime(uint64_t *Finishseconds, uint64_t *Finishuseconds);
-void EbComputeOverallElapsedTime(uint64_t Startseconds, uint64_t Startuseconds, uint64_t Finishseconds, uint64_t Finishuseconds, double *duration);
-void EbComputeOverallElapsedTimeMs(uint64_t Startseconds, uint64_t Startuseconds, uint64_t Finishseconds, uint64_t Finishuseconds, double *duration);
-void EbSleep(uint64_t milliSeconds);
-void EbInjector(uint64_t processedFrameCount, uint32_t injector_frame_rate);
+EB_API void EbStartTime(uint64_t *Startseconds, uint64_t *Startuseconds);
```
hassount: I understand the need to combine these calls into one, but I don't think that we will need to make it an API as such has other consequences. maybe we can keep them with different names at this point (one with a prefix eb then the app side would be without an eb prefix).

Source/Lib/Encoder/Codec/EbEncHandle.c
```diff
@@ -275,8 +228,8 @@ EbErrorType InitThreadManagmentParams() {
 #elif defined(__linux__)
     const char* PROCESSORID = "processor";
     const char* PHYSICALID = "physical id";
-    int processor_id_len = strnlen_ss(PROCESSORID, 128);
-    int physical_id_len = strnlen_ss(PHYSICALID, 128);
+    int processor_id_len = strnlen(PROCESSORID, 128);
```
hassount: we can't use strnlen as it is an unsafe c function. We'll need to use strnlen_ss

Source/API/EbSvtAv1Time.h
jstebbins: If there are no objections, I'll do as you suggest.

Source/Lib/Encoder/Codec/EbEncHandle.c
jstebbins: Then I'll need to also duplicate these functions and give them a different name as you suggested for the time functions above.  I was told previously that you don't want these string functions to be part of the library API.

hassount: Sure, let's do so, eb inside the library, no eb at the application

Source/API/EbSvtAv1Time.h
hassount: @andrey-norkin any issues with this ?

jstebbins:
Last 2 commits should satisfy your requests...

1480c1:
Please make sure to check the ffmpeg patches, they have references to EbSvtAv1Time.h

1480c1:
Either modify those directly or ask @jsunintel to generate new patches with the right headers to add to this pr.

jstebbins:
Yes, I see the failures.  Working on fixing...

EwoutH:
@hassount @kylophone Ready to merge?

EwoutH:
@kylophone Could you please review?

hassount:
@jstebbins 
Would you be able to add a little switch to swap between building shared and or static libraries in build.sh ?

jstebbins:
Sorry for the delay. Was travelling.

I have a build script I wrote that's based on Build/linux/build.sh, but I've modified it quite a lot.  Generalized it to make mingw cross compiling easier and such.  I can replace the current linux script with that and let you have a look to see if it's acceptable.

jstebbins:
I'm confused by the travis failure that just appeared.  The build log indicates an error in code that does not appear in my local copy of this branch.  I.e. mrp_mode is not referenced in my copy of Source/Lib/Common/Codec/EbModeDecision.c

jstebbins:
My last commit fixes an  error in master that is unmasked by my changes...

hassount:
@jstebbins 
can you reproduce the travis build failure on your end ?

jstebbins:
Yes.  I'll see if I can track it down.  I won't have time to look at it till Friday though.

jstebbins:
I had a little time to investigate.  It looks to me like a bad organization of ```#define``` and header file dependencies.  MEMORY_FOOTPRINT_OPT_ME_MV is defined in Lib/Common/Codec/EbDefinitions.h.  API/EvSvtAv1Enc.h uses MEMORY_FOOTPRINT_OPT_ME_MV.  App includes API/EvSvtAv1Enc.h, but does not (and can not) include EbDefinitions.h.  So structure layout is different between what the app expects and what the library expects.

So it appears to be a problem in git head that is exposed by my build changes.

hassount:
> I had a little time to investigate. It looks to me like a bad organization of `#define` and header file dependencies. MEMORY_FOOTPRINT_OPT_ME_MV is defined in Lib/Common/Codec/EbDefinitions.h. API/EvSvtAv1Enc.h uses MEMORY_FOOTPRINT_OPT_ME_MV. App includes API/EvSvtAv1Enc.h, but does not (and can not) include EbDefinitions.h. So structure layout is different between what the app expects and what the library expects.
> 
> So it appears to be a problem in git head that is exposed by my build changes.

Good catch, let's fix the master, this will hopefully address the merge conflicts.

jstebbins:
> The command "wget -nc http://github.com/OpenVisualCloud/SVT-AV1-Resources/raw/master/video.tar.gz" failed and exited with 8 during

Build failure appears to be a network issue

1480c1:
I seems to be repo wide since https://github.com/OpenVisualCloud/SVT-AV1/pull/264 also failed due to similar issues

jstebbins:
Resolved recent conflicts and updated build script with unit test options

hassount:
@kylophone @joelsole 
Any issues with this ?

hassount:
@jstebbins ready to merge ? 

@kylophone @joelsole any issues with this PR ?

1480c1:
Although this is a bit late, would it be possible to add some way to disable building shared entirely through something like a separate variable? It doesn't have to be in this pr

jstebbins:
@hassount I believe so.  I discovered that last little bit while updating my mingw branch to fix building of unit tests.

@1480c1 ```build.sh -x``` disables building shared library.  Only static libSvtAv1Enc.a (and libSvtAv1Dec.a) is built and the binary is linked to it.  If you are talking about linking only static *system* libs, that's another PR I think with it's own variable to enable. It *should* simply be a matter of adding ```-static``` to ```CMAKE_EXE_LINKER_FLAGS```.  I did a quick test by manually adding this in the top CMakeLists.txt file and I got a completely static binary.

1480c1:
>  It should simply be a matter of adding -static to CMAKE_EXE_LINKER_FLAGS. I did a quick test by manually adding this in the top CMakeLists.txt file and I got a completely static binary.

Nice.

kylophone:
Yeah, let's merge it. @hassount, I'll let you do the honours.

PR #69, opened 2019-02-20
Improved encoder M0 settings

NaderMahdi:
** Better speed/quality for M0 (About 70% speed up for similar BDrate)**

** Intra_pred_mode:
 level 4 for base layer pictures and Level 2 otherwise.

**SG-filter level:
  Use gi_step equal to 16

**NSQ_search_Level
	Use level 2;
** Tx Search
    use tx_weight equal to 125.


hassount:
@NaderMahdi 
lgtm, but can you please fix the linux warnings?

NaderMahdi:
Done.

Source/Lib/Codec/EbPictureDecisionProcess.c
```diff
@@ -791,17 +812,62 @@ EbErrorType signal_derivation_multi_processes_oq(
...
+        break;
+    case 3:
+        intra_pred_level = 2; //ENC_M3
+        break;
+    default:
+        intra_pred_level = 4; //MR_MODE
+        break;
+    }
+
+    if (intra_pred_level == 4) {
```
kylophone: You could combine all this logic into the switch above. Might be easier to read.

NaderMahdi: This is just a temporal code to allow the testing team to build their own combination as they wish. I will be fixed once we converge on the intra_pred_levels for all encoder modes.

hassount:
@kylophone 
Can we proceed with this PR ?

kylophone:
> @kylophone
> Can we proceed with this PR ?

OK

Issue #65, opened 2019-02-17
Crash on YUV4MPEG input


jbeich:
Affects both Clang 8.0 and GCC 8.2. Tested 56df866a7c00 on FreeBSD.

```
$ ffmpeg -i https://media.xiph.org/ED/ED-1080-png/03150.png -pix_fmt yuv420p 03150.y4m
$ SvtAv1EncApp -i 03150.y4m -b 03150.ivf
-------------------------------------------
SVT-AV1 Encoder
SVT [version]:  SVT-AV1 Encoder Lib v0.4.0
SVT [build]  :  GCC 4.2.1        64 bit
LIB Build date: Feb 17 2019 03:08:07
-------------------------------------------
Number of logical cores available: 8
Number of PPCS 31
-------------------------------------------
SVT [config]: Main Profile      Tier (auto)     Level (auto)
SVT [config]: EncoderMode                                                       : 3
SVT [config]: EncoderBitDepth / CompressedTenBitFormat                          : 8 / 0
SVT [config]: SourceWidth / SourceHeight                                        : 1920 / 1080
SVT [config]: Fps_Numerator / Fps_Denominator / Gop Size / IntraRefreshType     : 25 / 1 / 32 / 1
SVT [config]: HierarchicalLevels / BaseLayerSwitchMode / PredStructure          : 4 / 0 / 2
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange                  : CQP / 50 / 33 / 0
-------------------------------------------
Encoding          Assertion failed: (fresult != NULL), function readY4mFrameDelimiter, file Source/App/EbAppInputy4m.c, line 213.
```

kylophone:
@mafonsoN, source is here: https://media.xiph.org/ED/ED-1080-png/03150.png

EwoutH:
Encountered the same error. On a 240 frame file, with no `-n` specified, it crashed instantly.
```
SvtAv1EncApp -i Morocco_1080p_YT_240.y4m -enc-mode 3 -b SVT-test.ivf
Encoding
Assertion failed: fresult != NULL, file C:\projects\svt-av1\Source\App\EbAppInputy4m.c, line 213
```
With `-n 240` it crashed halfway
```
SvtAv1EncApp -i Morocco_1080p_YT_240.y4m -enc-mode 3 -b SVT-test.ivf -n 240
Encoding       109
[...]

PR #76, opened 2019-02-20
Small fixes to the YUV4MPEG2 input support

mafonsoN:
Solves issue #65

### DIFF
--- a/Source/App/EbAppConfig.c
+++ b/Source/App/EbAppConfig.c
@@ -956,7 +956,7 @@ int32_t ComputeFramesToBeEncoded(
     uint32_t frameSize;
     long currLoc;
 
-    currLoc = ftell(config->inputFile); // get current fp location
+    currLoc = ftello64(config->inputFile); // get current fp location
 
     if (config->inputFile) {
         fseeko64(config->inputFile, 0L, SEEK_END);
--- a/Source/App/EbAppConfig.c
+++ b/Source/App/EbAppConfig.c
@@ -954,9 +954,14 @@ int32_t ComputeFramesToBeEncoded(
     uint64_t fileSize = 0;
     int32_t frameCount = 0;
     uint32_t frameSize;
+    long currLoc;
+
+    currLoc = ftell(config->inputFile); // get current fp location
+
     if (config->inputFile) {
         fseeko64(config->inputFile, 0L, SEEK_END);
         fileSize = ftello64(config->inputFile);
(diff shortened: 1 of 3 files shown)


Source/App/EbAppConfig.c
```diff
@@ -954,9 +954,14 @@ int32_t ComputeFramesToBeEncoded(
     uint64_t fileSize = 0;
     int32_t frameCount = 0;
     uint32_t frameSize;
+    long currLoc;
+
+    currLoc = ftell(config->inputFile); // get current fp location
```
hassount: ftello64 ? for files larger than 2GB ?

Source/App/EbAppInputy4m.c
```diff
@@ -50,27 +51,40 @@ int32_t readY4mHeader(EbConfig_t *cfg){
...
+                fprintf(cfg->errorLogFile, "interlace type not supported\n");
                 return EB_ErrorBadParameter;
             }
             if(PRINT_HEADER)
                 printf("scan_type = %c\n", scan_type);
             break;
         case 'C': /* color space, not required: default "420" */
-            if (strncmp("420p16", tokstart, 6) == 0) {
+            if (strncmp("420jpeg", tokstart, 7) == 0) {
+                strcpy(chroma, "420");
```
hassount: @mafonsoN 
While we are at it, these functions are not safe can you use the following instead ?
EB_STRCPY instead of strcpy
EB_STRNCPY instead of strncpy
EB_STRCMP instead of strcmp (just for consistency)

kylophone: Do we have these same requirements for non-library code?

hassount: Yes, this is also causing the warnings on windows: 
https://ci.appveyor.com/project/OpenVisualCloud/svt-av1/builds/22524375

kylophone: That means we need to export all these string functions with the encoding API? Seems like the wrong way to go about it.

hassount: can't we just duplicate them in the app, we don't have to export them from the library ?

mafonsoN: Fixed

Source/App/EbAppConfig.c
mafonsoN: Fixed

hassount:
@mafonsoN thank you for handling the safe functions

hassount:
@mafonsoN  Thank you for updating the c functions.

PR #86, opened 2019-02-21
FFmpeg-Plugin: Add the support to svt hevc and av1 plugins coexistance

jsunintel:


ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
```diff
@@ -103,9 +103,9 @@ index 0000000..1eba049
 +*/
 +
 +#include <stdint.h>
-+#include "EbErrorCodes.h"
-+#include "EbTime.h"
-+#include "EbApi.h"
++#include "svt-av1/EbErrorCodes.h"
++#include "svt-av1/EbTime.h"
++#include "svt-av1/EbApi.h"
```
jbeich: This assumes the parent directory of `pkg-config SvtAv1Enc --cflags` is appended to `CFLAGS`. If SVT-AV1 is installed under non-default `CMAKE_INSTALL_PREFIX` or FFmpeg is configured to minimize external dependencies while such dependencies are separated from base system (e.g., on BSDs) then build would fail.

jsunintel: The prefix before "svt-av1" is set by the PKG_CONFIG_PATH env. If the libsvt-av1 is installed to /usr/local/, the headers will be installed to /usr/local/include/svt-av1/, while if it's installed to /usr/, those will be at /usr/include/svt-av1/.

It's always in [PREFIX]/svt-av1/, and "svt-av1/*" could find the header file.

jsunintel:
FYI，the other patch：https://github.com/intel/SVT-HEVC/pull/111

ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
jbeich: `${prefix}/include` unlike `${prefix}/include/svt-av1` isn't guaranteed to be passed via `CFLAGS` during build. And other dependencies aren't guaranteed to have the same `${prefix}` as SVT-AV1.

jsunintel: install(DIRECTORY ../../API/ DESTINATION "${CMAKE_INSTALL_PREFIX}/include/svt-av1" FILES_MATCHING PATTERN "*.h")

jbeich: But you didn't adjust CFLAGS as returned by pkg-config.

https://github.com/OpenVisualCloud/SVT-AV1/blob/425ba7775b84c0177f0ffd576faea4ed7314af94/Source/Lib/pkg-config.pc.in#L2

```c
$ cat >a.c
#include "svt-av1/EbApi.h"

int main()
{
  eb_init_encoder(0);
  return 0;
}

$ cc a.c $(pkg-config SvtAv1Enc --cflags --libs)
a.c:1:10: fatal error: 'svt-av1/EbApi.h' file not found
#include "svt-av1/EbApi.h"
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

jsunintel: [root@localhost work]# cat a.c
#include "svt-av1/EbApi.h"

int main()
{
  eb_init_encoder(0);
  return 0;
}

[root@localhost work]# cc a.c $(pkg-config SvtAv1Enc --cflags --libs)
[root@localhost work]# ls -al a.*
-rw-r--r--. 1 root root   78 Feb 25 14:53 a.c
-rwxr-xr-x. 1 root root 8432 Feb 25 14:54 a.out

No error observed.

jbeich: GCC/Clang search under `/usr` or `/usr/local` by default. Try installing elsewhere e.g.,
```
$ cmake -DCMAKE_INSTALL_PREFIX=/tmp/test .
$ make -j
$ make install
$ export PKG_CONFIG_PATH=/tmp/test/lib/pkgconfig
$ cc a.c $(pkg-config SvtAv1Enc --cflags --libs)
```

jsunintel: In that case, the headers are still to be placed under svt-av1, that is /tmp/test/svt-av1.
It's the users' job to include /tmp/test and let GCC search there.

jbeich: > It's the users' job to include /tmp/test and let GCC search there.

Every other ffmpeg plugin that uses pkg-config isn't affected. Why SVT-AV1 is special? For one, `pkg-config aom --cflags` returns `-I/tmp/test/include`, not `-I/tmp/test/include/aom`, so it can use `<aom/aom.h>` rather than `<aom.h>`.

jsunintel: SVT-HEVC, SVT-AV1 & SVT-VP9, have a similar set of header files, with the same file names, in which contain different content for each codec, hence the confliction.

hassount: I believe the best way would be to change the file names to include the product name #135  should address that and so we hopefully won't have this problem ? if so, let's make the appropriate changes so that pkg-config would not be affected ?

Issue #84, opened 2019-02-21
segfault with 720x480 yuv420p input


jstebbins:
Using the ffmpeg plugin to test encoding with a 720x480 yuv420p source file.  Crashes in _mm_store_si128.

I had a hunch that the crash was related to the stride or alignment of buffers, so I did a little digging to see where this function's buffer comes from. I created a "fix" for this by aligning strideY, strideCb, and strideCr to 64 bytes (512 bits) and adjusting lumaSize and chromaSize accordingly in EbPictureBufferDescCtor.  I'm not sure this is the right way to fix this or if 64 byte alignment is the right value.  Can someone with more experience with the code comment please?  If this seems like an appropriate fix, I can create a PR

log including gdb backtrace:
```
$ gdb ./ffmpeg_g 
GNU gdb (GDB) Fedora 8.2-6.fc29
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./ffmpeg_g...done.
(gdb) handle SIG32 noprint nostop
Signal        Stop	Print	Pass to program	Description
SIG32         No	No	Yes		Real-time event 32
(gdb) run -i ~/Videos/Samples/PES/ps_mpg.ts  -c:v libsvt_av1 foo.mp4
Starting program: /home/jstebbins/Source/ffmpeg/ffmpeg_g -i ~/Videos/Samples/PES/ps_mpg.ts  -c:v libsvt_av1 foo.mp4
Missing separate debuginfos, use: dnf debuginfo-install glibc-2.28-26.fc29.x86_64
warning: Loadable section ".note.gnu.property" outside of ELF segments
warning: Loadable section ".note.gnu.property" outside of ELF segments
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[...]

PR #88, opened 2019-02-21
Fix crash with 720x480 yuv420p input

jstebbins:
Align strideY, strideCb, and strideCr to 64 bytes (512 bits)

Fixes https://github.com/OpenVisualCloud/SVT-AV1/issues/84

### DIFF
--- a/Source/Lib/Codec/EbDefinitions.h
+++ b/Source/Lib/Codec/EbDefinitions.h
@@ -150,6 +150,8 @@ extern "C" {
 #else
 #define PAD_VALUE                                (128+32)
 #endif
+#define BUFFER_ALIGN                              64
+#define BUFFER_ALIGN_MASK                        (~(BUFFER_ALIGN-1))
 
 //  Delta QP support
 #define ADD_DELTA_QP_SUPPORT                      0  // Add delta QP support - Please enable this flag and iproveSharpness (config) to test the QPM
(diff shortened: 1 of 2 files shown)


ttrigui:
@chkngit do you see any problem with the alignment ?

kylophone:
Does this assume a 64-bit platform? Is there a better way to set this?

jstebbins:
> Does this assume a 64-bit platform?

No 
> Is there a better way to set this?

I don't understand.  What is "this" in this context?

ttrigui:
We can always use unaligned instructions.
@jstebbins Can you point me to the kernel where you saw the crash please?

jstebbins:
@ttrigui did you see the issue referenced in the initial PR comment? https://github.com/OpenVisualCloud/SVT-AV1/issues/84

There's a gdb backtrace there and more explanation of how and where this happens.  In my particular test case, strideY was 1040 and strideCr/Cb was 520.  It crashes in _mm_store_si128 while processing Cr because 520 is not divisible by 16.

ttrigui:
@jstebbins Thank you I actually missed that.
To avoid increasing memory we will just replace _mm_store_si128 by _mm_storeu_si128 to fix the crash.

jstebbins:
I don't know what the performance cost of using unaligned instructions is, but it seems like this is a relatively trivial amount of memory to add.  You already add 320 bytes of padding per row of image.  This adds at most 63 more bytes per row.  This could be reduced to 15 (align to 16 bytes instead of 64) if you know for sure that these buffers are not used by instructions that expect wider alignment. I chose 64 to be on the safe side since I don't know all the ways these buffers are used.

hassount:
@jstebbins 
I think it's best to do the alignment later on for all buffers and not do it for only one buffer, so I prefer the non-aligned solution for now until this is fixed properly.

hassount:
@joelsole @andrey-norkin @kylophone 
Any thoughts on this  ?

ttrigui:
@hassount We will use unaligned instructions.
@spawlows will take care of it

hassount:
Closing this PR as it is covered in #128

Issue #77, opened 2019-02-20
Crash in AVX2 when encoding via FFmpeg


jbeich:
Regressed by #59. Tested on i7-6700K (Skylake) which does have AVX2 support. Lines correspond to a39acb2f2f4a snapshot

```
$ curl -O https://media.xiph.org/video/derf/y4m/bus_cif.y4m
$ ffmpeg -hide_banner -i bus_cif.y4m -c:v libsvt_av1 -y bus_cif.ivf
Input #0, yuv4mpegpipe, from 'bus_cif.y4m':
  Duration: 00:00:05.00, start: 0.000000, bitrate: 36496 kb/s
    Stream #0:0: Video: rawvideo (I420 / 0x30323449), yuv420p(progressive), 352x288, SAR 128:117 DAR 1408:1053, 30 fps, 30 tbr, 30 tbn, 30 tbc
Stream mapping:
  Stream #0:0 -> #0:0 (rawvideo (native) -> av1 (libsvt_av1))
Press [q] to stop, [?] for help
SVT [version]:  SVT-AV1 Encoder Lib v0.4.0
SVT [build]  :  GCC 4.2.1        64 bit
LIB Build date: Feb 21 2019 09:47:00
-------------------------------------------
Number of logical cores available: 8
Number of PPCS 37
-------------------------------------------
SVT [config]: Main Profile      Tier (auto)     Level (auto)
SVT [config]: EncoderMode                                                       : 1
SVT [config]: EncoderBitDepth / CompressedTenBitFormat                          : 8 / 0
SVT [config]: SourceWidth / SourceHeight                                        : 352 / 288
SVT [config]: Fps_Numerator / Fps_Denominator / Gop Size / IntraRefreshType     : 30 / 1 / 12 / 1
SVT [config]: HierarchicalLevels / BaseLayerSwitchMode / PredStructure          : 3 / 0 / 2
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange                  : CQP / 32 / 17 / 0
-------------------------------------------
Output #0, ivf, to 'bus_cif.ivf':
  Metadata:
    encoder         : Lavf58.20.100
    Stream #0:0: Video: av1 (libsvt_av1) (AV01 / 0x31305641), yuv420p, 352x288 [SAR 128:117 DAR 1408:1053], q=10-48, 7000 kb/s, 30 fps, 30 tbn, 30 tbc
    Metadata:
[...]

PR #89, opened 2019-02-21
Fix crash with nonalignment memory when encode with resolution 864x480

spawlows:
…tion 864x480


jbeich:
I confirm, this fixes #77.

ttrigui:
looks fine to me.

Issue #74, opened 2019-02-20
Debug build fails on Linux with NASM


jbeich:
```
$ cmake -GNinja -DCMAKE_BUILD_TYPE=Debug .
$ ninja
[...]
FAILED: Source/Lib/ASM_SSE2/CMakeFiles/ASM_SSE2.dir/EbPictureOperators_SSE2.asm.o
/usr/bin/nasm -ISource/API -ISource/Lib/Codec -ISource/Lib/C_DEFAULT -ISource/Lib/ASM_SSE2 -DUNIX64 -g null -f elf64 -o Source/Lib/ASM_SSE2/CMakeFiles/ASM_SSE2.dir/EbPictureOperators_SSE2.asm.o Source/Lib/ASM_SSE2/EbPictureOperators_SSE2.asm
null: fatal: more than one input file specified: Source/Lib/ASM_SSE2/EbPictureOperators_SSE2.asm
ninja: build stopped: subcommand failed.
```

hassount:
@jbeich 
Thank you for your feedback.
Ninja is currently not supported with SVT-AV1, we do welcome contributions though for additional support.

jbeich:
@hassount, I've used `Ninja` only because it provides more clear error message. The issue affects `Unix Makefiles` (default) as well. 
```
$ cmake -DCMAKE_BUILD_TYPE=Debug .
$ gmake
[...]
[ 17%] Building ASM_NASM object Source/Lib/ASM_SSE2/CMakeFiles/ASM_SSE2.dir/EbPictureOperators_SSE2.asm.o
null: fatal: more than one input file specified: Source/Lib/ASM_SSE2/EbPictureOperators_SSE2.asm

gmake[2]: *** [Source/Lib/ASM_SSE2/CMakeFiles/ASM_SSE2.dir/build.make:206: Source/Lib/ASM_SSE2/CMakeFiles/ASM_SSE2.dir/EbPictureOperators_SSE2.asm.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:238: Source/Lib/ASM_SSE2/CMakeFiles/ASM_SSE2.dir/all] Error 2
gmake: *** [Makefile:130: all] Error 2

$ gmake VERBOSE=1
[...]
[ 17%] Building ASM_NASM object Source/Lib/ASM_SSE2/CMakeFiles/ASM_SSE2.dir/EbPictureOperators_SSE2.asm.o
[...]

PR #90, opened 2019-02-21
Add debug build to Travis

EwoutH:
Added a debug build on Linux to Travis CI.

Seems to build fine, so #74 is probably about something else.


kylophone:
Thanks.

PR #95, opened 2019-02-22
Add 422/444 support at API level

lijing0010:
- Enable 422/444 in internal data structures
- Down sample to 420 and use 420 for MD in PictureAnalysisKernel
- Enhance check for settings of different profile
- Clean up sample app, remove un-used paddings

### DIFF
--- a/Source/Lib/Common/Codec/EbPictureControlSet.c
+++ b/Source/Lib/Common/Codec/EbPictureControlSet.c
@@ -69,7 +69,9 @@ EbErrorType picture_control_set_ctor(
     uint16_t sb_origin_y;
     EbErrorType return_error = EB_ErrorNone;
 
-    EbBool is16bit = initDataPtr->is16bit;
+    EbBool is16bit = initDataPtr->bit_depth > 8 ? EB_TRUE : EB_FALSE;
+    const uint16_t subsampling_x = (initDataPtr->color_format == EB_YUV444 ? 1 : 2) - 1;
+    const uint16_t subsampling_y = (initDataPtr->color_format >= EB_YUV422 ? 1 : 2) - 1;
 
     EB_MALLOC(PictureControlSet_t*, object_ptr, sizeof(PictureControlSet_t), EB_N_PTR);
 
@@ -77,6 +79,7 @@ EbErrorType picture_control_set_ctor(
     input_picture_buffer_desc_init_data.maxWidth = initDataPtr->picture_width;
     input_picture_buffer_desc_init_data.maxHeight = initDataPtr->picture_height;
     input_picture_buffer_desc_init_data.bit_depth = initDataPtr->bit_depth;
+    input_picture_buffer_desc_init_data.color_format = initDataPtr->color_format;
     input_picture_buffer_desc_init_data.bufferEnableMask = PICTURE_BUFFER_DESC_FULL_MASK;
 
     input_picture_buffer_desc_init_data.left_padding = PAD_VALUE;
@@ -89,6 +92,7 @@ EbErrorType picture_control_set_ctor(
     coeffBufferDescInitData.maxWidth = initDataPtr->picture_width;
     coeffBufferDescInitData.maxHeight = initDataPtr->picture_height;
     coeffBufferDescInitData.bit_depth = EB_16BIT;
+    coeffBufferDescInitData.color_format = initDataPtr->color_format;
     coeffBufferDescInitData.bufferEnableMask = PICTURE_BUFFER_DESC_FULL_MASK;
 
     coeffBufferDescInitData.left_padding = PAD_VALUE;
@@ -98,18 +102,19 @@ EbErrorType picture_control_set_ctor(
 
     coeffBufferDescInitData.splitMode = EB_FALSE;
 
-
     *object_dbl_ptr = (EbPtr)object_ptr;
 
     object_ptr->sequence_control_set_wrapper_ptr = (EbObjectWrapper *)EB_NULL;
 
     object_ptr->recon_picture16bit_ptr = (EbPictureBufferDesc_t *)EB_NULL;
     object_ptr->recon_picture_ptr = (EbPictureBufferDesc_t *)EB_NULL;
+    object_ptr->color_format = initDataPtr->color_format;
 
     EbPictureBufferDescInitData_t coeffBufferDes32bitInitData;
     coeffBufferDes32bitInitData.maxWidth = initDataPtr->picture_width;
     coeffBufferDes32bitInitData.maxHeight = initDataPtr->picture_height;
     coeffBufferDes32bitInitData.bit_depth = EB_32BIT;
+    coeffBufferDes32bitInitData.color_format = initDataPtr->color_format;
     coeffBufferDes32bitInitData.bufferEnableMask = PICTURE_BUFFER_DESC_FULL_MASK;
     coeffBufferDes32bitInitData.left_padding = 0;
     coeffBufferDes32bitInitData.right_padding = 0;
@@ -123,7 +128,7 @@ EbErrorType picture_control_set_ctor(
         (EbPtr)&coeffBufferDes32bitInitData);
 
     // Reconstructed Picture Buffer
-    if (initDataPtr->is16bit == EB_TRUE) {
+    if (is16bit) {
         return_error = eb_recon_picture_buffer_desc_ctor(
             (EbPtr*) &(object_ptr->recon_picture16bit_ptr),
             (EbPtr)&coeffBufferDescInitData);
@@ -143,7 +148,7 @@ EbErrorType picture_control_set_ctor(
 
     // Film Grain Picture Buffer
     if (initDataPtr->film_grain_noise_level) {
-        if (initDataPtr->is16bit == EB_TRUE) {
+        if (is16bit) {
             return_error = eb_recon_picture_buffer_desc_ctor(
                 (EbPtr*) &(object_ptr->film_grain_picture16bit_ptr),
                 (EbPtr)&coeffBufferDescInitData);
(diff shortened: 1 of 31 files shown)


Source/API/EbApi.h
```diff
@@ -228,7 +235,15 @@ typedef struct EbSvtAv1EncConfiguration
...
-
+    /* Specifies the chroma subsampleing format of input video.
+     *
+     * 0 = mono.
+     * 1 = 420.
+     * 2 = 422.
+     * 3 = 444.
+     *
+     * Default is 1. */
+    EB_COLOR_FORMAT          encoder_color_format;
```
hassount: @lijing0010 
can you please use CamelCase for EB_COLOR_FORMAT, EbColorFormat

Source/App/EbAppConfig.c
```diff
@@ -820,6 +824,13 @@ static EbErrorType VerifySettings(EbConfig_t *config, uint32_t channelNumber)
         return_error = EB_ErrorBadParameter;
     }
 
+    //if (config->encoderColorFormat < 1 || config->encoderColorFormat > 3)
+    if (config->encoderColorFormat != 1)
```
hassount: coding style
if (config->encoderColorFormat != 1){

Source/Lib/Codec/EbMcp.c
```diff
@@ -34,9 +34,9 @@ EbErrorType motion_compensation_prediction_context_ctor(
     EB_MALLOC(MotionCompensationPredictionContext_t *, context_ptr, sizeof(MotionCompensationPredictionContext_t), EB_N_PTR);
     *(context_dbl_ptr) = context_ptr;
 #if !EXTRA_ALLOCATION
-    EB_MALLOC(EbByte, context_ptr->avc_style_mcp_intermediate_result_buf0, sizeof(uint8_t)*max_cu_width*max_cu_height * 6 * 3 / 2 + 16, EB_N_PTR);        //Y + U + V;
+    EB_MALLOC(EbByte, context_ptr->avc_style_mcp_intermediate_result_buf0, sizeof(uint8_t)*max_cu_width*max_cu_height * 6 * 3 + 16, EB_N_PTR);        //Y + U + V;
```
hassount: Is this not dependent on the color format ?

Source/API/EbApi.h
lijing0010: OK

Source/App/EbAppConfig.c
lijing0010: OK

Source/Lib/Codec/EbMcp.c
lijing0010: this is a intermediate buffer for 420. if incoming data is 422/444, will overwritten the memory boundary and result in a crash

lijing0010: Hi Hassene
I changed it to color format dependent in the latest commit. Previously I just want to make it simple:)

hassount:
is there a way we could combine this with #135 ?

andrey-norkin:
To combine this PR with #135, we could pull the EbColorFormat enum into #135. 

Currently, in #135, the chroma format is defined in EbAV1StreamInfo as the following trio, in the same was it is signaled in the AV1 sequence header:
    int32_t     subsampling_x;
    int32_t     subsampling_y;
    int32_t     monochrome;

We could also add the  enum EbColorFormat { EB_YUV400, EB_YUV420, EB_YUV422, EB_YUV444}  into the EbFormats.h of #135 and make sure it is aligned with the trio above. We could also return this enum instead of the trio above in the decoder's API EbAV1StreamInfo if we think this is a better approach than signaling this variables in the way they are signaled in AV1 stream (e.g. for unification purpose with other codecs' API).

lijing0010:
> To combine this PR with #135, we could pull the EbColorFormat enum into #135.
> 
> Currently, in #135, the chroma format is defined in EbAV1StreamInfo as the following trio, in the same was it is signaled in the AV1 sequence header:
> int32_t subsampling_x;
> int32_t subsampling_y;
> int32_t monochrome;
> 
> We could also add the enum EbColorFormat { EB_YUV400, EB_YUV420, EB_YUV422, EB_YUV444} into the EbFormats.h of #135 and make sure it is aligned with the trio above. We could also return this enum instead of the trio above in the decoder's API EbAV1StreamInfo if we think this is a better approach than signaling this variables in the way they are signaled in AV1 stream (e.g. for unification purpose with other codecs' API).

How about put it in EbCodec.h, the same place as EbBitDepth

andrey-norkin:
Could be either (the EbFormats.h is included into EbCodec.h). Was thinking to put it into EbFormat.h , since this looks somewhat as the video format description, along with enum aom_chroma_sample_position in EbFormat.h. Actually, maybe the EbBitDepth also belongs there, to the EbFormats.h. Of course, another difference is that things in EbFormats.h do not really affect the decoding (except, possibly, some headers), whereas EbColorFormat does.

lijing0010:
either is OK for me, just keep EbColorFormat together with EbBitDepth is fine

Source/Lib/Codec/EbPictureBufferDesc.c
```diff
@@ -33,6 +33,11 @@ EbErrorType eb_picture_buffer_desc_ctor(
     EbPictureBufferDescInitData_t  *pictureBufferDescInitDataPtr = (EbPictureBufferDescInitData_t*)object_init_data_ptr;
 
     uint32_t bytesPerPixel = (pictureBufferDescInitDataPtr->bit_depth == EB_8BIT) ? 1 : (pictureBufferDescInitDataPtr->bit_depth <= EB_16BIT) ? 2 : 4;
+    const uint16_t subsampling_x = (pictureBufferDescInitDataPtr->color_format == EB_YUV444 ? 1 : 2) - 1;
+
+    //if (pictureBufferDescInitDataPtr->color_format != 1 && pictureBufferDescInitDataPtr->bit_depth != 8) {
```
hassount: Do we need this commented part ?

Source/Lib/Codec/EbSequenceControlSet.h
```diff
@@ -77,13 +77,13 @@ extern "C" {
...
         int32_t                                 cropping_bottom_offset;
-
+ 
         // Conformance Window flag
         uint32_t                                conformance_window_flag;
 
         // Bitdepth
-        EB_BITDEPTH                             input_bitdepth;
-        EB_BITDEPTH                             output_bitdepth;
+        //EB_BITDEPTH                             input_bitdepth;
```
hassount: can we just delete these ?

Source/Lib/Codec/EbPictureBufferDesc.c
lijing0010: no, that's for debug, I'll remove it

Source/Lib/Codec/EbSequenceControlSet.h
lijing0010: I saw a few similar parameters which is confusing(like this xxx_bitdepth) and clean it up a bit.
Left it here in case someone misses it:), I'll just remove it then

lijing0010:
rebase to the latest one

hassount:
@andrey-norkin 
can we merge this patch ?

hassount:
@andrey-norkin
can we merge this patch ?

Source/Lib/Common/Codec/EbPictureControlSet.c
```diff
@@ -925,6 +933,29 @@ EbErrorType picture_parent_control_set_ctor(
     object_ptr->reference_picture_wrapper_ptr = (EbObjectWrapper_t *)EB_NULL;
 
     object_ptr->enhanced_picture_ptr = (EbPictureBufferDesc_t *)EB_NULL;
+    if (initDataPtr->color_format >= EB_YUV422) {
+        EbPictureBufferDescInitData_t input_picture_buffer_desc_init_data;
+        input_picture_buffer_desc_init_data.maxWidth     = initDataPtr->picture_width;
+        input_picture_buffer_desc_init_data.maxHeight    = initDataPtr->picture_height;
+        input_picture_buffer_desc_init_data.bit_depth = 8; //Should be 8bit
```
andrey-norkin: Why is bit depth fixed to 8 bits?

Source/Lib/Common/Codec/EbPictureControlSet.h
```diff
@@ -13958,6 +13960,7 @@ extern "C" {
         EbObjectWrapper_t                    *reference_picture_wrapper_ptr;
         EbObjectWrapper_t                    *pa_reference_picture_wrapper_ptr;
         EbPictureBufferDesc_t                *enhanced_picture_ptr;
+        EbPictureBufferDesc_t                *chroma_downsampled_picture_ptr; //if 422/444 input, down sample to 420 for MD
```
andrey-norkin: Do we have an estimation of how much coding efficiency we would loose by using downsampled chroma in MD?

Source/Lib/Common/Codec/EbPictureControlSet.c
lijing0010: we'll use 420 8bit for the MD

Source/Lib/Common/Codec/EbPictureControlSet.h
lijing0010: The down sampling will only happen for 422/444.
In previous SVT-HEVC, compared to 420, the overall performance drop is around 10%.
(depending on resolution and enc-mode, for M0 the slowest one, perf drop is within 5%)

andrey-norkin: Is if the performance drop in terms of the coding time? What about the BD-rate difference when using 420 8-bit in MD for 444 10-bit input compared to if we used MD on 444 10-bit? 

Or is it 5-10%?

lijing0010: It's the overall performance drop(fps drop), and a big part of the drop is due to coding the extra chroma block(transform, quantization, entropy etc).
We have never done the MD on 444/422 10-bit due to the consideration of performance. 
I think for 420 10bit, we also use 8bit for the MD.

lijing0010: @hassount ，I'm not sure if you have done the comparison for MD on 420 8bit and 420 10-bit before

lijing0010:
rebase to the latest one

PR #106, opened 2019-02-22
Remove calculation of unused variable "sum".

spawlows:
I see increase performance around 6% on all encode in M3 mode.


hassount:
@ttrigui 
Has this been reviewed / tested ?

PR #117, opened 2019-02-23
Travis: Rewrite with Windows and properly reflect macOS build failing

1480c1:
Similar to: <https://github.com/intel/SVT-HEVC/pull/114>
<https://github.com/OpenVisualCloud/SVT-VP9/pull/14>

>Adds a MSVC Release build and a Ninja debug build.
The ninja build is allowed to fail, meaning if the final program fails, it will not mark the whole commit as failing.
>I put it there because I'm interested in using mingw and ninja in the future, but I don't want to have to do the work that travis does on my own computer every update.

One thing I would like to point out with this is that it might be better to set the CFLAG /MP only if MSVC and not if WINDOWS in case in the future if compilers other than MSVC gets supported.

Also add to <https://github.com/OpenVisualCloud/SVT-AV1/issues/4>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,55 +4,86 @@ dist: xenial
 addons:
   apt:
     packages:
-     - cmake
-     - yasm
+      - cmake
+      - yasm
   homebrew:
     packages:
       - yasm
-
-jobs:
+matrix:
+  allow_failures:
+    - os: windows
+    - name: FFmpeg patch
+    - name: macOS build
+  fast_finish: true
   include:
-   # General Linux build job
-   - name: Build
-     script:
-     - cd Build/linux
-     - ./build.sh release
-   # General Linux debug build job
-   - name: Debug build
-     script:
-     - cd Build/linux
-     - ./build.sh debug
-   # General macOS build job
-   - name: macOS build
-     os: osx
-     script:
-     - cd Build/linux
-     - ./build.sh release     
-   # Coveralls test job
-   - name: Coveralls
-     before_install:
-     - pip install --user cpp-coveralls
-     script:
-     - cd Build/linux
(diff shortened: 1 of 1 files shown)


EwoutH:
I would suggest removing the MSVC build, since AppVeyor already does exactly that.

The Ninja build job is a nice addition. We could also consider moving the Ninja build to AppVeyor, since Travis is already slower due to the number of jobs (only 4 are executed simultaneous).

hassount:
@1480c1, thank you for the contribution

> One thing I would like to point out with this is that it might be better to set the CFLAG /MP only if MSVC and not if WINDOWS in case in the future if compilers other than MSVC gets supported.

Agreed, feel free to change it. otherwise we will try to change it when we do a cleanup of the Cmake files

@kylophone 
Any issues with this PR?

1480c1:
> Agreed, feel free to change it. otherwise we will try to change it when we do a cleanup of the Cmake files
Sure, I'll try to change it so that it will work with windows/not using msvc, no promises on how well it will work because I only learned cmake last week.

kylophone:
MacOS should be a supported platform, and not be allowed to fail.

1480c1:
Truly, however, as of right now, all svt-* fails due to cpu_set_t and others.

hassount:
@1480c1 
Thank you for the rewrite, we still should have mac os building and will need to be fixed instead of allowed to fail. I believe #130  is working on addressing this.

1480c1:
That's fine. I'll wait until Mac os can build successfully and rebase with Mac not an allowed failure.

1480c1:
I'm going to close these until MacOS builds are fixed. I'll reopen when they are.

PR #128, opened 2019-02-28
Fix crash with 720x480 yuv420p input

spawlows:
Store with not alignment require.


spawlows:
Fixes [#84](https://github.com/OpenVisualCloud/SVT-AV1/issues/84)
Instead of [PR:88](https://github.com/OpenVisualCloud/SVT-AV1/pull/88)

hassount:
@jstebbins
Could you please confirm that this actually fixes the crash you see on your end ?

jstebbins:
Confirmed, it fixes the crash

PR #133, opened 2019-03-01
Remove unneeded memset() in av1_quantize_inv_quantize().

spawlows:
Increase encode speed ~1.5%


hassount:
@joelsole 
do you have a test case where you see a difference at the output ? because we're not seeing such on our end. 
@spawlows can you check again if you get the same streams before and after ?

spawlows:
@hassount  I checked again with more streams, and different configurations, and I have always that same output. It can depend from encoder configuration, streams or OS.

joelsole:
I did a short test an also got the same output.

PR #135, opened 2019-03-02
Decoder/Encoder API

andrey-norkin:
This branch contains the proposed decoder API. The encoder API has also been changed accordingly.


hassount:
@andrey-norkin thank you for the cleanup and the detailed description. 
On the file names, can we just call them EbSvtAv1.h (for EbCodec.h) and so on for the other files (and enc and dec) that way we won't have much conflicts with the other svt plugins into ffmpeg
I would like to get the input of @jsunintel maintaining the ffmpeg plugin and @tianjunwork for the gstreamer plugin

hassount:
@lijing0010
to review these changes for 422 444 api

tianjunwork:
@andrey-norkin could you also update user guide APIs accordingly?

tianjunwork:
@hassount for AV1 gst/ffmpeg plugin, should decoder plugin be added also?

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+/********************************
+* Defines
+********************************/
+#define EB_PICTURE           uint32_t
+#define EB_B_PICTURE         0
+#define EB_P_PICTURE         1
+#define EB_I_PICTURE         2
+#define EB_IDR_PICTURE       3
+#define EB_NON_REF_PICTURE   4
+#define EB_INVALID_PICTURE   0xFF
```
kylophone: Use an enum here. If a picture can have multiple states, make sure each one of the enum values is or-able.

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+*/
+
+#define EbBool   uint8_t
+#define EB_FALSE  0
+#define EB_TRUE   1
+
+typedef struct EbBufferHeaderType
+{
+    // EbBufferHeaderType size
+    uint32_t size;
```
kylophone: `size_t`

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+
+    // picture (input or output) buffer
+    uint8_t* p_buffer;
+    uint32_t n_filled_len;
+    uint32_t n_alloc_len;
```
kylophone: `n_filled_len` and `n_alloc_len` can be of `size_t` type.

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+
+    // pic private data
+    void*    p_app_private;
+    void*    wrapper_ptr;
+
+    // pic timing param
+    uint32_t n_tick_count;
+    int64_t  dts;
+    int64_t  pts;
```
kylophone: if these don't have an explicit upper-bound, use the type `unsigned`.

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+    uint32_t qp;
+    uint32_t pic_type;
+
+    // pic flags
+    uint32_t flags;
+} EbBufferHeaderType;
+
+typedef struct EbComponentType
+{
+    uint32_t size;
```
kylophone: use `size_t` again here. What does this keep track of?

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+    uint8_t *cr;
+
+    // Hosts LSB 2 bits of 10bit input/output when the compressed 10bit format is used
+    uint8_t *lumaExt;
+    uint8_t *cbExt;
+    uint8_t *crExt;
+
+    uint32_t yStride;
+    uint32_t crStride;
+    uint32_t cbStride;
```
kylophone: use `ptrdiff_t` for all stride values.

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+
+    uint32_t width;
+    uint32_t height;
+
+    uint32_t origin_x;
+    uint32_t origin_y;
```
kylophone: `unsigned` for all values that don't have an explicit upper bound.

Source/API/EbDecApi.h
```diff
@@ -0,0 +1,332 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+#include "stdint.h"
+#include "EbCodec.h"
+#include "EbExtFrameBuf.h"
+
+typedef struct EbAV1OperatingPoint{
```
kylophone: put a space before the `{`. I think all the struct declarations need this.

Source/API/EbDecApi.h
```diff
@@ -0,0 +1,332 @@
...
+    aom_matrix_coefficients_t matrix_coefficients;
+
+    aom_chroma_sample_position_t chroma_sample_position;
+    aom_color_range_t color_range;
+
+    /* Film Grain Synthesis Present */
+    uint32_t    film_grain_params_present;
+
+    /* The stream is in annex_b format */
+    uint32_t    is_annex_b;
```
kylophone: 32 bits to hold a bool?

Source/API/EbCodec.h
andrey-norkin: Essentially, most of the content of EbCodec.h (including the referred one one above) is from the SVT-AV1 encoder API. The intention was to keep the encoder API unchanged at this stage and introduce only changes needed for the decoder API.

Source/API/EbCodec.h
andrey-norkin: The types here belong to the Encoder API and might have been used on purpose. I will let @hassount comment on whether the changes to the types in API are desirable).

Source/API/EbCodec.h
andrey-norkin: Same as above (@hassount )

Source/API/EbDecApi.h
andrey-norkin: Absolutely. Two reasons (as for the some other "bool" types above): 
1. Keep the data in the structure aligned with 32-bit words. 
2. If char is used, it needs to be converted to int for comparisons anyway.

Source/API/EbCodec.h
andrey-norkin: These values have upper bound. This was the SVT-AV1 input format with these types. @hassount, please comment if there is a rational for keeping the current types.

Source/API/EbCodec.h
andrey-norkin: Do we want to use ptrdiff_t on the API?

Source/API/EbDecApi.h
andrey-norkin: It does not matter for a compiler. But I will put it for better looks.

Source/API/EbCodec.h
kylophone: it's part of `stdef.h` since C99.

Source/API/EbDecApi.h
kylophone: > Absolutely. Two reasons (as for the some other "bool" types above):
> 
> 1. Keep the data in the structure aligned with 32-bit words.
> 2. If char is used, it needs to be converted to int for comparisons anyway.

OK, I buy that. (How cache-sensitive is this data?) My suggestion would be to just use `int` and let the compiler choose the right size for the platform. Should already be 32-bits for most platforms.

Source/API/EbCodec.h
andrey-norkin: It is. But does it assume any length of the field or is it whatever the system defines?

Source/API/EbCodec.h
```diff
@@ -0,0 +1,156 @@
...
+********************************/
+#define EB_PICTURE           uint32_t
+#define EB_B_PICTURE         0
+#define EB_P_PICTURE         1
+#define EB_I_PICTURE         2
+#define EB_IDR_PICTURE       3
+#define EB_NON_REF_PICTURE   4
+#define EB_INVALID_PICTURE   0xFF
+
+/** The EbBool type is intended to be used to represent a true or a false
```
hassount: we could remove this comment

Source/API/EbCodec.h
hassount: This could be removed

Source/API/EbEncApi.h
```diff
@@ -210,12 +109,14 @@ typedef struct EbSvtAv1EncConfiguration
      *
```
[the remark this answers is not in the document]
hassount: Good point on changing the names of the picture types to align with av1 - vp9 types. 
The eb_svt_enc_eos_nal() is a call that would return the closing packet of the bitstream ahead of the encode time. It si needed mainly by Gstreamer for HEVC. If it's not needed here, we can remove it. Currently this call is not implemented yet.
The EB_BUFFERFLAG_EOS on the other side signals to the library that this is the last picture it's going to get.

Source/API/EbCodec.h
andrey-norkin: Combined into an enum

kylophone: Add `EB_ALT_REF_PICTURE` too?

Source/API/EbEncApi.h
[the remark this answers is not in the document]
andrey-norkin: CRA should be EB_AV1_FW_KEY_PICTURE

PR #138, opened 2019-03-04
[ffmpeg_plugin] Sync pkg-config include directory

jbeich:
pkg-config is supposed to provide all the flags required to build against this library. Regressed by #86.


jbeich:
Obsolete after c091bdb0ff95.

PR #142, opened 2019-03-05
Clangformat

edmond-zhu:
[Edmond]1. Add clang-format configuration file in test.


edmond-zhu:
commit in a wrong branch and close it.

PR #152, opened 2019-03-06
ois based Intra precition

okhlif:
Upgrade OIS to AV1
OIS based intra prediction in M7 and higher
Changed Intra prediction settings for M0 and M1


okhlif:
Sorry wrong branch

PR #162, opened 2019-03-07
Port to FreeBSD and maybe more

jbeich:
Thanks to POSIX many less known Unix systems are mostly compatible with Linux: Solaris, DragonFly, FreeBSD, NetBSD, OpenBSD. Only special systems like Android or Fuchsia may need a lot of platform-specific code. In general, unless you're poking under `/proc` or use `_np` functions the code may work on other POSIX systems.

Depends on #130 to avoid merge conflicts.

Example build:
```
$ pkg install cmake yasm
$ cmake -DCMAKE_ASM_NASM_COMPILER=yasm .
$ make -j$(getconf NPROCESSORS_ONLN)
```
or (recommended):
```
$ pkg install cmake nasm ninja
$ cmake -GNinja .
$ ninja
```

Example run:
```
$ fetch https://media.xiph.org/video/derf/y4m/bus_cif.y4m
$ ./Bin/SvtAv1EncApp -i bus_cif.y4m -b bus_cif.ivf
```


Source/App/EbTime.c
```diff
@@ -10,34 +10,29 @@
...
 #endif
 
 void EbStartTime(
     uint64_t *Startseconds,
     uint64_t *Startuseconds)
 {
 
-#if defined(__linux__) || defined(__APPLE__) //(LINUX_ENCODER_TIMING || LINUX_DECODER_TIMING)
+#ifdef _WIN32 //(WIN_ENCODER_TIMING || WIN_DECODER_TIMING)
+    *Startseconds = (uint64_t) clock();
```
jbeich: Such code churn can be avoided at the expense of readability e.g.,
```c
#ifndef _WIN32
// Linux & macOS code
#else
// Windows code
#endif
```

jbeich:
Builds and runs ([almost](http://ix.io/1AvY)) fine on NetBSD as well.

kylophone:
#130 is merged.

hassount:
@jbeich 
Thank you for the contribution on this merge this contribution in but we can not at this point officially support FreeBSD. so Could you please remove the support from the Readme file ?

kylophone:
I'd like to merge this since it greatly improves Unix compatibility. I'll do so in the next few days, but I will update README.md to address the comments above. OK, @jbeich?

CMakeLists.txt
```diff
@@ -18,35 +18,32 @@ set(CAN_USE_ASSEMBLER TRUE)
 set(CMAKE_NASM_LINK_EXECUTABLE "ld -o <TARGET> <OBJECTS>")
 set(CMAKE_ASM_NASM_SOURCE_FILE_EXTENSIONS asm)
 
-
-if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
```
1480c1:
```suggestion
if(WIN32)
```

CMakeLists.txt
```diff
@@ -18,35 +18,32 @@ set(CAN_USE_ASSEMBLER TRUE)
...
+    set(CMAKE_EXE_LINKER_FLAGS " ")
+    set(CMAKE_SHARED_LINKER_FLAGS " ")
+    set(CMAKE_C_FLAGS_RELEASE "-O2")
+    set(CMAKE_ASM_NASM_FLAGS "-DWIN64")
+    set(CMAKE_ASM_NASM_FLAGS_DEBUG "")
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
1480c1:
```suggestion
elseif(APPLE)
```

CMakeLists.txt
```diff
@@ -18,35 +18,32 @@ set(CAN_USE_ASSEMBLER TRUE)
...
-endif (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-
-if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
-    set(CMAKE_EXE_LINKER_FLAGS " ")
-    set(CMAKE_SHARED_LINKER_FLAGS " ")
-    set(CMAKE_C_FLAGS_RELEASE "-O2")
-    set(CMAKE_ASM_NASM_FLAGS "-DWIN64")
-    set(CMAKE_ASM_NASM_FLAGS_DEBUG "")
-endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+endif (${CMAKE_SYSTEM_NAME} MATCHES "Windows")
```
1480c1:
```suggestion
endif (WIN32)
```

CMakeLists.txt
```diff
@@ -18,35 +18,32 @@ set(CAN_USE_ASSEMBLER TRUE)
...
     set(CMAKE_MACOSX_RPATH 1)
     set(CMAKE_C_ARCHIVE_CREATE   "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
     set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
     set(CMAKE_C_ARCHIVE_FINISH   "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
     set(CMAKE_CXX_ARCHIVE_FINISH "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
-endif (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-
-if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-    set(CMAKE_C_FLAGS "-Wall -Wextra -Wformat -Wformat-security -mavx -fstack-protector-strong -fPIE -fPIC -D_FORTIFY_SOURCE=2 ")
+else()
```
1480c1: Probably best to not gate the cflags to systems other than WIN and APPLE, as then they will not be applied for macOS and mingw etc.

CMakeLists.txt
jbeich: The existing style is more popular in this project:
```
$ rg --sort-files -t cmake 'if.*WIN32'
Source/App/DecApp/CMakeLists.txt
35:if (MSVC OR MSYS OR MINGW OR WIN32)

Source/App/EncApp/CMakeLists.txt
35:if (MSVC OR MSYS OR MINGW OR WIN32)

test/CMakeLists.txt
27:if (MSVC OR MSYS OR MINGW OR WIN32)
62:if (MSVC OR MSYS OR MINGW OR WIN32)
72:endif(MSVC OR MSYS OR MINGW OR WIN32)

third_party/googletest/CMakeLists.txt
18:if (MSVC OR MSYS OR MINGW OR WIN32)
```
vs.
```
$ rg --sort-files -t cmake 'if.*Windows'
CMakeLists.txt
21:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
46:endif (${CMAKE_SYSTEM_NAME} MATCHES "Windows")

Source/Lib/Common/ASM_SSE2/CMakeLists.txt
14:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
16:endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")

Source/Lib/Common/ASM_SSE4_1/CMakeLists.txt
17:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
19:endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")

Source/Lib/Common/ASM_SSSE3/CMakeLists.txt
17:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
19:endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")

Source/Lib/Common/C_DEFAULT/CMakeLists.txt
13:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
15:endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")

Source/Lib/Common/Codec/CMakeLists.txt
16:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
18:endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")

Source/Lib/Decoder/Codec/CMakeLists.txt
29:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
31:endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")

Source/Lib/Encoder/Codec/CMakeLists.txt
29:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
31:endif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
42:if(${CMAKE_SYSTEM_NAME} MATCHES "Windows|Darwin")
```

CMakeLists.txt
1480c1:
```suggestion
endif()

if(CMAKE_C_COMPILER_ID MATCHES "GNU|Clang")
```

CMakeLists.txt
1480c1: Indeed, however I am against using `${CMAKE_SYSTEM_NAME} MATCHES "Windows"` simply because it will match CYGWIN, MINGW, and MSVC compilers, which although in this context, is okay since it's only setting NASM flags. However, imo it would be better to use predefined variables for cmake so that way they can be properly written to accommodate the differences in the compilers.

CMakeLists.txt
jbeich: `-DUNIX64` and `elf64` cannot be used on Windows.

1480c1: Yes, so
``` cmake
if(UNIX)
    set(CMAKE_ASM_NASM_FLAGS "-DUNIX64")
    if(APPLE)
        set(CMAKE_ASM_NASM_OBJECT_FORMAT macho64)
    else()
        set(CMAKE_ASM_NASM_OBJECT_FORMAT elf64)
    endif()
elseif(WIN32)
    set(CMAKE_ASM_NASM_FLAGS "-DWIN32")
endif()
```
IDK of the other formats off the top of my head

1480c1: Only set the cflags for the if clang or gnu, with the rest of the OS specific options under their own section

jbeich: Some CFLAGS can probably be dropped in favor of [CMake defaults](https://github.com/Kitware/CMake/blob/v3.14.1/Modules/Compiler/GNU.cmake#L49-L54).

1480c1: Since I can't really suggest anything outside of one line suggestions, I've put this here tentatively.

1480c1: Probably

jbeich:
Anything else? `__unix__` conditionals maybe ugly but avoid assuming stuff on Fuchsia, CloudABI, Emscripten and other weird targets. ;)

1480c1:
How about switching around the ifdefs to`_WIN32`, else instead of `__unix__ || __APPLE__` else `_WIN32` and let those who want to try other targets to go ahead, but offer no official support? (I'm not sure if SVT-AV1 would be good on Fuchsia and on a phone, but it would be interesting to see)

jbeich:
Done. Required moving code around (error-prone) to avoid the following:
```
#if !_WIN32
...
#elif _WIN32
...
#endif
```

Source/App/EncApp/EbTime.c
```diff
@@ -12,135 +12,109 @@
...
 void ComputeOverallElapsedTimeMs(uint64_t Startseconds, uint64_t Startuseconds, uint64_t Finishseconds, uint64_t Finishuseconds, double *duration)
 {
-#if defined(__linux__) || defined(__APPLE__) //(LINUX_ENCODER_TIMING || LINUX_DECODER_TIMING)
+#ifdef _WIN32 //(WIN_ENCODER_TIMING || WIN_DECODER_TIMING)
+    //double  duration;
+    *duration = (double)(Finishseconds - Startseconds);
+    //printf("\nElapsed time: %3.3f seconds\n", *duration);
+    (void)(Startuseconds);
+    (void)(Finishuseconds);
+#els //(LINUX_ENCODER_TIMING || LINUX_DECODER_TIMING)
```
1480c1:
```suggestion
#else //(LINUX_ENCODER_TIMING || LINUX_DECODER_TIMING)
```

jbeich: Done and rebased to check new code.

AlexBoswellVCD:
Found a 2% speed gain averaged over M1,M4, and M8. No memory deviations across these modes. No loss when comparing M4 and M8, there was a run-to-run issue in M1 in Master using a 360p clip so could not validate that the changes are lossless in this context.

EwoutH:
@AlexBoswellVCD Could you report standard deviation of your test results? Or 95% confidence intervals? This PR shouldn't improve performance, only compatibility.

hassount:
> @AlexBoswellVCD Could you report standard deviation of your test results? Or 95% confidence intervals? This PR shouldn't improve performance, only compatibility.

@EwoutH agreed, @AlexBoswellVCD this is more of a r2r speed difference let's stick to a 5% bar otherwise it's a similar speed level

PR #164, opened 2019-03-08
test

Foch123:


hassount:
@Foch123 
Is this a mistaken PR ? if not can we have a better description on what is this PR supposed to add?

EwoutH:
Looks like rewritten CI build scripts (that should properly output exit codes when failing).

1480c1:
That would be nice

1480c1:
Although a better pr name would be better

citest/cpplint_test.cpp
```diff
@@ -0,0 +1,11 @@
+nclude <iostream>
```
joelsole: typo

joelsole:
I am going to close this PR, unless there are opinions on leaving it open

Issue #73, opened 2019-02-20
API/ABI conflict with other SVT-* encoders


jbeich:
FFmpeg cannot build against more than one SVT-* encoder:
- `<Eb*.h>` have same header guards, so including one prevents others
- SVT-AV1 and SVT-VP9 share snake_case functions e.g., `eb_init_encoder`
- SVT-AV1 and SVT-HEVC share CamelCase functions e.g., `EbStartTime`
- ELF symbols are GLOBAL by default in shared libraries, so first binding wins

For example,
```
$ LD_PRELOAD=/usr/lib/libSvtVp9Enc.so ffmpeg -hide_banner -framerate 3 -f x11grab -i :0 -to 1 -c:v libsvt_av1 -y foo.avi
Input #0, x11grab, from ':0':
  Duration: N/A, start: 1550662447.185332, bitrate: N/A
    Stream #0:0: Video: rawvideo (BGR[0] / 0x524742), bgr0, 640x480, 3 fps, 3.08 tbr, 1000k tbn, 1000k tbc
Stream mapping:
  Stream #0:0 -> #0:0 (rawvideo (native) -> av1 (libsvt_av1))
Press [q] to stop, [?] for help
SVT [version]:  SVT-AV1 Encoder Lib v0.4.0
SVT [build]  :  GCC 4.2.1        64 bit
LIB Build date: Feb 20 2019 10:06:20
-------------------------------------------
Segmentation fault
[...]
Thread 1 received signal SIGSEGV, Segmentation fault.
0x000000080071a482 in eb_svt_enc_set_parameter (svt_enc_component=0x8013ffc00,
    p_component_parameter_structure=0x8084ed008)
    at SVT-VP9-2437369/Source/Lib/Codec/EbEncHandle.c:2440
2440        scs_ptr             = p_enc_comp_data->sequence_control_set_instance_array[0]->sequence_control_set_ptr;
(gdb) bt
#0  0x000000080071a482 in eb_svt_enc_set_parameter (svt_enc_component=0x8013ffc00,
    p_component_parameter_structure=0x8084ed008)
    at SVT-VP9-2437369/Source/Lib/Codec/EbEncHandle.c:2440
#1  0x000000080206f29f in eb_enc_init (avctx=0x808471e00) at libavcodec/libsvt_av1.c:251
[...]

PR #166, opened 2019-03-08
Allow inclusion of both SVT-AV1 and SVT-HEVC headers

jbeich:
Currently, both SVT-AV1 and SVT-HEVC use the same header guards. Including one skips the other.
Required by #73.

```c
$ cat a.c
#include <EbApi.h>
#include <EbSvtAv1Enc.h>

$ cc -E a.c $(pkg-config SvtAv1Enc SvtHevcEnc --cflags) | fgrep EbSvtAv1EncConfiguration

$ cat b.c
#include <EbSvtAv1Enc.h>

$ cc -E b.c $(pkg-config SvtAv1Enc SvtHevcEnc --cflags) | fgrep EbSvtAv1EncConfiguration
typedef struct EbSvtAv1EncConfiguration
} EbSvtAv1EncConfiguration;
        EbSvtAv1EncConfiguration *config_ptr);
        EbSvtAv1EncConfiguration *pComponentParameterStructure);
```

### DIFF
--- a/Source/API/EbSvtAv1.h
+++ b/Source/API/EbSvtAv1.h
@@ -3,8 +3,8 @@
 * SPDX - License - Identifier: BSD - 2 - Clause - Patent
 */
 
-#ifndef EbCodec_h
-#define EbCodec_h
+#ifndef EbSvtAv1_h
+#define EbSvtAv1_h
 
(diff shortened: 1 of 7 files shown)


mafonsoN:
Lgmt

PR #181, opened 2019-03-16
Encoder-common-decoder setup

andrey-norkin:
This PR splits the library code into three parts: encoder-decoder-common
The build produces two dynamic libraries: SvtAv1Enc and SvtAv1Dec, each includes Common libraries (on request from Intel). 

An example decoder application and decoder library are added (the decoder is not functional yet). 

This PR does not disentangle the encoder-common code, only small part of the code is moved to the encoder library, the rest is still in the common library. It is planned to further separate the code in the future PRs.


hassount:
@jsunintel @tianjunwork 
Can you please review this PR and let us know if there's anything that would conflict with the ffmpeg / gstreamer integrations.

Source/Lib/Common/ASM_AVX2/CMakeLists.txt
```diff
@@ -8,13 +8,13 @@
 # Include Encoder Subdirectories
 include_directories(${PROJECT_SOURCE_DIR}/Source/API/)
 include_directories(${PROJECT_SOURCE_DIR}/Source/API/OpenMAX/IL/)
```
tianjunwork: please clean up all openmax in cmakelists.

tianjunwork:
> @jsunintel @tianjunwork
> Can you please review this PR and let us know if there's anything that would conflict with the ffmpeg / gstreamer integrations.

@hassount this PR doesn't affect encoder api. So there is no change for gstreamer integration. Is there a requirement for decoder plugin?

jsunintel:
> 
> 
> @jsunintel @tianjunwork
> Can you please review this PR and let us know if there's anything that would conflict with the ffmpeg / gstreamer integrations.

Do we need to expose the SVT AV1 decode functionalities in ffmpeg and gstreamer?

hassount:
@jsunintel @tianjunwork 
As the decoder is just starting, let's focus for now on the encoder API / installation. If none of you sees an issue with this PR on the current gstreamer / ffmpeg then we're good.

andrey-norkin:
> shouldn't EbTime.c also be used by the DecApp?

It depends on if the decoder developers decide to use these functions to measure time. Do you propose creating a CommonApp directory as well? It may be an overkill if there is just one file there. Of course, there could also be some common command line parameters handling. Maybe let us wait if there is a need to do this, we can always add another directory later.

Source/Lib/Common/ASM_AVX2/CMakeLists.txt
andrey-norkin: Done

Source/Lib/Common/ASM_AVX2/CMakeLists.txt
```diff
@@ -7,7 +7,6 @@
 
 # Include Encoder Subdirectories
 include_directories(${PROJECT_SOURCE_DIR}/Source/API/)
```
tianjunwork: There are "openmax" in multiple CMakeLists.txt. Could you clean all of them?

andrey-norkin: Could you give me an example, in which files you can still find them?

tianjunwork: I did a grep, but just noticed, they are in Build/ folder. We are fine here:)

PR #187, opened 2019-03-20
AppVeyor: Don't upload & deploy unnecessary artifacts

EwoutH:
Stops AppVeyor from uploading the gtest and .exp artifacts after #181.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,12 +12,13 @@ build:
 
 artifacts:
     - path: bin\Release\SvtAv1*.exe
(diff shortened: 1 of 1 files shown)


EwoutH:
Stops publishing the .exp, gtest and .lib files after the encoder-decoder-common split. If anyone uses those files and want them published please indicate so.

New (pre-)release will now look [like this](https://github.com/EwoutH/SVT-AV1/releases/tag/0.0.1-test4).

PR #188, opened 2019-03-20
Print out clang version instead of gcc

1480c1:
It's been slightly confusing when it would show gcc 4.2.1 even though it was clang.

Added a distinguisher between Apple, mingw, msvc, linux, and others.
Idk what the identifiers for other types would be.

### DIFF
--- a/Build/linux/build.sh
+++ b/Build/linux/build.sh
@@ -24,7 +24,7 @@ function build {
         -DCMAKE_ASM_NASM_COMPILER=$CMAKE_ASSEMBLER  \
 
     # Compile the Library
-    make -j $(nproc) SvtAv1EncApp SvtAv1DecApp SvtAv1UnitTests
+    make -j $(if [ "$(uname -s)" = "Darwin" ]; then sysctl -n hw.ncpu; else nproc; fi) SvtAv1EncApp SvtAv1DecApp SvtAv1UnitTests
     cd ..
 }
 
@@ -43,7 +43,7 @@ else
     CMAKE_COMPILER=$ICC_COMPILER
 fi
 
-cd $(dirname $(realpath $0))
+cd $(dirname $(if [ "$(uname -s)" = "Darwin" ]; then perl -e 'use Cwd "abs_path";print abs_path(shift)' $0; else realpath $0; fi))
 
 if [ $# -eq 0 ]; then
     build Debug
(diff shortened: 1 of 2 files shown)


1480c1:
Going to switch to define later to see if that's better, ignore for now

Source/Lib/Encoder/Codec/EbEncHandle.c
```diff
@@ -3368,10 +3368,29 @@ EbErrorType init_svt_av1_encoder_handle(
...
+#define clangtype "Linux"
+#else
+#include <sys/param.h>
+#ifdef BSD
+#define clangtype "BSD"
+#else
+#define clangtype "Generic"
+#endif
+#endif
+    printf("SVT [build]  :\t%s Clang %d.%d.%d\t", clangtype, __clang_major__,
```
jbeich: Better simplify GCC line by using `__VERSION__` e.g.,
```
$ cc -dM -E -</dev/null | fgrep -i version
#define __VERSION__ "4.2.1 Compatible FreeBSD Clang 8.0.0 (tags/RELEASE_800/final 356365)"
#define __clang_version__ "8.0.0 (tags/RELEASE_800/final 356365)"

$ gcc8 -dM -E -</dev/null | fgrep -i version
#define __VERSION__ "8.2.0"
```
```c
 #elif defined(__GNUC__)
    printf("SVT [build]  :\tGCC %s\t", __VERSION__);
```

Build/linux/build.sh
```diff
@@ -24,7 +24,7 @@ function build {
         -DCMAKE_ASM_NASM_COMPILER=$CMAKE_ASSEMBLER  \
 
     # Compile the Library
-    make -j $(nproc) SvtAv1EncApp SvtAv1DecApp SvtAv1UnitTests
+    make -j $(if [ "$(uname -s)" = "Darwin" ]; then sysctl -n hw.ncpu; else nproc; fi) SvtAv1EncApp SvtAv1DecApp SvtAv1UnitTests
```
jbeich: If `make` is a GNU make then `-j` without number is another option:
> If the `-j` option is given without an argument, `make` will not limit the number of jobs that can run simultaneously.

Source/Lib/Encoder/Codec/EbEncHandle.c
1480c1: I was tempted to do so, however I wasn't sure if it would be a good Idea to include such a long line.
I'll change it since it does seem cover everything.

kylophone:
Ah, thanks. I'll merge this. In the future, try to avoid mixing clean-ups into your patches.

kylophone:
> Ah, thanks. I'll merge this. In the future, try to avoid mixing clean-ups into your patches.

Not saying that clean-ups aren't important, but it makes the focus of the diff harder to understand.

Issue #122, opened 2019-02-28
Add Valgrind Tests to Travis


kylophone:
@EwoutH, any thoughts on this? Would like to check for leaks in CI.

EwoutH:
Yes that should definitely be done. I will look into it tomorrow.

PR #189, opened 2019-03-21
Travis: Add valgrind testing, rewrite with clang and gcc and xenial and macOS

1480c1:
Adds valgrind+linux testing, macOS valgrind is not properly working yet
Adds clang and gcc variants to linux
Adds ccache so that way builds don't take over two hours
Adds testing of SvtAv1EncApp, ffmpeg plugin, and SvtAv1EncApp with stdin input
Allow for the debug build to fail since basic testing is already in the release builds
   - Also because if not, it would take ~1-2 hours because of valgrind(~40 minutes alone for each) and ffmpeg builds
   - Base build (Release and ffmpeg integration) is about 10 minutes after second run (needs to build cache first)

Todo: 
 - see how to integrate gtest into travis. I had some problems with it when using `--coverage` in the flags.
 - Maybe add mingw builds. Might be better on appveyor since their support for windows might be better.


kylophone:
I'm fine with this. @EwoutH, do you have any thoughts on this?

kylophone:
This PR addresses issue #122

1480c1:
Will rebase and make sure it works still once I get to an actual computer. Right now on my phone

EwoutH:
Additions look nice!

I personally prefer using jobs instead of if statements as much as possible.

Maybe add stages, that you only run dcoverage and valgrind if the build jobs pass.

1480c1:
Sure, I can do that. I'm right now trying to figure out why the gcc+linux+coveralls is failing.

1480c1:
Found out that the perms for some ccache files were wrong and that didn't allow ffmpeg to compile. Found a fix.

1480c1:
I will try to see what I can do in the morning

1480c1:
Does anyone have any interlaced video that I can use as an example test for interlaced source for the coverall jobs?

kylophone:
> Does anyone have any interlaced video that I can use as an example test for interlaced source for the coverall jobs?

I wouldn't worry about it. I would also argue for removing that functionality (interlaced sources) from the encoder.

kylophone:
@1480c1, can you look at this failing Travis job?: https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/188764775#L136

1480c1:
> @1480c1, can you look at this failing Travis job?: https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/188764775#L136

Sure, I think the reason why is that I put the wrong group for changing the ownership of the ccache. I saw somewhere that it was travis:travis, but I guess it is wrong.

kylophone:
> > @1480c1, can you look at this failing Travis job?: https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/188764775#L136
> 
> Sure, I think the reason why is that I put the wrong group for changing the ownership of the ccache. I saw somewhere that it was travis:travis, but I guess it is wrong.

Thanks! I'll merge this as soon as all the red goes away.

1480c1:
@kylophone Do you think it would be better to test using gcc-8 and clang-8, or use the default gcc 5.4 and clang 6?

kylophone:
> @kylophone Do you think it would be better to test using gcc-8 and clang-8, or use the default gcc 5.4 and clang 6?

Most recent stable version of each should be preferred.

1480c1:
Can I just merge the test encode job into the release builds?

1480c1:
@kylophone also, can the stop build if there is a new commit be added to the Travis-ci.com site? I see it on org, but not com

PR #190, opened 2019-03-21
travis: add a small encode/decode smoke test

kylophone:


1480c1:
I think that libaom may need to be build from source because it does not seem to be available for xenial

.travis.yml
```diff
@@ -15,13 +15,12 @@ jobs:
    # General Linux build job
    - name: Build
      before_install:
-     - sudo apt-get install -y libaom-dev valgrind
+     - sudo apt-get install -y valgrind
      script:
      - cd Build/linux
      - ./build.sh release
      - wget -nc https://mf4.xiph.org/~ltrudeau/videos/nyan.y4m
      - valgrind --leak-check=full ../../Bin/Release/SvtAv1EncApp -i nyan.y4m -enc-mode 0 -lp 1 -q 24 -o out.yuv -b out.ivf
```
hassount: + this looks good, it might be a little too slow though, could we maybe run -enc-mode 4 or 5 ?

1480c1:
Maybe it would be better to enable the auto stop builds if there is an overriding commit for travis so that way there isn't a backlog of builds.
image

1480c1:
@kylophone You may need to add `--no-check-certificate` to the wget command since it seems that wget thinks that xiph's cert is invalid

.travis.yml
joelsole: Fewer frames should be enough.  Somewhere from -n 2  to  -n 8.

Issue #186, opened 2019-03-19
Clang: Segfault with SimpleApp


1480c1:
`lldb -- ./SvtAv1EncSimpleApp bus_cif.y4m test3.mkv 352 288 8`
```
Process 35714 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x8)
    frame #0: 0x0000000100001b52 SvtAv1EncSimpleApp`InitEncoder(config=0x0000000100c00270, callbackData=0x0000000100c003f0, instanceIdx=0) at EbSimpleAppContext.c:218
   215          if (return_error != EB_ErrorNone) {
   216              return return_error;
   217          }
-> 218          fwrite(outputStreamBuffer->p_buffer, 1, outputStreamBuffer->n_filled_len, config->bitstreamFile);
   219      }
   220      ///************************* LIBRARY INIT [END] *********************///
   221      return return_error;
```
also at
<https://travis-ci.com/1480c1/SVT-AV1/jobs/186012440#L4196>

1480c1:
Happens on clang+linux too <https://travis-ci.com/1480c1/SVT-AV1/jobs/186074128#L3593>
going to see if I can get a report from a linux machine.

1480c1:
`gdb --args ./SvtAv1EncSimpleApp bus_cif.y4m test3.mkv 352 288 8`
```
Program received signal SIGILL, Illegal instruction.
whole_frame_rect (cm=0x5555561d2100, is_uv=0) at /home/ccom1/Desktop/SVT-AV1/Source/Lib/Codec/EbRestoration.c:184
184         return rect;
```

hassount:
@1480c1 
Thank you for the report. it seems that the simple app is out of synch. I am curious as you are using it if you see a value in having it around or we can just remove it.

1480c1:
I honestly see no value in it since everything that the simple app provides is already in the main app. I'm just using it because I'm trying to test coveralls.

1480c1:
Probably best to just remove it since it's like an appendix and will probably be ignored and unused

hassount:
@1480c1 
Thank you for the feedback. agreed.
@joelsole @andrey-norkin @kylophone @mafonsoN @lijing0010 
Any issues with removing the simple app ?

joelsole:
No issues

kylophone:
> @1480c1
[...]

PR #197, opened 2019-03-27
Remove SimpleApp

kylophone:
This removes SimpleApp. The reason for the PR is discussed in #186.


1480c1:
Will this be replicated across the encoders?

hassount:
@1480c1 
very likely :)

PR #215, opened 2019-04-06
Travis: Use github for video mirror, speed up mac build, no valgrind if pr

1480c1:
should speed up builds by ~5 minutes or more. Should also prevent the error of connecting to xiph causing the build to fail.

Should the valgrind job run for pr's? it seems to take a long while and if there are multiple prs at one time, it can clog up the backlog.


kylophone:
lgtm. It might be better to host the files in a OpenVisualCloud github.io.

kylophone:
Let's use this URL.

https://github.com/OpenVisualCloud/SVT-AV1-Resources/raw/master/video.tar.gz

PR #216, opened 2019-04-07
Use prefix for asm files using AOM's method, add a note about macOS fd limit

### LINKED ISSUE #178: Semaphore issue on MacOS
This error happens very often in Mac OS even with the latest commit f2a87ad. Not sure where the problem is. Any easy fix?

```
errno: 28
Encoding          
Could not allocate enough memory for channel 1
```

1480c1:
Asm prefix part taken from <https://aomedia.googlesource.com/aom/+/master/aom_ports/x86_abi_support.asm#95>.

Named semaphores must be unlinked else they will be orphaned but still contribute to the global semaphore count. As long as they are unlinked after being opened, they should not be orphaned. 
<https://bugs.chromium.org/p/nativeclient/issues/detail?id=686>

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ The SVT-AV1 Encoder adapts to the system that is being ran on. The memory requir
 * __Build Instructions__
 	 -	./Build/linux/build.sh <release | debug> (if none specified, both release and debug will be built)
 
+  - __Note about macOS__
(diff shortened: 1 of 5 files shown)


EwoutH:
@hassount Could you make sure this decently sized PR that has been open for over 6 weeks actually gets reviewed?

1480c1:
Sure

1480c1:
done

PR #220, opened 2019-04-09
WIP: Exclude documentation from coverage tests

EwoutH:
Exclude Docs directory from Coveralls.

There is probably more that should be excluded, feel free to leave suggestions.


EwoutH:
Docs is already left out, only the Source directory is included.

1480c1:
If there is a commit dealing with docs and nothing related to code, travis
can be skipped by adding `[ skip ci ]` somewhere.

PR #227, opened 2019-04-17
Fix intra_edge_filter problem for 4x4 blocks

lijing0010:
Otherwise for 4x4 chroma case, while doing checking "is_smooth()", may
get the wrong filter_type


hassount:
@lijing0010 can you confirm and close ?

lijing0010:
yeah, it's in 235

PR #229, opened 2019-04-19
Replace with the correct mutex

lijing0010:


lijing0010:
any comments on this PR? Should be a straightforward typo in original code

PR #230, opened 2019-04-23
Add switches for batch file for windows. Add VS2019

1480c1:
Renamed generate_vs2017 to generate and added switches for using different cmake generators.
Also only define char_bit if not defined (one of mingw's header already defines it as 8)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,11 +11,11 @@
 *.bin
 *.opensdf
 *.log
-Build/linux/debug/
-Build/linux/release/
-Build/VS/x64
+.vscode
 Bin
+Build
 build
 test/vectors/
 test/docs/
-.vscode
+!Build/linux/build.sh
+!Build/windows/build.bat
--- a/Build/windows/build.bat
+++ b/Build/windows/build.bat
@@ -1,0 +1,124 @@
+:: Copyright(c) 2019 Intel Corporation
+:: SPDX-License-Identifier: BSD-2-Clause-Patent
+@echo off
+
+setlocal
+cd /d "%~dp0"
+set instdir=%CD%
+
+set "build=y"
+if NOT -%1-==-- call :args %*
+if exist CMakeCache.txt del /f /s /q CMakeCache.txt 1>nul
+if exist CMakeFiles rmdir /s /q CMakeFiles 1>nul
+if NOT "%GENERATOR%"=="" set GENERATOR=-G"%GENERATOR%"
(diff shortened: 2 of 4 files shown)


1480c1:
Would it be better to name it to build.bat to match the linux script?

kylophone:
I don't have any comments on the patch (I'll leave that to a Windows dev), but try not to mix clean-ups with actual code changes. It makes the patches harder to read.

1480c1:
I'm not entirely sure I understand what you mean and so can you please elaborate? Do you mean separate out the changes, or something else? If it's that, I can do that, but I usually only change commits at the end when I am at least certain that I would not make any more changes so I try to force-push once.

Also, is there a list of who primarily works on what? I tried looking at the git history to see who works on certain code but for most of the windows specific ones, the history ends at the initial commit

EwoutH:
@1480c1 Can you resolve conflicts and rebase?

@hassount Can you review or assign reviewers?

1480c1:
Sure

1480c1:
Done. As @kylophone said, I'm going to split the formatting issues to another pr in the future

hassount:
@1480c1 
Thank you for this PR, and I apologize for the late review on it. 
I have one issue with this PR and I am not sure whether or not it would be addressable with the current infra-structure:
- the current work flow is that a dev would go under Build\windows, double click on the batch file which generates the sln files

Now with this one, they would have to open the command line prompt, and add which version to build. is there a way to have the default version being the highest version detected on the machine ?

1480c1:
Sure, I can try to make it so that the default is blank ( cmake chooses the generator)

1480c1:
Would it be better to only build if 'build' is specified on the commandline? As in if you double click, it will generate the sln file only

1480c1:
@Cidana-Developers or @RyanZhao-Cidana, are the unit tests supposed to be able to work on windows?

1480c1:
Added testing for now. The default (double click) action is to generate the files for whatever is the default generator for cmake (`cmake -G` and look for the asterisk).

hassount:
> Would it be better to only build if 'build' is specified on the commandline? As in if you double click, it will generate the sln file only

well, if we're calling the batch build.bat, then it's best that it builds by default I guess

1480c1:
> > Would it be better to only build if 'build' is specified on the commandline? As in if you double click, it will generate the sln file only
> 
> well, if we're calling the batch build.bat, then it's best that it builds by default I guess

Reverted not autobuilding, changed it to a nobuild switch instead

PR #234, opened 2019-04-25
y4m: fix parser error for y4m input

xuguangxin:
EB_STRNCPY assume the dest is a static array, so it will use sizeof(dst) to get the size
But it's not true in copyUntilCharacterOrNewLine. So the y4m as input is totally breaking.

Since we can't check the type of dst, assume it's static array is harmful.
Better let the user pass the dst size.


xuguangxin:
@mafonsoN , any feedback for this?
thanks

mafonsoN:
Looks good. Thanks

xuguangxin:
@mafonsoN rebased.
 when we can merge this? else I need continue to rebase the patch.
thanks

xuguangxin:
Thanks mafonso.

PR #247, opened 2019-05-13
Fix CMakeLists issue in ASM_AVX2

Cidana-Developers:
CMAKE_C_FLAGS is overwrote in CMakeLists.txt in ASM_AVX2.


kylophone:
LGTM.

PR #253, opened 2019-05-18
Remove some unused code and '#if 0' / '#if 1'

joelsole:


kylophone:
LGTM, it's better to remove this. If anyone needs to recover, it's always in the Git history.

Source/Lib/Common/Codec/EbPictureControlSet.c
```diff
@@ -546,19 +540,6 @@ EbErrorType picture_control_set_ctor(
 
     }
 #if !MEMORY_FOOTPRINT_OPT
-    return_error = neighbor_array_unit_ctor(
```
hassount: I don't believe this should be removed until we cleanup MEMORY_FOOTPRINT_OPT otherwise the code would not build when it's off ?

Source/Lib/Common/Codec/EbProductCodingLoop.c
```diff
@@ -2880,27 +2880,6 @@ void AV1PerformFullLoop(
         candidateBuffer->candidate_ptr->full_distortion = (uint32_t)(y_full_distortion[0]);
 
         candidateBuffer->y_coeff_bits = y_coeff_bits;
-
-#if 0 //AMIR_DEBUG
```
hassount: @anaghdin 
if this is needed, maybe we can create a named MACRO for it?

Source/Lib/Common/Codec/EbPictureControlSet.h
```diff
@@ -13847,7 +13847,6 @@ extern "C" {
         // Mode Decision Refinement Neighbor Arrays
         NeighborArrayUnit                  *md_refinement_intra_luma_mode_neighbor_array;
         NeighborArrayUnit                  *md_refinement_mode_type_neighbor_array;
-        NeighborArrayUnit                  *md_refinement_luma_recon_neighbor_array;
```
hassount: ? is this a part of an #if 0 / #if 1

Source/Lib/Common/Codec/EbCodingUnit.h
```diff
@@ -217,15 +217,14 @@ extern "C" {
 
     typedef struct MacroBlockPlane 
     {
-#if 0
```
hassount: @chkngit @NaderMahdi 
I just want to make sure you guys are ok removing this

Source/Lib/Common/Codec/EbPictureControlSet.h
joelsole: This is unused code. This array is defined by not used anywhere.

Source/Lib/Common/Codec/EbPictureControlSet.c
joelsole: Fixed.

Source/Lib/Common/Codec/EbProductCodingLoop.c
anaghdin: No, It can be removed

PR #254, opened 2019-05-18
script: redo build.sh

1480c1:
Allows for custom c,cpp, and asm compiler
Allows for custom generator

PR #256, opened 2019-05-19
build: rename "linux" directory to "posix"

tmatth:
Rationale: there are already some Darwin specifics here.


tmatth:
If accepted, this should probably not be merged until #67 and/or #254 are merged.

kylophone:
Not arguing against it, but another thing to consider is this will probably break everyone's automatic builds.

hassount:
I agree with @kylophone this is going to impact all of our automated testing scripts. Is there a reason why we can't keep it as linux

tmatth:
> I agree with @kylophone this is going to impact all of our automated testing scripts. Is there a reason why we can't keep it as linux

I added a symlink to maintain compatibility (although it would not be my first choice).

PR #264, opened 2019-05-21
Switch CMake release builds to -O3

EwoutH:
[Improves performance](https://docs.google.com/spreadsheets/d/1p3PJQMkyhIrXEL6MBBQiwTnWmGyHY_zuw7_XAD0def4/edit#gid=1509095926) by about 2,5%.


hassount:
@EwoutH 
It seems that Windows does not recognize O3:
cl : Command line warning D9002: ignoring unknown option '-O3' [C:\projects\svt-av1\Source\Lib\Common\ASM_SSE4_1\COMMON_ASM_SSE4_1.vcxproj]

1480c1:
https://docs.microsoft.com/en-us/cpp/build/reference/o-options-optimize-code?view=vs-2019
O3 does not seem to exist

kylophone:
Yeah, we'll definitely want this. What's and equivalent on Windows? Googling tells me `/Ox`.

EwoutH:
Visual Studio wants -O2 or -Ox indeed, but is that the only compiler we need to consider on Window?

Also, do we need to add something for macOS?

I think a review of the CMakeList would be useful in genral.

1480c1:
Ox is a subset of O2 According to their docs. I would also like to support mingw soon ( my primary compiler on windows). For macOS, It should be the same.

I would like to redo the cmakefiles lists to go away from matching system name to matching compiler names etc to prevent stuff such as blocking mingw because it thinks /MP is a directory

1480c1:
O2 seems to be their designated max speed

EwoutH:
So for now we could only change Linux to -o3, but the CMakeList should be thoroughly reviewed.

kylophone:
> Visual Studio wants -O2 or -Ox indeed, but is that the only compiler we need to consider on Window?
@hassount ?

> 
> Also, do we need to add something for macOS?
Default CC on MacOS is clang, `-O3` should be fine.

> 
> I think a review of the CMakeList would be useful in genral.
Yeah there's a lot of hardcoding that could be a little more general, @1480c1 do you volunteer?

1480c1:
> Yeah there's a lot of hardcoding that could be a little more general, @1480c1 do you volunteer?

Sure, I have some of it done in my mingw-w64 pr, I can just separate them out and continue

PR #269, opened 2019-05-23
CI: Update AppVeyor to Visual Studio 2019

EwoutH:
Updates AppVeyor to Visual Studio 2019 (version 16.1).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,12 +60,12 @@ The SVT-AV1 Encoder adapts to the system that is being ran on. The memory requir
 ## Windows* Operating Systems (64-bit):
 
 * __Build Requirements__
-    -    Visual Studio* 2017 (can be downloaded [here](https://www.visualstudio.com/vs/older-downloads/))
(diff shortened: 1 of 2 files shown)


kylophone:
@hassount, any issue with this?

hassount:
> @hassount, any issue with this?

@kylophone the concern is that we would have a mismatch with the version that we are using locally (vs 17) which might make it a little tough to duplicate issues. We have to wait for IT to push VS19 for us to be able to use it. 
I guess we can go ahead as long as there's someone who can duplicate and fix specific VS19 issues,

1480c1:
@EwoutH , could you see if the vs2019 image has vs2017 also? if so we could make two jobs with one running vs2019 and the other vs2017

1480c1:
It seems that their image does not include 2017.

hassount:
ok, let's merge this and see how it goes, if we face major issues, we can revert back

PR #281, opened 2019-05-26
Coverity scan: add coverity scan

1480c1:
I noticed that a lot of the 01org projects are on coverity (https://scan.coverity.com/projects/01org-libva) and thought that it might be okay to add svt-av1 to it too.

Looks like this: https://scan.coverity.com/projects/1480c1-svt-av1

The reports look like: 
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -142,12 +142,6 @@ matrix:
             - yasm
       before_install:
         - "sudo chown -R travis: $HOME/.ccache"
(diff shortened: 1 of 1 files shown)


1480c1:
To actually use it, someone will need to activate it at the website and create a branch named `coverity_scan` and merge to it after major commits.

1480c1:
Going to temp close

PR #282, opened 2019-05-27
Verify bitstream writer using svt_reader

Cidana-Developers:
1. Remove aom_reader since svt_reader is ready. The build of
unit test is broken since c7b108ad518b2f223fa7d56c3fdbd482086bc261,
and this CL fixes it.


test/CMakeLists.txt
```diff
@@ -59,6 +60,8 @@ file(GLOB all_files
     "ref/*.h"
     "ref/*.cc"
     "../Source/Lib/Encoder/Codec/*.c"
+    "../Source/Lib/Decoder/Codec/EbDecBitReader.c"
```
hassount: seems like a duplication as all c files are added already

Cidana-Developers: Actually EbDecBitReader.c is in Decoder.

hassount: you're right, my bad

kylophone:
lgtm

PR #284, opened 2019-05-27
EbDefinitions: Don't redefine fseeko64 and ftello64

1480c1:
musl already defines fseeko64 as fseek and ftello64 as ftello

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -25,8 +25,12 @@
 #define inline __inline
 #elif __GNUC__
 #define inline __inline__
+#ifndef fseeko64
 #define fseeko64 fseek
+#endif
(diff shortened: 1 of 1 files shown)


kylophone:
> musl already defines fseeko64 as fseek and ftello64 as ftello

What about other libc implementations?

1480c1:
Unsure about other libc as I have not yet dabbled with them. For trying to only do this with musl, as far as I can find, musl purposely made it so you can't differentiate between musl and other libc implementations

hassount:
@kylophone 
do you see this impacting much our supported platforms ? we have tested this on Linux / Windows and it's working well

1480c1:
Would it be better to undef and then define?

hassount:
@kylophone 
Is this resolved? I plan to merge it tonight assuming no objections from your side

PR #288, opened 2019-05-28
 Decoder: Added support for Tiles and 10bit

nijilk:
Decoder: Added support for Tiles and 10bit
Minor bug fixes related to incomplete SB at frame boundary


1480c1:
Could you run `git rebase OVC/master`?

1480c1:
Either that or make a new branch for a new feature branching off from master and do a pr from there to keep the commit history clean

kylophone:
Please also add a changelog entry

PR #248, opened 2019-05-14
Style clean-up: trailing spaces, multiple new lines, single-statement…

joelsole:
… braces

PR #292, opened 2019-05-28
CI: Add coding style check to Travis, cleanup

EwoutH:
- Add coding style check (Borrowed from dav1d's [.gitlab-ci.yml](https://code.videolan.org/videolan/dav1d/blob/master/.gitlab-ci.yml))
- Sets Ubuntu 16.04 Xenial and GCC as default settings, only uses "os:" and "compiler:" if different
- Fixes coding style error (trailing white space) in "name: unittest"

For now the Style check is allowed to fail (waiting on #248), but that should be reversed as soon as possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,6 +89,9 @@ matrix:
     # Coding style check
     - name: Style check
       stage: style
+      addons: skip
+      before_install: skip
+      install: skip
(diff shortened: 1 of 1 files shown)


EwoutH:
Replaced by #293.

PR #293, opened 2019-05-28
[WIP] CI: Add coding style check to Travis, cleanup

EwoutH:
- [x] Add coding style check (Borrowed from dav1d's [.gitlab-ci.yml](https://code.videolan.org/videolan/dav1d/blob/master/.gitlab-ci.yml))
 - [x] Fixes coding style error (trailing white space) in "name: unittest"

For now the Style check is allowed to fail (waiting on #248), but that should be reversed as soon as possible.

**Todo:**
 - [ ] Select files and directories to exclude
 - [ ] Remove those from dav1d

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,7 +103,7 @@ matrix:
               fi;
           done
         - git remote rm upstream 2> /dev/null || true
-        - git remote add upstream https://github.com/OpenVisualCloud/SVT-AV1
+        - git remote add upstream https://github.com/OpenVisualCloud/SVT-AV1.git
         - git fetch -q upstream master
         - for i in $(git rev-list HEAD ^upstream/master); do
               echo "Checking commit message of $i";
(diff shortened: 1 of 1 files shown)


kylophone:
> * Add coding style check (Borrowed from dav1d's [.gitlab-ci.yml](https://code.videolan.org/videolan/dav1d/blob/master/.gitlab-ci.yml))
> * Fixes coding style error (trailing white space) in "name: unittest"
> 
> For now the Style check is allowed to fail (waiting on #248), but that should be reversed as soon as possible.

Yeah, I love this. The current state of the code does not match the style guide in the least bit. For the time being, I think we should at least checking the PR diffs. That would cut down on the noise and manual effort involved when it comes to patch review. Will let others chime in as well.

EwoutH:
We should probably exclude the third_party libraries (googletest and aom) from the style check.

kylophone:
> We should probably exclude the third_party libraries (googletest and aom) from the style check.

Yes, that would be preferable.

EwoutH:
Any other files we want to exclude except from `third_party`?

PR #297, opened 2019-05-28
Docs: Make the decoder guide readable

EwoutH:
- Removed double spaces
- Added code blocks
- Fixed interpunction
- Removed main profile reference (AV1 main profile requires 10-bit and 4:0:0 support)

### DIFF
--- a/Docs/svt-av1_decoder_user_guide.md
+++ b/Docs/svt-av1_decoder_user_guide.md
@@ -3,25 +3,25 @@
 ## Table of Contents
 1. [Introduction](#introduction)
 2. [Known Limitations](#known-limitations)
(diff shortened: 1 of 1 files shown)


kylophone:
Thanks. It would also be a good idea to eventually add a section about the purpose of the decoder, since people have been asking about it. Not meant to compete with any production decoder (use dav1d!), it's meant to aide in codec research, and provide a framework with shared encoder/decoder structures. The hope is eventually this will be a place to do research of new tools, etc.

I say we merge this as-is for now, and maybe I'll add a small section to the decoder doc outlining the above?

cc @andrey-norkin for visibility.

PR #311, opened 2019-05-30
Disable overlay pictures (used by the Alt-refs) by default

mafonsoN:
## Description

A bug was found after merging the Alt-ref pictures support and disabling the encoding of the overlay pictures seems to be fixing it. So, this should be disabled by default until the problem is fixed. 

## TODO items

The issue will be debugged and a bug fix PR will be submitted.


hassount:
@mafonsoN 
Could you please create an issue with a sample command line so that we could track it ?

PR #317, opened 2019-05-31
Move .clang-format file from ./test/ to ./

joelsole:


1480c1:
Be careful as the clang-format seems somewhat unfinished.
Plus it has the TabWidth set as 8.

PR #320, opened 2019-06-02
cleanup: add a bash script

1480c1:
add a cleanup script to auto remove trailing whitespace and remove multiple newlines in a row.

Currently WIP

### DIFF
--- a/.cleanup.sh
+++ b/.cleanup.sh
@@ -1,0 +1,204 @@
+#!/bin/bash
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX - License - Identifier: BSD - 2 - Clause - Patent
+#
+set -x
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$script_dir"
+
+call_help(){
+    cat <<EOF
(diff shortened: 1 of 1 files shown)


kylophone:
Would it be better to use a `.clangformat` file to do this?

1480c1:
I was thinking so too but there are a couple of problems with the clang-format, 
1. It requires clang-format
2. Only works on c and h files
3. Inconsistent ATM

1480c1:
So if you wanted to clean up white space in cmakelists or MD files, clang-format would not be able to do it

ttrigui:
@1480c1 is this still a WIP?

1480c1:
Yes for now since I can't come up with a better way. I would prefer to set it up with clang format, however the style guide lines are very vague

EwoutH:
@1480c1 Are you still working on this?

1480c1:
Currently waiting to see if a clangformat can be added, I could add it as is, but it won't really be a cleanup script then

EwoutH:
So blocked by #574?

1480c1:
Somewhat yes

hassount:
@1480c1 should we close this and make another one after the cleanup ?

1480c1:
Sure

PR #331, opened 2019-06-04
Add intra pred unit test

Cidana-Developers:
1. add unit test for all the intra prediction modes except palatte
   and intrabc.
2. add unit test for upsampling and edge filter process.
3. cfl_predict_lbd_avx2 and cfl_predict_hbd_avx2 will produce
    mismatch result compared with c version.

### DIFF
--- a/test/intrapred_cfl_test.cc
+++ b/test/intrapred_cfl_test.cc
@@ -50,7 +50,7 @@ using CFL_PRED_LBD = void (*)(const int16_t *pred_buf_q3, uint8_t *pred,
  * alpha_q3: [-16, 16]
  * BitDepth: 8bit and 10bit
  */
-template <typename Pixel, typename FuncType>
+template <typename Sample, typename FuncType>
 class CflPredTest {
   public:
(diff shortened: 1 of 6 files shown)


test/intrapred_cfl_test.cc
```diff
@@ -0,0 +1,169 @@
...
+ * Output from assemble functions should be the same with output from c.
+ *
+ * Test coverage:
+ * Test cases:
+ * pred buffer and dst buffer: Fill with random values
+ * TxSize: all the TxSize.
+ * alpha_q3: [-16, 16]
+ * BitDepth: 8bit and 10bit
+ */
+template <typename Pixel, typename FuncType>
```
andrey-norkin: I suggest we use Sample if we refer to a value of a single component, such as a luma sample or a chroma sample. Pixel usually denotes a triplet of values corresponding to a certain spatial position, such as <R,G,B>.

Cidana-Developers: OK, done.

PR #335, opened 2019-06-05
Feature/segmentationmerge4

anush798:
Initial set of commits to enable segmentation. Uses a very basic variance-based adaptive quantization to test the path. While not thoroughly tested, the feature is relatively stable. PR created so that future changes can easily be added and we do not keep running into merge conflicts. 
Rec-dec match tested for RC off. There may still be errors though, so not expected for wide use yet.


anush798:
Incorrect branch.

PR #336, opened 2019-06-05
[Bug fix] filtering strength adjustment for Alt-ref tool

mafonsoN:
### Description

Fixes a bug in the adjustment of the temporal filtering strength and introduces a better handling of the corner cases where not enough pictures are available for the temporal filtering. Finally, it also contains some small clean-ups of unused code (lossless) and small fixes. 

### Type of change

Bug fix and clean-up

## Tests and performance

PSNR-YUV BD-rate gains of approximately **-0.3%** on the AOM test set.
[4 QP values: {20, 32, 43 and 55}]

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -827,19 +827,6 @@ static EbErrorType VerifySettings(EbConfig *config, uint32_t channelNumber)
         return_error = EB_ErrorBadParameter;
     }
 
-    // --- start: ALTREF_FILTERING_SUPPORT
-    // alt-ref frames related
-    if (config->altref_strength > 6 ) {
-        fprintf(config->error_log_file, "Error instance %u: invalid altref-strength, should be in the range [0 - 6] \n",channelNumber+1);
-        return_error = EB_ErrorBadParameter;
-    }
-
-    if (config->altref_nframes > 15 ) {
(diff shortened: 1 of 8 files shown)


Source/App/EncApp/EbAppConfig.c
```diff
@@ -834,8 +834,8 @@ static EbErrorType VerifySettings(EbConfig *config, uint32_t channelNumber)
         return_error = EB_ErrorBadParameter;
     }
 
-    if (config->altref_nframes > 7 ) {
-        fprintf(config->error_log_file, "Error instance %u: invalid altref-nframes, should be in the range [0 - 7] \n",channelNumber+1);
+    if (config->altref_nframes > 15 ) {
```
hassount: is 15 due to the fact that the mini-gop length is 16 ? 
- if yes, can we make this dependent on the minigop size? 
- if not, can we use a MACRO for it to define what 15 stands for ? ALTREF_MAX_NFRAMES ? 
last but not least, could we move these checks to EbEncHandle.c as they will not be there if the library is built into ffmpeg.

mafonsoN: No, 15 is the value supported in libaom. Actually, it's better to stick to a maximum of 10 for now in order to keep the same delay in the picture decision process. 

I've added two commits to address this and also move the check to the library.

PR #342, opened 2019-06-06
Decoder Bug fix in Mac/Linux platform

nijilk:
ENUM argument exceeding the range was causing the problem in Mac/Linux


kylophone:
LGTM. 

Can you explain this? Is the enum getting optimized into something smaller than an `int`? I wonder if this is UB?

nijilk:
> LGTM.
> 
> Can you explain this? Is the enum getting optimized into something smaller than an `int`? I wonder if this is UB?

Yes.
This ENUM was getting optimized to a char.

PR #252, opened 2019-05-18
IoK/FB/051719 - Offer new downsample_2d function to perform 2x2 bilinear

ikatsavounidisFB:
filtering instead of decimation

- Controlled by #define DECIMATION (line 4954, EbPictureAnalysisProcess.c)
- Minor bug fixing in original decimation_2d
- Fix compiler warning in cfl_avx2.c
- Add XCode project generator script

PR #353, opened 2019-06-10
Downsampling filtering

hguermaz:
## Description
* Added down-sampling filtering support.
* Used the down-sampled filtered 1/16th and 1/4th input/reference frame(s) @ both ME and alt-ref temporal filtering .
* Added a multi-mode signal to control the down-sampling method; use the filtered if M0, and use the decimated if M1 or  higher.
* Removed HME Level 0 check @ 1/16 th decimation to guarantee valid ZZ SAD and SCD data when HME Level 0 is OFF (lossless bug fix).

## Author

@ikatsavounidisFB 
@hguermaz 

## Type of change

New feature/tool and bug fix.

## Tests and performance
* Bd-Rate gain(s):
Full sub-4K aom list: ~0.4% to aom_2pass_cpu0 on the.
Per resolution: ~0.7%, ~0.4% and ~0.2% to aom_2pass_cpu0 on 360p, 720p and 1080p respectively.

* Speed & memory footprint:
~0% speed deviation and ~0.3% memory increase in M0.
No memory or speed overhead in M1 & higher.

### DIFF
--- a/Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
+++ b/Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
@@ -163,11 +163,11 @@ static void DownSampleChroma(EbPictureBufferDesc* input_picture_ptr, EbPictureBu
 void decimation_2d(
     uint8_t *  input_samples,      // input parameter, input samples Ptr
     uint32_t   input_stride,       // input parameter, input stride
-    uint32_t   input_area_width,    // input parameter, input area width
-    uint32_t   input_area_height,   // input parameter, input area height
+    uint32_t   input_area_width,   // input parameter, input area width
+    uint32_t   input_area_height,  // input parameter, input area height
     uint8_t *  decim_samples,      // output parameter, decimated samples Ptr
     uint32_t   decim_stride,       // input parameter, output stride
-    uint32_t   decim_step)        // input parameter, decimation amount in pixels
+    uint32_t   decim_step)         // input parameter, decimation amount in pixels
 {
     uint32_t horizontal_index;
     uint32_t vertical_index;
@@ -4985,11 +4985,13 @@ void* picture_analysis_kernel(void *input_ptr)
                 (EbPictureBufferDesc*)paReferenceObject->sixteenth_decimated_picture_ptr);
 
             // 1/4 & 1/16 input picture downsampling through filtering
-            DownsampleFilteringInputPicture(
-                picture_control_set_ptr,
-                input_padded_picture_ptr,
-                (EbPictureBufferDesc*)paReferenceObject->quarter_filtered_picture_ptr,
-                (EbPictureBufferDesc*)paReferenceObject->sixteenth_filtered_picture_ptr);
+            if (sequence_control_set_ptr->down_sampling_method_me_search == 0) {
+                DownsampleFilteringInputPicture(
+                    picture_control_set_ptr,
+                    input_padded_picture_ptr,
+                    (EbPictureBufferDesc*)paReferenceObject->quarter_filtered_picture_ptr,
+                    (EbPictureBufferDesc*)paReferenceObject->sixteenth_filtered_picture_ptr);
+            }
 #else
             // 1/4 & 1/16 input picture decimation
             DecimateInputPicture(
--- a/Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
+++ b/Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
@@ -174,11 +174,9 @@ void decimation_2d(
     uint32_t input_stripe_stride = input_stride * decim_step;
 
     for (vertical_index = 0; vertical_index < input_area_height; vertical_index += decim_step) {
-        for (horizontal_index = 0; horizontal_index < input_area_width; horizontal_index += decim_step) {
-
+        for (horizontal_index = 0; horizontal_index < input_area_width; horizontal_index += decim_step) 
             decim_samples[(horizontal_index >> (decim_step >> 1))] = input_samples[horizontal_index];
 
-        }
         input_samples += input_stripe_stride;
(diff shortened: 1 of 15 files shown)


hassount:
@OmranAbdelkafi 
Can you please evaluate this PR on M1-M8.

Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
```diff
@@ -4825,23 +5009,44 @@ void* picture_analysis_kernel(void *input_ptr)
             // Pad input picture to complete border LCUs
             PadPictureToMultipleOfLcuDimensions(
                 input_padded_picture_ptr);
-
+#if DOWN_SAMPLING_FILTERING
+            // 1/4 & 1/16 input picture decimation
+            DownsampleDecimationInputPicture(
+                picture_control_set_ptr,
+                input_padded_picture_ptr,
```
mafonsoN: If I understand correctly, in this implementation, downsampling through decimation is always carried out and filtering is done only if the down_sampling_method_me_search is 0? Is the decimation necessary in this case?

Source/Lib/Encoder/Codec/EbEncHandle.c
```diff
@@ -2231,6 +2263,15 @@ void SetParamBasedOnInput(SequenceControlSet *sequence_control_set_ptr)
 #else
     sequence_control_set_ptr->nsq_present = 1;
 #endif
+#if DOWN_SAMPLING_FILTERING
+    // Set down-sampling method     Settings
+    // 0                            0: filtering
+    // 1                            1: decimation
+    if (sequence_control_set_ptr->static_config.enc_mode == ENC_M0)
+        sequence_control_set_ptr->down_sampling_method_me_search = 0;
```
mafonsoN: Maybe consider using an enum here instead of 0 and 1 for better code readability

Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
hguermaz: Filtered 1/16th and 1/4th input/reference frame(s) are only used @ ME and alt-ref temporal filtering . ZZ SAD, SCD, .. are still using the decimated 1/16th and 1/4th input/reference frame(s For example we performed a quick analysis @ picture activity derivation, and found that the average deviation between filtered-to-filtered SADs and decimated-to-decimated SADs on niklas360p is pretty high ~-40%.  Therefore, to preserve similar or better behavior, the thresholds used to derive the picture activity class from (0,0) SAD should be decreased (i.e. apply a weight based on offline data to the existing (block size, resolution) normalized thresholds) or keep using decimated @ ZZ SAD.

Source/Lib/Encoder/Codec/EbEncHandle.c
hguermaz: I agree.

Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
```diff
@@ -4733,6 +4838,75 @@ static int is_screen_content(const uint8_t *src, int use_hbd,
...
+                downsample_2d(
+                    &quarter_picture_ptr->buffer_y[quarter_picture_ptr->origin_x + quarter_picture_ptr->origin_y * quarter_picture_ptr->stride_y],
+                    quarter_picture_ptr->stride_y,
+                    quarter_picture_ptr->width,
+                    quarter_picture_ptr->height,
+                    &sixteenth_picture_ptr->buffer_y[sixteenth_picture_ptr->origin_x + sixteenth_picture_ptr->origin_x*sixteenth_picture_ptr->stride_y],
+                    sixteenth_picture_ptr->stride_y,
+                    2);
+            }
+            else {
```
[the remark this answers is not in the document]
hguermaz: Done.

Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
```diff
@@ -155,7 +155,73 @@ static void DownSampleChroma(EbPictureBufferDesc* input_picture_ptr, EbPictureBu
...
+    uint8_t *  decim_samples,      // output parameter, decimated samples Ptr
+    uint32_t   decim_stride,       // input parameter, output stride
+    uint32_t   decim_step)         // input parameter, decimation amount in pixels
+{
+    uint32_t horizontal_index;
+    uint32_t vertical_index;
+    uint32_t input_stripe_stride = input_stride * decim_step;
+
+    for (vertical_index = 0; vertical_index < input_area_height; vertical_index += decim_step) {
+        for (horizontal_index = 0; horizontal_index < input_area_width; horizontal_index += decim_step) {
```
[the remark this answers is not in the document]
hguermaz: Done.

Source/Lib/Common/Codec/EbPictureAnalysisProcess.c
hguermaz: Speed cost is not significant in M0 where ME represents less than 5% of the whole encoder. However for high density modes, the impact in term of speed will become noticeable. 

Please note that by "No memory or speed overhead in M1 & higher" we mean that the added infrastructure/support does not impact memory or speed in M1 & higher (when the feature is OFF), but we did not test yet the case when the the feature is actually ON.

mafonsoN: Ok, got it. Later perhaps we could consider updating these thresholds by using data gathered from using the filtering.

Source/Lib/Encoder/Codec/EbEncHandle.c
hguermaz: Done. Added DownSamplingMethod enum.

hassount:
This PR is a replacement to #252

hassount:
@mafonsoN is this good to go ?

Issue #346, opened 2019-06-07
Segfaults after merging PR #333 


mafonsoN:
Encoding modes > 0 are crashing on Mac in the SIMD functions. Reverted back to commit 2f72785, just before #333 was merged and it seems to be working fine, so guessing this introduced the problem?

Example command line:

./SvtAv1EncApp -i speed_bag_640x360_60f.yuv -w 640 -h 360 -fps-num 50000 -fps-denom 1000 -q 63 -n 60 -b speed_bag_640x360_60f_svtav1_cq63.ivf -o speed_bag_640x360_60f_svtav1_cq63.yuv -intra-period 119 -bit-depth 8 -enc-mode 5

Error: 

EXC_BAD_ACCESS (code=EXC_I386_GPFLT)

Call stack: 
yy_loadu_256 synonyms_avx2.h:31
cross_sum_fast_even_row selfguided_avx2.c:438
final_filter_fast selfguided_avx2.c:511

hassount:
@mafonsoN thank you for reporting this.
it's weird how the CI actually passed while testing -enc-mode 8 :/

1480c1:
Probably because the macOS vms did not have a avx2 CPU so it did not use the avx2 functions

mafonsoN:
> @mafonsoN thank you for reporting this.
> it's weird how the CI actually passed while testing -enc-mode 8 :/

Actually, enc-mode 8 is the only one working on my tests :).

hassount:
gotcha, then the wiener filter opt is a good suspect. then @1480c1 question is accurate, do you have avx2 supported on your machine?

mafonsoN:
> Probably because the macOS vms did not have a avx2 CPU so it did not use the avx2 functions
> […](#)
> On Jun 7, 2019, at 2:55 PM, Hassene Tmar ***@***.***> wrote: @mafonsoN thank you for reporting this. it's weird how the CI actually passed while testing -enc-mode 8 :/ — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or mute the thread.



> gotcha, then the wiener filter opt is a good suspect. then @1480c1 question is accurate, do you have avx2 supported on your machine?


Yes, I have. I'm running it on my Macbook Pro.

lzhangnj:
[...]

PR #333, opened 2019-06-04
[SIMD] Wiener Filter AVX2

lzhangnj:
## Description
Redo weiner filter avx2 optimizations, and add unit test code.

## Expected gain per kernel
Speedup against original c and avx2 functions (low bitdepth):
TAP 3:  66x 17x
TAP 5: 125x 21x
TAP 7: 188x 20x

Speedup against original c and avx2 functions (high bitdepth):
8-bit
TAP 3:  57x N/A
TAP 5: 113x 59x
TAP 7: 167x 65x

10-bit or 12-bit
TAP 3:  41x N/A
TAP 5:  80x 42x
TAP 7: 128x 50x

Associated with Issue #262

PR #354, opened 2019-06-10
Weiner filter avx2 bug fix

lzhangnj:
Fix segfaults in Wiener Filter AVX2

Issue #346 on MAC is caused by large stack usage. Moved the buffer to heap.

Related to PR #333


hassount:
@lzhangnj 
Does this change the bitstream for windows / linux?

lzhangnj:
@hassount  No. Should be bit-exact for all platforms.

mafonsoN:
I did some quick tests and the bug seems to be solved for Mac. Are there any speed implications of these changes?

hassount:
@mafonsoN 
thank you for confirming, I don't see a speed difference on my end. @lzhangnj do you expect any speed loss with the change ?

lzhangnj:
@hassount It's called in the size of 256x256 blocks, so the impact on speed would be very trivial.

PR #357, opened 2019-06-11
Travis: inline anchors and add chown before ffmpeg configure to fix ccache ownership problems

1480c1:
Inline base_script, coveralls_script, and after_coveralls_script, add chown before configuring ffmpeg to try to prevent ffmpeg issues of ccache file ownership


1480c1:
Also fixes problem related macOS Homebrew not installing yasm

1480c1:
Most of the changes other than inline is already in master

PR #319, opened 2019-06-02
Travis CI: Remove allow_failure from Style job

EwoutH:
Merge when #318 is fixed.

PR #360, opened 2019-06-12
cleanup: remove trailing whitespace and convert tab to four spaces

1480c1:
cleanup according to the cleanup job in travis

convert exit 1 to travis_terminate since it is bad practice to call exit in travis directly
Add the cfg and txt file from test since it looks like it would be bad to add newlines to the EOF for those
Completely excluded third_party since that was also excluded ( plus there were a lot of vcproj etc files that I was not sure if they should be touched )

I looked through to make sure that the indention stayed the same, but if there is any I missed, please tell me.

#319 should be able to be unblocked with this


EwoutH:
Style check passes, nice!

tianjunwork:
Hi @1480c1 could you share the tool and how did you do that? Thank you.
I wonder we should do the same for HEVC and VP9.

1480c1:
I used sed and find (https://github.com/OpenVisualCloud/SVT-AV1/blob/master/STYLE.md#post-coding) I will add the commands to change tabs to spaces in the readme pr

1480c1:
I can do the same for hevc and vp9 too later

tianjunwork:
@1480c1 I think introduce STYLE.md to HEVC and VP9 (needs update) is also necessary. Could you include STYLE.md in your PR to HEVC and VP9? Thanks:)

1480c1:
Sure will do

hassount:
@mafonsoN @kylophone @joelsole 
if there are no objections with this, I ll be merging this PR to close the coding style item.

EwoutH:
@hassount Thanks for the P1 label. Have a sec on IRC?

PR #364, opened 2019-06-12
Add SOVERSION for Encoder and Decoder shared libraries

### LINKED ISSUE #363: Could you please add a SOVERSION to libraries?
Currently libraries are unversioned which prevents inclusion in Fedora.

eclipseo:
Fixes #363

Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

### DIFF
--- a/Source/Lib/Decoder/CMakeLists.txt
+++ b/Source/Lib/Decoder/CMakeLists.txt
@@ -4,6 +4,12 @@
 #
 # Decoder Library directory CMakeLists.txt
 
+# Shared Decoder Version
+set(DEC_VERSION_MAJOR 0)
+set(DEC_VERSION_MINOR 5)
+set(DEC_VERSION_PATCH 0)
+set(DEC_VERSION ${DEC_VERSION_MAJOR}.${DEC_VERSION_MINOR}.${DEC_VERSION_PATCH})
+
 # Include Decoder Subdirectories
 include_directories(${PROJECT_SOURCE_DIR}/Source/API/
     ${PROJECT_SOURCE_DIR}/Source/Lib/Common/Codec/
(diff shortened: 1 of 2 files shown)


Source/Lib/Encoder/Codec/CMakeLists.txt
```diff
@@ -1,9 +1,15 @@
-# 
+#
 # Copyright(c) 2019 Intel Corporation
 # SPDX - License - Identifier: BSD - 2 - Clause - Patent
-# 
+#
 # EncoderLib Directory CMakeLists.txt
 
+# Shared Encoder Version
+set(ENC_VERSION_MAJOR 1)
```
jbeich: According to [semantic versioning](https://semver.org/) until API (and ABI) stabilizes SOVERSION can only be 0. For one, 653d26c0f32e moved `stat_report` in the struct -> old consumers may assign value to the wrong field.

hassount: agreed, code is not yet stable for a 1.x similarly with the decoder

eclipseo:
Rebased.
Used version 0.5.0 following: https://github.com/OpenVisualCloud/SVT-AV1/pull/396

1480c1:
@hassount, are the encoder and decoder going to be considered to have separate versions? As in once the encoder hits 1.0.0 status, will the decoder be considered 1.0.0? Or will the development be considered separate?

hassount:
the process will probably be different for both so I think we can use different versions for now and Synch after we hit a 1.0 that said, I would let @joelsole @andrey-norkin @kylophone  jump in here

kylophone:
LGTM.

PR #374, opened 2019-06-13
Incomplete SBs fix

anaghdin:
Description
Handle the incomplete SBs properly based on the standard and consider all allowed blocks

Authors
@anaghdin

Type of change
Feature improvement

Tests and performance
Expected Average PSNR-SSIM BD-rate gain in the range of -0.9% on the AOM test set, using 4 QP values: {20, 32, 43 and 55}.

No Speed impact.

TODOs:
If the cropped width is not 4, 8, 16, 32, 64 and 128, the block is not allowed. To be removed after intrinsic functions for NxM spatial_full_distortion_kernel_func_ptr_array are added
Only active for M0, to be evaluated for other modes


Source/Lib/Common/Codec/EbDefinitions.h
```diff
@@ -34,6 +34,8 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
```
hassount: extra space

PR #376, opened 2019-06-14
CI: Add binary identicality test to Travis

EwoutH:
Adds a binary identicality test to Travis. The test compares two encodes (enc-mode 0 and 8) encoded from the PR to encodes from the master branch to check if they are bit-identical. It can be used as a check for PRs where no change in bitstream is expected.

**Examples:**
- [A run that passed](https://travis-ci.com/EwoutH/SVT-AV1/jobs/207970232)
- [A run that failed](https://travis-ci.com/EwoutH/SVT-AV1/jobs/207970997)

The whole test is allowed to fail, because of course sometimes we actually want to change the bitstream.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,14 +148,13 @@ matrix:
     - name: Binary Identical?
       stage: test
       script:
-       - export PATH="$HOME:$PATH"
        - *base_script
        - mv -t $HOME akiyo_cif.y4m
        - SvtAv1EncApp -enc-mode 8 -i $HOME/akiyo_cif.y4m -n 120 -b test-pr-m8.ivf
        - SvtAv1EncApp -enc-mode 0 -i $HOME/akiyo_cif.y4m -n 3 -b test-pr-m0.ivf
        - mv -t $HOME test-pr-m8.ivf test-pr-m0.ivf
        - cd $HOME && rm -rf SVT-AV1
-       - git clone https://github.com/OpenVisualCloud/SVT-AV1.git && cd SVT-AV1
+       - git clone --depth 1 https://github.com/OpenVisualCloud/SVT-AV1.git && cd SVT-AV1
        - *before_install
        - *base_script
        - SvtAv1EncApp -enc-mode 8 -i $HOME/akiyo_cif.y4m -n 120 -b test-master-m8.ivf
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,7 @@ before_cache:
 matrix:
   fast_finish: true
   allow_failures:
+    - name: Binary Identical?
     - name: Valgrind
     - env: COVERALLS_PARALLEL=true build_type=debug CMAKE_EFLAGS="-DCOVERAGE=ON"
     - name: Unit Tests Linux+gcc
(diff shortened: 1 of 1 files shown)


1480c1:
Can you test if you can just checkout master instead of re cloning?

EwoutH:
I still have to remove and rebuild right?

1480c1:
Mainly because if you can checkout master, all you would have to remove would be Bin and Build/${build_type}. 

I was mainly asking if Travis kept master locally or only checkout the working branch/commit

1480c1:
I'll try in a little bit, working with travis on the phone is a bit annoying

EwoutH:
Would be great, if you have a more elegant way of moving files that would also be welcome.

EwoutH:
@hassount Since this is CI code only, doesn't affect the encoder and for-fills it purpose, I would advise merging immediately, so that rebased from now on can check on bitstream identicality. Any improvements to the CI config can be added later.

hassount:
@EwoutH thank you very much :) this is great. for -enc-mode 0 can we do -n 17 ? a minigop size at least so that we get some code coverage.

EwoutH:
@hassount Made a branch with that change, it takes about 106 seconds per run, and the total job about 7 minutes. https://travis-ci.com/EwoutH/SVT-AV1/jobs/208048086

joelsole:
Another useful test would be run-2-run matching. What about including a second run for the PR to check whether it is bit-identical results with the first run?

EwoutH:
@joelsole How big is the chance that two runs from different commits are identical, but two runs from different PRs are not?

I don't think Travis is the best tool for bitstream checks because of the low thread count (no parallel jobs possible) and limited control of environment (turbo clock speeds, noisy neighbours, etc.). I'm working with Nathan (from Mozilla) on a far more extensive codec test platform, which will evaluate coding efficiency (PSNR, SSIM, VMAF, etc.), speed and use of resources. Bitstream identicality will be part of it.

joelsole:
> @joelsole How big is the chance that two runs from different commits are identical, but two runs from the same PR are not?

It should be zero, but it is not. Agree that Travis is not the best tool for this - and a single check would not be enough. 
Great news that you are working on a codec test platform! When will be it be available?

EwoutH:
@joelsole Soon™. We're in an very early stage, so maybe we can collaborate on some parts.

.travis.yml
```diff
@@ -143,6 +144,25 @@ matrix:
...
+      stage: test
+      script:
+       - export PATH="$HOME:$PATH"
+       - *base_script
+       - mv -t $HOME akiyo_cif.y4m
+       - SvtAv1EncApp -enc-mode 8 -i $HOME/akiyo_cif.y4m -n 120 -b test-pr-m8.ivf
+       - SvtAv1EncApp -enc-mode 0 -i $HOME/akiyo_cif.y4m -n 3 -b test-pr-m0.ivf
+       - mv -t $HOME test-pr-m8.ivf test-pr-m0.ivf
+       - cd $HOME && rm -rf SVT-AV1
+       - git clone https://github.com/OpenVisualCloud/SVT-AV1.git && cd SVT-AV1
```
1480c1: Maybe add a depth and branch flag here and set to a low number (might shave off a few secs especially since we aren't doing anything else other than checking out master)

.travis.yml
```diff
@@ -143,6 +144,25 @@ matrix:
       - cmake .
       - make --quiet -j$(nproc)
       - sudo make install
+    # Tests if .ivf files are identical on binary level
+    - name: Binary Identical?
+      stage: test
+      script:
+       - export PATH="$HOME:$PATH"
```
1480c1: What's the point of adding `$HOME` to `$PATH`? I don't necessarily see anything that would require that

EwoutH: I used it before, but in the current config it should work without it indeed. Removed!

.travis.yml
EwoutH: Good one, done!

EwoutH:
@1480c1 @ttrigui Any more comments?

1480c1:
None as far as I can see.

I did confirm that Travis does not checkout any branches, but checkouts the commit directly, meaning we would have to add upstream, pull, and checkout master manually

EwoutH:
@1480c1 Thanks! We should discuss that for a more extensive test suite.

Do you happen to know any kind of CI that integrates with GitHub and can report metrics instead of a pass/fail? I'm now looking into a [Probot](https://probot.github.io/) implementation to report metrics back in a comment, but it's not ideal.

1480c1:
Not as far as I know, but I will also research it too

EwoutH:
@1480c1 We should talk sometime about this stuff. Do you have a preferred platform?

1480c1:
Not really, although I might have problems with irc due to shoddy connection from my phone

1480c1:
Maybe discord since it's async and I already have it on my phone ( I remember you saying something about it in an earlier issue)

1480c1:
My discord is 1480c1#3982

ttrigui:
@EwoutH I did not get the chance to test it yet but this looks good to me.

EwoutH:
Thanks, @hassount lets merge

EwoutH:
It really demotivates to have PRs open so long without a good reason.

hassount:
Sorry, forgot to merge this.

Issue #262, opened 2019-05-21
SIMD assembly


EwoutH:
Based on testing in https://github.com/OpenVisualCloud/SVT-AV1/issues/98#issuecomment-494560666, this functions without AVX2 are taking up more than 2% runtime on enc-mode 8, 6, 4, 2 or 0 (in no particular order):

**AVX2**
- [ ] av1_inter_prediction
- [x] av1_warp_affine_c
- [ ] av1_quantize_inv_quantize
- [ ] open_loop_me_get_search_point_results_block
- [x] ExtSadCalculation (#108)
- [ ] av1_cost_coeffs_txb
- [ ] avc_style_luma_interpolation_filter_vertical_ssse3_intrin
- [x] av1_quantize_fp_facade (done)
- [x] av1_fwd_txfm2d_32x16_c 
- [x] av1_inv_txfm_add_ssse3
- [x] av1_lowbd_inv_txfm2d_add_ssse3
- [x] transpose16x16_sse2
- [x] spatial_full_distortion_kernel_func_ptr_array ( to develop avx2 kernel that supports MxN sizes where N and M are multiple of 4)
- [ ] convolve_2d_for_intrabc

**AVX512**
- [ ] See [lm-opt](https://github.com/hguermaz/SVT-AV1/tree/lm-opt) branch

For these function AVX2 assembly should be written. `warp_affine` can maybe be used from [dav1d](https://code.videolan.org/videolan/dav1d/blob/master/src/x86/mc.asm).

EwoutH:
@spawlows @kirithika7 @lzhangnj @srikanthkurapati @agopikrishna13 or anyone else: If you are already working on one of those functions or plan on doing so, please tell. Also let me know if there are any important functions missing.

Also please open PRs sooner than later, even if they're still work in progress (just start them with WIP:) so that we can track them.

hassount:
Ok Let's use this issue to track the kernels we are working on.

hassount:
the ExtSadCalculation kernel has been optimized but it has a bug and so the optimizations is currently turned off until the bug is fixed

hassount:
@ttrigui @spawlows 
are all of these items done ?

ttrigui:
> @ttrigui @spawlows
> are all of these items done ?
Yes only convolve_2d_for_intrabc still not using optimized code it is just a matter of replacing
eb_av1_convolve_2d_sr_c
eb_av1_convolve_x_sr_c
[...]

PR #377, opened 2019-06-14
[SIMD] Port av1_inv_txfm_avx2.c from aom

lzhangnj:
Related to Issue #262


joelsole:
There seemed to be a mismatch between the inverse HBD transform in libaom and svt. Is this PR solving that?

ttrigui:
@joelsole I don't think it does. This is just porting the avx2 version of av1_inv_txfm_add. So far it produces the same bitstream as the reference.

nijilk:
> 
> 
> There seemed to be a mismatch between the inverse HBD transform in libaom and svt (checked using the svt-av1 decoder @nijilk ). Is this PR solving that?

@joelsole 
As @ttrigui mentioned, I don't think this is related to the issue I faced in decoder. I will share the details of that issue.

PR #397, opened 2019-06-20
[SIMD] Update self guided restoration x86 optimization

lzhangnj:
Related to Issue #262


hassount:
@lzhangnj can you please fix the warnings on this PR ?

PR #409, opened 2019-06-25
De-coupled completely the HME/ME/Subpel used for ME Inter prediction …

hguermaz:
## Description
De-coupled completely the HME/ME/Subpel used for ME Inter prediction and the one used for ALTREF Temporal Filtering.

## Author

@hguermaz 

## Type of change

Code restructuring .

## Tests and performance
* Checked bitstream exactness.
* Speed & memory footprint: no difference.


Source/Lib/Common/Codec/EbMotionEstimationProcess.c
```diff
@@ -179,11 +179,46 @@ EbErrorType signal_derivation_me_kernel_oq(
...
+        // Set HME flags
+        context_ptr->me_context_ptr->enable_hme_flag = picture_control_set_ptr->enable_hme_flag;
+        context_ptr->me_context_ptr->enable_hme_level0_flag = picture_control_set_ptr->enable_hme_level0_flag;
+        context_ptr->me_context_ptr->enable_hme_level1_flag = picture_control_set_ptr->enable_hme_level1_flag;
+        context_ptr->me_context_ptr->enable_hme_level2_flag = picture_control_set_ptr->enable_hme_level2_flag;
+
+        // Set the default settings of subpel
+        if (picture_control_set_ptr->sc_content_detected)
+            if (picture_control_set_ptr->enc_mode <= ENC_M1)
+                context_ptr->me_context_ptr->use_subpel_flag = 1;
```
mafonsoN: Isn't it better to set the subpel settings along with the other speed-dependent settings in svt?

hguermaz: The multi-mode signal(s) used by more that 1 process should be signaled under the PCS, and derived under signal_derivation_multi_processes_oq(). However the multi-mode signal(s) used by only 1 process could be signaled under the process context, and derived under signal_derivation_PROCESS_NAME_kernel_oq(). The multi-mode signal use_subpel_flag is actually used by only the ME process, and therefore use_subpel_flag has been moved from the PCS to the ME context, and the derivationof use_subpel_flag has been moved from signal_derivation_multi_processes_oq() to signal_derivation_me_kernel_oq(). The performed clean-up helped the decoupling operation; before the clean-up the decoupling could not happen without adding a dedicated multi-mode signal for temporal filtering case under PCS (i.e. fp_use_subpel_flag), but not after the clean-up.

PR #410, opened 2019-06-26
travis: Fix Binary Identical test

1480c1:
The Binary test weren't actually running because the master branch was not actually being rebuilt

Moved the workspace dir to $HOME so that way, there doesn't have to be a lot of moving around

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,20 +150,21 @@ matrix:
     - name: Binary Identical?
       stage: test
       script:
-       - *base_script
-       - mv -t $HOME akiyo_cif.y4m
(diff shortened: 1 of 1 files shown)


hassount:
@EwoutH 
LGTM

EwoutH:
Can you give me an example of a Travis build where the master branch wasn't build?

1480c1:
Any of the pr builds with the the new job. I noticed it when cmake said up to date for the executable instead of installing and the lack of recompiling

EwoutH:
Ah I see the problem, thanks for fixing.

LGTM

PR #421, opened 2019-06-28
Travis: Add Werror CFLAG when building SVT-AV1

### LINKED ISSUE #419: Make Travis fail on warnings
After fixing #417 we would like to impose that checks do not pass if there are warnings in the build.

We need to add such to travis only where the build fails on any warning and rebase all PRs to make sure we don't introduce any warnings in the master anymore.

1480c1:
Closes #419
Relies on #417

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,8 @@ install:
   #- if [ $TRAVIS_OS_NAME = osx ]; then brew install --HEAD valgrind; fi
 base_script: &base_script |
   cd $TRAVIS_BUILD_DIR/Build/linux/$build_type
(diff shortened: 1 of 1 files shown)


hassount:
Thank you very much @1480c1 it seems that it's working. Can we restrict this to gcc and one of the xcode that does not have warnings?

1480c1:
Do you mean clang?

1480c1:
Or do you mean macOS?

joelsole:
clang MacOs has no warnings => https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/211980307

kylophone:
LGTM, will merge today.

Issue #424, opened 2019-06-29
Rainbow color artifacts


maikmerten:
version: git master 9e67b31961c59166cbeca85b9838061607c9ad09 
os: Ubuntu 19.04, 64 bit, x86-64

My test encodes with SVT-AV1 have some very nice rainbow-colored artifacts.

Input file: https://media.xiph.org/video/derf/y4m/park_joy_420_720p50.y4m

Encoded file: https://people.xiph.org/~maikmerten/av1/svt/parkjoy-720p-svt-rainbow.av1

Transcoded to what I'm seeing: https://people.xiph.org/~maikmerten/av1/svt/parkjoy-720p-svt-rainbow.av1.transcode.webm

Encoder parameters and output:  https://people.xiph.org/~maikmerten/av1/svt/parkjoy-720p-svt-rainbow.av1.txt

I'm using ffmpeg to what the file. Ubuntu's ffmpeg is configured to decode AV1 using libaom and appears to produce nominal results with test files created with other encoders.

edit: Appears to be a duplicate of https://github.com/OpenVisualCloud/SVT-AV1/issues/400 - which I missed. Sorry for the noise.

kylophone:
Closing this for now to avoid duplication. This is being tracked in #400.

PR #426, opened 2019-07-02
Fix chroma artifacts in high density presets

### LINKED ISSUE #400: Abnormality of Chroma in a video with intense movement.
SVT-AV1 ac75c3e

Commandline
``ffmpeg -i park_joy_1080p50_2s.y4m  -an -nostdin -pix_fmt yuv420p -f yuv4mpegpipe - | SvtAv1EncApp -i stdin -n 100 -enc-mode 8 -intra-period 63 -q 44 -b park_joy_1080p50_2s_svt-av1_q44.ivf``

input video
https://people.xiph.org/~tdaede/sets/video-hd-1/
park_joy_1080p50_2s.y4m

output video
https://drive.google.com/open?id=1XRxUqw-nuiX4KjyHwkg0JUHMneihqI4C
名称未定-2

hassount:
## Description
Fix the atb check in the encode pass to separate the 10-bit and 8bit paths rather than atb_mode paths

## Issues
Closes #400 
Ref #424

### DIFF
--- a/Source/Lib/Common/Codec/EbCodingLoop.c
+++ b/Source/Lib/Common/Codec/EbCodingLoop.c
@@ -2510,7 +2510,7 @@ EB_EXTERN void av1_encode_pass(
                     // Partition Loop
                     context_ptr->txb_itr = 0;
                     // Transform partitioning path (INTRA Luma/Chroma)
-                    if (picture_control_set_ptr->parent_pcs_ptr->atb_mode && cu_ptr->av1xd->use_intrabc == 0) {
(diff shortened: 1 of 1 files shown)


joelsole:
I don't understand this. Please, could you extend the explanation? How is this change solving the issues?

hassount:
> I don't understand this. Please, could you extend the explanation? How is this change solving the issues?

There are 2 paths in the encode pass:
- 8 bit path (this includes the atb on and atb off)
- 10 bit and intra BC path  (this only has the atb off path without the atb support)
Now for 8bit, we need to use the 8bit path whether atb is on or off because the support is assumed. For 10-bit we need to use the atb off path for now until we add that support.

PR #427, opened 2019-07-04
appveyor: Fix artifacts

### LINKED ISSUE #425: Appveyor artifcts
@1480c1 
starting from this commit / PR #276, Appveyor does not upload any artifacts after the build, any thoughts ?

_Originally posted by @hassount in https://github.com/OpenVisualCloud/SVT-AV1/pull/276#issuecomment-507819775_

1480c1:
Fixes #425 
Not ideal, should be fixed CMake side, but due to CMake and multi config generators, I'm unable to force it  to be set properly

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,10 +11,14 @@ build:
   project: svt-av1.sln
 
 artifacts:
-    - path: bin\Release\SvtAv1*.exe
-      name: $(APPVEYOR_PROJECT_NAME)
-    - path: bin\Release\SvtAv1*.dll
-      name: $(APPVEYOR_PROJECT_NAME)
+  - path: bin\**\SvtAv1EncApp.exe
+    name: $(APPVEYOR_PROJECT_NAME)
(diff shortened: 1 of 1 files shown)


EwoutH:
This PR does way more:

- Switches from Yasm to Nasm
- Switches from MSBuild to CMake

Do you have any idea if there is a performance impact of those changes? It would be good to do some testing before making them.

1480c1:
I could switch back to using task, I'm only using nasm since it's easier to maintain is it's using chocolatey (yasm is outdated on choco) 

calling cmake --build with call msbuild in the background

hassount:
@1480c1 thank you for the work around, could you please restrict the changes to addressing the appveyor artifacts, then we can propose / evaluate the nasm changes in another PR

1480c1:
Sure, once I'm on a computer

1480c1:
Done

hassount:
thnx @1480c1

PR #428, opened 2019-07-04
travis: Fix Werror, set os and compiler for BinIden

1480c1:
Defaults build_type to release and generator to Unix Makefiles
Sets OS and compiler for BinIden job to always use the same compiler and os

Forgot to export CFLAGS


1480c1:
Actually, need to exclude gstreamer from Werror

1480c1:
I'm having problems pushing to Werror. Going to reopen with a different branch name

PR #431, opened 2019-07-04
Set compiling option to SSE4.1

ttrigui:
## Description
Set the high level project vectorization depth to sse4.1 [close to 100% of the processors SVT would run on]. Such would restrict the avx2 and avx512 usage to the avx2 and avx512 libraries allowing for better cross-platform execution.


1480c1:
What is this change for?

CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ if(NOT MSVC)
     list(INSERT flags_to_test 0 -Wall)
         option(NATIVE "Build for native performance (march=native)" ON)
         if(NATIVE)
-        list(APPEND flags_to_test -march=native)
+        list(APPEND flags_to_test -march=sse4.1)
```
1480c1:
```suggestion
        list(APPEND flags_to_test -march=native)
    else()
        list(APPEND flags_to_test -march=sse4.1)
```

1480c1: The first march is for people who wish to build using all optimization available to their cpu family type.

As I asked before, why are you setting march=sse4.1? I do not see such an option in the [gcc docs](https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/x86-Options.html#x86-Options)

Helenerineium:
```suggestion
        list(APPEND flags_to_test -march=native)
    else()
        list(APPEND flags_to_test -msse4.1)
```

Helenerineium:
```suggestion
        list(APPEND flags_to_test -march=native)
    else()
        list(APPEND flags_to_test -msse4.1)
```

DesWurstes: I believe they meant [this](https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Source/Lib/Common/ASM_SSE4_1/CMakeLists.txt#L18)

1480c1: I would hope so

1480c1:
@ttrigui or @hassount, is this trying to use -msse4.1 in toplevel, or set a minimum arch to a cpu family with sse4.1?

hassount:
> @ttrigui or @hassount, is this trying to use -msse4.1 in toplevel, or set a minimum arch to a cpu family with sse4.1?

@1480c1 
The idea here is not to use avx2 or avx512 compiled in any libraries apart from the specific avx2 and avx512 libraries. 
So -msse4.1 in the top level would probably be the way to go

1480c1:
> > @ttrigui or @hassount, is this trying to use -msse4.1 in toplevel, or set a minimum arch to a cpu family with sse4.1?
> 
> @1480c1
> The idea here is not to use avx2 or avx512 compiled in any libraries apart from the specific avx2 and avx512 libraries.
> So -msse4.1 in the top level would probably be the way to go

Theoretically, since none of the CMakeLists other than the ASM_AVX2 and the soon to be ASM_AVX512 should have mavx2 or avx512 related cflags other than the march=native (depending on the CPU), avx2 and avx512 functions should not be used outside of those sources.

Applying sse4.1 will not do much other than making it so that we can remove it from ASM_SSE4_1 since the toplevel doesn't add any source files to compile directly

CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ if(NOT MSVC)
     list(INSERT flags_to_test 0 -Wall)
         option(NATIVE "Build for native performance (march=native)" ON)
```
1480c1:
```suggestion
        option(NATIVE "Build for native performance (march=native)")
```
It would instead be better to do this to not build for the native cpu arch by default

CMakeLists.txt
1480c1:
```suggestion
        list(APPEND flags_to_test -march=native)
```

CMakeLists.txt
```diff
@@ -74,9 +74,9 @@ set(flags_to_test
     -D_FORTIFY_SOURCE=2)
 if(NOT MSVC)
     list(INSERT flags_to_test 0 -Wall)
-        option(NATIVE "Build for native performance (march=native)" ON)
+        option(NATIVE "Build for native performance (march=native)")
         if(NATIVE)
-        list(APPEND flags_to_test -march=native)
+        list(APPEND flags_to_test -march=sse4.1)
```
1480c1:
```suggestion
        list(APPEND flags_to_test -march=native)
```
`-march=sse4.1` is still not an actual option, plus this is supposed to be reserved for the native option (I messed up on the indentation)

1480c1:
Is this ready to merge?

hassount:
doesn't seem to be passing due to the base master. @ttrigui can you please rebase the branch to make sure we don't break anything ?

PR #432, opened 2019-07-04
Frame end CDF update

anaghdin:
## Description
Add the frame end CDF update. Uses the CDF from previous frames in mode decision and entropy coding. 

## Authors
@anaghdin

## Type of change
New Feature

## Tests and performance
Expected Average PSNR-SSIM BD-rate gain in the range of -1.3% on the AOM test set, using 4 QP values: {20, 32, 43 and 55}.

No Speed impact.


Source/Lib/Common/Codec/EbPictureControlSet.h
```diff
@@ -13870,6 +13870,11 @@ extern "C" {
         FRAME_CONTEXT * ec_ctx_array;
         struct MdRateEstimationContext* rate_est_array;
         uint8_t  update_cdf;
+#if ENABLE_CDF_UPDATE
+        FRAME_CONTEXT           ref_frame_context[REF_FRAMES];
```
hassount: this is a type and so it should be CamelCased

anaghdin: It is already defined in the code.

hassount: agreed

hassount:
@joelsole @mafonsoN 
any issues with this moving forward ?

hassount:
@anaghdin there seems to be some warnings, could you please fix them or rebase to make sure we're not adding warnings?

https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/213397869

anaghdin:
@hassount  done

PR #434, opened 2019-07-05
Fix the stability issue caused by BRC

lijing0010:


lijing0010:
Hard to reproduce, but will crash or hang after a few million frames.
Easier to reproduce on SVT-HEVC which runs faster and has the similar mechanism :)

lijing0010:
Yes, you are right. there are 2 more places which svt-hevc doesn't have.
updated the patch

hassount:
@anaghdin, is this good to go now ?

anaghdin:
@hassount Yes. looks go to me. Thanks

PR #443, opened 2019-07-09
travis: Fix Werror, set os and compiler for BinIden

1480c1:
Defaults build_type to release and generator to Unix Makefiles
Sets OS and compiler for BinIden job to always use the same compiler and os

Forgot to export CFLAGS

Original PR was #428, can't push to Werror anymore even when deleting it everywhere

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,15 +43,15 @@ before_install:
   - cd $TRAVIS_BUILD_DIR
   - wget -nc https://raw.githubusercontent.com/OpenVisualCloud/SVT-AV1-Resources/master/video.tar.gz || wget -nc http://randomderp.com/video.tar.gz
   - tar xf video.tar.gz
-  - mkdir -p $TRAVIS_BUILD_DIR/Build/linux/$build_type
+  - mkdir -p $TRAVIS_BUILD_DIR/Build/linux/${build_type:-Release}
 install:
   - pip install --user cpp-coveralls
   - export PATH=/Users/travis/Library/Python/2.7/bin:${PATH}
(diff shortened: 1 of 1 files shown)


kylophone:
Alright. So this adds `-Werror` to everything that's not MacOS? Looks like there are some warnings still in the Travis build. I guess we would have expected it to have failed, right?

1480c1:
> `Source/Lib/Common/ASM_SSE2/highbd_intrapred_sse2_.asm` should not be part of this diff.

Accidental save by vscode 

> Alright. So this adds `-Werror` to everything that's not MacOS? Looks like there are some warnings still in the Travis build. I guess we would have expected it to have failed, right?

Yea, mainly because for some reason, when `-Werror` is applied for macOS, one of the asm file gives an error https://travis-ci.com/1480c1/SVT-AV1/jobs/213675006#L1366

Seems there are still array bounds stuff from clang, but they are excluded for now

kylophone:
> > `Source/Lib/Common/ASM_SSE2/highbd_intrapred_sse2_.asm` should not be part of this diff.
> 
> Accidental save by vscode
> 
> > Alright. So this adds `-Werror` to everything that's not MacOS? Looks like there are some warnings still in the Travis build. I guess we would have expected it to have failed, right?
> 
> Yea, mainly because for some reason, when `-Werror` is applied for macOS, one of the asm file gives an error https://travis-ci.com/1480c1/SVT-AV1/jobs/213675006#L1366
> 
> Seems there are still array bounds stuff from clang, but they are excluded for now

Ah, ok. Cool. I guess this can be merged once you remove the extraneous file from the diff.

1480c1:
Done

kylophone:
Merged. Thanks.

PR #412, opened 2019-06-26
Interpred tools test

Cidana-Developers:

PR #444, opened 2019-07-09
Fix assembly implementation of sad_loop_kernel_sparse

Cidana-Developers:
1. add c implementation
2. enable sparse loop for all the block size, currently
   the sparse loop is enabled only when width is 16 and height
   is less than or equal to 16.


Cidana-Developers:
> Could you please add unit test between C and assembly?

The unit test is added in PR #412 FYI.

hassount:
@Cidana-Developers 
seems that this is ready to be merged, can you please rebase >?

Cidana-Developers:
> @Cidana-Developers
> seems that this is ready to be merged, can you please rebase >?
Sure, I also removed c implementation to avoid potential conflicts since its included in PR #412.

hassount:
@ttrigui can we make sure to integrate this in m0-test

PR #448, opened 2019-07-10
Change CMakeLists.txt to have all debug symbols in lldb

joelsole:


kylophone:
Does anyone know why there were multiple optimization flags set in the first place? @1480c1?

https://stackoverflow.com/a/5557417

1480c1:
`-O0` is a common disable opt flag for clang and gcc, `-Og` is a opt flag that enables debugging information, but does a little bit of optimization that is not always available.

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html 
``` none
Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

Like -O0, -Og completely disables a number of optimization passes so that individual options controlling them have no effect. Otherwise -Og enables all -O1 optimization flags except for those that may interfere with debugging:
```

https://clang.llvm.org/docs/CommandGuide/clang.html
``` none
-Og Like -O1. In future versions, this option might disable different optimizations in order to improve debuggability.
```

kylophone:
> `-O0` is a common disable opt flag for clang and gcc, `-Og` is a opt flag that enables debugging information, but does a little bit of optimization that is not always available.

Yeah, but we were setting both for some reason. In any case it seems `-Og` doesn't play nicely with lldb while `-O0` does.

kylophone:
I think idea it would also be a good idea if we make the cmake build on travis verbose by default, so we can see all the compiler flags in the log.

1480c1:
Yes, if the compiler accepts both, it will accept them in that order with Og being the last one, I only added Og because I remembered that it existed for debugging. This change is fine though

1480c1:
> I think idea it would also be a good idea if we make the cmake build on travis verbose by default, so we can see all the compiler flags in the log.

We can add `-DCMAKE_VERBOSE_MAKEFILE=1` to the cmake command

kylophone:
> > I think idea it would also be a good idea if we make the cmake build on travis verbose by default, so we can see all the compiler flags in the log.
> 
> We can add `-DCMAKE_VERBOSE_MAKEFILE=1` to the cmake command

Yep. Do you have a second to put together a quick PR for this?

1480c1:
> > > I think idea it would also be a good idea if we make the cmake build on travis verbose by default, so we can see all the compiler flags in the log.
> > 
> > 
> > We can add `-DCMAKE_VERBOSE_MAKEFILE=1` to the cmake command
> 
> Yep. Do you have a second to put together a quick PR for this?

Yea

kylophone:
Merged. Thanks.

PR #449, opened 2019-07-10
Travis: add verbose makefile to print commands

1480c1:
Prints all commands including c flags etc, might make the travis log a bit log

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ install:
   #- if [ $TRAVIS_OS_NAME = osx ]; then brew install --HEAD valgrind; fi
 base_script: &base_script |
   cd $TRAVIS_BUILD_DIR/Build/linux/${build_type:-Release}
-  cmake $TRAVIS_BUILD_DIR -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release} ${CMAKE_EFLAGS[@]}
+  cmake $TRAVIS_BUILD_DIR -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release} -DCMAKE_VERBOSE_MAKEFILE=1 ${CMAKE_EFLAGS[@]}
   cmake -j $(if [ $TRAVIS_OS_NAME = osx ]; then sysctl -n hw.ncpu; else nproc; fi) --build . &&
   sudo cmake --build . --target install && cd $TRAVIS_BUILD_DIR
 coveralls_script: &coveralls_script |


1480c1:
@kylophone or @hassount, should unused-variable also be added to the Werror exclusion list?

kylophone:
> @kylophone or @hassount, should unused-variable also be added to the Werror exclusion list?

No, I don't think so. The whole point of this is to avoid committing code that causes compiler warnings.

1480c1:
> > @kylophone or @hassount, should unused-variable also be added to the Werror exclusion list?
> 
> No, I don't think so. The whole point of this is to avoid committing code that causes compiler warnings.

Okay, because travis found in ASM_AVX2/highbd_inv_txfm_avx2.c
``` none
In file included from /home/travis/build/OpenVisualCloud/SVT-AV1/Source/Lib/Common/ASM_AVX2/highbd_inv_txfm_avx2.c:16:0:
/home/travis/build/OpenVisualCloud/SVT-AV1/Source/Lib/Common/ASM_SSSE3/av1_inv_txfm_ssse3.h:90:20: error: ‘NewSqrt2list’ defined but not used [-Werror=unused-variable]
     static int32_t NewSqrt2list[TX_SIZES] = { 5793, 2 * 4096, 2 * 5793, 4 * 4096,
```

hassount:
> > @kylophone or @hassount, should unused-variable also be added to the Werror exclusion list?
> 
> No, I don't think so. The whole point of this is to avoid committing code that causes compiler warnings.

If we're trying to not have compiler warnings then I believe it's best that it gets added. To disable it in the cmake lists. what we'd like to avoid is having a list of these every time we compile the code which would make it hard to distinguish which ones are unused and which ones are more severe warnings

1480c1:
> To disable it in the cmake lists.

What does this sentence refer to? Not exactly sure of my interpretation

1480c1:
Also @hassount, can you look at https://github.com/OpenVisualCloud/SVT-AV1/blob/58d8e6f2b28617a04ebee3be3a0b7146299233c4/Source/Lib/Common/Codec/EbMotionEstimationProcess.c#L153-L214 and tell me if that if the indentation of if and else is correct? gcc 7.4.0 is giving misleading-indentation warnings

kylophone:
> and tell me if that if the indentation of if and else is correct? gcc 7.4.0 is giving misleading-indentation warnings

These blocks need brackets.

hassount:
> > To disable it in the cmake lists.
> 
> What does this sentence refer to? Not exactly sure of my interpretation

my understanding is that you could instruct the compiler to specifically suppress a set of warnings and such could be done in the CMakeLists, unless such is not possible anymore.

1480c1:
It is possible, although I was referring to adding it to Travis rather than to the cmakelists since that is where werror is being set

mafonsoN:
The travis output is indeed very long with verbose. Since that Werror is working, is this still necessary?

1480c1:
Unsure if it really is, since the only times we would need to see the commands would be if we need to confirm what flags are used etc

mafonsoN:
Closing this for now. Thanks @1480c1. We can consider enabling verbose prints later.

PR #452, opened 2019-07-12
Fix different outputs between Windows and Linux

spawlows:
Windows: sizeof(InterpFilters) is 4, Linux: sizeof(InterpFilters) is 1

Kernel: av1_make_interp_filters() Output value was be casted on uint8.


kylophone:
lgtm

PR #451, opened 2019-07-12
Avx512 support and spatial full distortion

lzhangnj:
Add AVX512 lib and spatial_full_distortion_kernel()

For spatial_full_distortion_kernel(), AVX512 is 0% to 40% faster than AVX2 version for width 64 and up.

Related to PR #414

PR #455, opened 2019-07-14
AVX512: Don't use apt clang and only compile avx512 code if function is defined

1480c1:
@lzhangnj, not for master branch

installing the clang and llvm from ubuntu's repo messed up with the clang provided with travis' images

Add CMake code to detect if `_mm512_extracti64x4_epi64` is defined in `immintrin.h` and add the define to not do avx512 stuff. There is a different function for checking functions, types, etc in c files.

For #451

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ addons:
       - valgrind
       - libgstreamer-plugins-base1.0-dev
       - libgstreamer1.0-dev
-      - llvm
-      - clang
   homebrew:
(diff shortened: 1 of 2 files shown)


1480c1:
Tested in an ubuntu 16.04 container with only cmake 3.5.1, clang 3.8.0, and yasm 1.3.0

lzhangnj:
@1480c1 Thanks a lot! Great!

PR #457, opened 2019-07-15
Improve SIMD alignment

lzhangnj:
Went through all AVX2 code and identified some memories which should be aligned.


hassount:
@mafonsoN @kylophone merging by EOD there are no issues with this

mafonsoN:
> @mafonsoN @kylophone merging by EOD there are no issues with this

Looks fine to me

PR #461, opened 2019-07-15
re-write sequence header parser

Cidana-Developers:
reuse the parser in svt av1 decoder


Source/Lib/Decoder/Codec/EbDecParseObu.c
```diff
@@ -2464,3 +2466,40 @@ EbErrorType decode_multiple_obu(EbDecHandle *dec_handle_ptr, uint8_t **data, siz
     }
     return status;
 }
+
+EB_API EbErrorType eb_get_sequence_info(
+    const uint8_t *obu_data,
+    uint32_t       size,
```
kylophone: size_t

edmond-zhu: thank you for reminding, it is changed.

PR #462, opened 2019-07-15
Fix encoder parameter setup issue in e2e test

Cidana-Developers:
Fix the issue that the encoder parameters does not take effect
in SvtAv1E2ETest.cc


Cidana-Developers:
Rebased and add new e2e test based on this pr.

PR #330, opened 2019-06-04
Improved sub-pel refinement

NaderMahdi:
** Improved subpel refinement**
- Added the ability to perform subpel refinement around multi-integer position.
- Added the ability to perform the quarter-pel around all

The changes are ON for the research mode and the M0 mode only.

Average bdrate gain 0.5%/0.5% for PSNR/SSIM bdrate.
Average speed drop is about 12%.
Average memory increase is about 0%.

PR #464, opened 2019-07-15
Sub-pel refinement optimization

NaderMahdi:
**Sub-pel refinement optimization**
Restructure the Sub-pel refinement to be done in an NxM window around the best integer-pel and where NxM is set to 3x3 by default.

The recorded data on the 360p clips is 11% speedup in encoding time for 0.1% bdrate gain and no impact on the memory.


Source/Lib/Common/Codec/EbMotionEstimation.c
```diff
@@ -4198,7 +4198,521 @@ static void FullPelSearch_LCU(MeContext *context_ptr, uint32_t listIndex,
...
+                    *p_best_mv =
+                        ((uint16_t)half_mv_y[3] << 16) | ((uint16_t)half_mv_x[3]);
+                }
+            }
+#endif
+            // TL position
+            search_region_index =
+                search_index_x +
+                (int16_t)context_ptr->interpolated_stride * search_index_y;
+            distortion_topleft_position =
```
kylophone: Avoid all these ternarys. Could be written in a more readable fashion using if/else blocks or maybe a switch.

mafonsoN:
@NaderMahdi can you please clarify the changes here compared to #330 ?

Source/Lib/Common/Codec/EbMotionEstimation.c
NaderMahdi: Addressed - Thanks.

NaderMahdi:
> @NaderMahdi can you please clarify the changes here compared to #330 ?
@mafonsoN basically no difference, just the implementation is different. #330 is more a simulation code hoewever the #464 is production code.

Issue #483, opened 2019-07-20
logical_processors (-lp) ignored on macOS


brion:
The logical_processors option (-lp) is ignored on macOS builds; it always uses the maximum number of logical processors (12 on my test machine).

The option works as expected on Linux builds.

PR #484, opened 2019-07-20
Fix for logical processors being ignored on macOS

brion:
The enforcement of the -lp option for logical processor count was #ifdef'd to Windows and Linux only, causing the option to be ignored on macOS builds.

Fixes https://github.com/OpenVisualCloud/SVT-AV1/issues/483


hassount:
@kylophone  merging this if you have no issues.

PR #488, opened 2019-07-22
Add unit test for in-loop filter

Cidana-Developers:
1. Test the sse2 implementation of in-loop filter.


Cidana-Developers:
> #if 0
> can we replace that with a properly defined macro ?
Good suggestion, it's done.

andrey-norkin:
The windows test has not succeeded (file download)

PR #463, opened 2019-07-15
symbols: fix conflicts with aom

xuguangxin:
We use many functions from liboam. We will see many duplicate define errors
if we enable liboam static library + svtav1.
this will fix #453

@jsunintel , @hassount @kylophone 
please help review this meet the requirement or not
thanks

PR #491, opened 2019-07-23
appveyor: download yasm.exe from github instead of tortall.net

xuguangxin:
www.tortall.net is down.We'd better use github to download the yasm file.
I updloaded the yasm 1.3 file here:
https://github.com/OpenVisualCloud/SVT-AV1/pull/463#issuecomment-514077315

Closes #490

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ configuration: Release
 
 install:
     - git submodule update --init
-    - appveyor DownloadFile http://www.tortall.net/projects/yasm/releases/yasm-1.3.0-win64.exe -FileName yasm.exe
+    - appveyor DownloadFile https://github.com/yasm/yasm/releases/download/v1.3.0/yasm-1.3.0-win64.exe -FileName yasm.exe
     - set PATH=%PATH%;%APPVEYOR_BUILD_FOLDER%
     - cmake . -G "Visual Studio 16 2019" -A x64 -DCMAKE_ASM_NASM_COMPILER="yasm.exe" -DBUILD_TESTING=ON
 


xuguangxin:
This will replace https://github.com/OpenVisualCloud/SVT-AV1/pull/490

1480c1:
@kylophone, would it be possible to add it to https://github.com/OpenVisualCloud/SVT-AV1-Resources?

xuguangxin:
yeah, it's a better place. I can send the pr to SVT-AV1-RESOURCE if @kylophone agreed.

kylophone:
Sure, do we really want to cache our own yasm, though? Does appveyor have some sort of package manager?

1480c1:
It has chocolaty, but it's community supported and outdated to 1.2.0, plus it still tries to pull from official sources

1480c1:
There should have been msys2's pacman to download mingw-w64 yasm, but it seems that they have it for vs2017 images and not for vs2019 images yet.

j-mas:
Could you please find a solution that AppVeyor the builds are formed correctly again! Please find a solution for www.tortall.net and the yasm 1.3 file.

1480c1:
That's the purpose of this pr

xuguangxin:
@1480c1 , If we put yasm to git repo, no matter I keep it as exe or [txt](https://ci.appveyor.com/project/OpenVisualCloud/svt-av1/builds/26201307#L7), it always shows a strange [error](https://ci.appveyor.com/project/OpenVisualCloud/svt-av1/builds/26201307#L276).
Seems we still need use https://github.com/OpenVisualCloud/SVT-AV1/pull/463#issuecomment-514077315

hassount:
I don't think that hosting a binary in the repo is right as it can present a security risk to some extend. is there another solution ?

xuguangxin:
@hassount , 
We have the other two solutions here.
one is https://github.com/OpenVisualCloud/SVT-AV1/pull/490, but it adds a dependency for msys2. It's no different with original one if msys2 site down. 
Another is downloading from https://github.com/OpenVisualCloud/SVT-AV1/pull/463#issuecomment-514077315.
I prefer the second one because it only depends on github.
For security issue, we only use it in appveyor, it's maybe not a big problem.
thanks

j-mas:
www.tortall.net is back !

xuguangxin:
@j-mas , great, thanks for remind!
But the problem is there, if we depend on an external resource, it will down in the future. All CI will be ceased.
@hassount how you think about this?

j-mas:
@xuguangxin , yes you are right. It is not a good basis to be dependent on a third party!

DesWurstes:
Why not https://github.com/yasm/yasm/releases/download/v1.3.0/yasm-1.3.0-win64.exe ?

xuguangxin:
@DesWurstes , it only released yesterday. https://github.com/yasm/yasm/releases. 
I will change it.
thanks

xuguangxin:
@hassount now yasm has an official release on github. 
I have pointed our yasm.exe to it.
Please help review. 
thanks

1480c1:
@xuguangxin, one thing you can do is in the original text, add `Closes #490` somewhere so that once this is merged, it will automatically close that one. That way it is easier to track which PRs cover/overlap/fixes/closes in terms of applied effects for other PRs or issues

xuguangxin:
@1480c1 , thanks for the good suggestion. Done.

1480c1:
I ment in the original post, not commit

xuguangxin:
@1480c1 , oh, sorry. seems it's added by mafonsoN. thanks, mafonsoN
thanks all for help on this.

PR #492, opened 2019-07-23
Optimizing interp_cubic(). Precalculated data added. Work ~5 times fa…

spawlows:
…ster


kylophone:
Great. Do you think this needs a unit test to check for equivalence?

ttrigui:
@kylophone I assume that @spawlows has already done his testing. I already looked at the code and done some testing and it looks fine to me. But I agree this needs extensive testing.

kylophone:
> @kylophone I assume that @spawlows has already done his testing. I already looked at the code and done some testing and it looks fine to me. But I agree this needs extensive testing.

Yeah, I don't doubt that. A unit test proves equivalence now, but also catches anything that might diverge in the future. Both functions are still in the code, so it could just be a simple loop and compare test.

PR #495, opened 2019-07-26
Add AVX512 code for high bd forward transform

kirithika7:
This PR has AVX512 high bd forward transform kernels for sizes 16x16,32x32,64x64,64x32,32x64,64x16,16x64,32x16,16x32 and their unit tests.The performance numbers are given in the commit messages,

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -2218,7 +2218,7 @@ extern    uint32_t                   lib_mutex_count;
 
 extern    uint32_t                   app_malloc_count;
 
-#define ALVALUE 32
+#define ALVALUE 64
 
 #define EB_ADD_APP_MEM(pointer, size, pointer_class, count, release, return_type) \
     do { \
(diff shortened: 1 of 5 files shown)


kirithika7:
It builds and executes fine with gcc 7.3.0 but the Travis test with gcc 5.4.0 fails.

1480c1:
It is possible to update the gcc and clang using the apt sources provided by ubuntu and llvm or using a docker,

kirithika7:
> It is possible to update the gcc and clang using the apt sources provided by ubuntu and llvm or using a docker,

I have updated the gcc and tested locally.It is building and executing without any errors.

Source/Lib/Common/Codec/EbDefinitions.h
```diff
@@ -40,7 +40,7 @@ extern "C" {
 #endif
 
 // Internal Marcos
-#define NON_AVX512_SUPPORT
+//#define NON_AVX512_SUPPORT
```
hassount: this should be put back on for now

JackZhouVCD:
After applying this PR, the encoder crashes during the dealloc process. This issue was addressed in #499 which is later on in the master than the base commit shown in this PR. Please rebase to sync up with master.

PR #497, opened 2019-07-28
Use a GCC version that is still maintained in unit tests

DesWurstes:
Upgrades GCC to 7.4.

Fails because #511 is needed.

**EDIT**: Also needs #550


DesWurstes:
Issue #614 updates compiler versions.

PR #502, opened 2019-07-31
Added support for Screen content tools and OBMC in Decoder

nijilk:


nijilk:
@andrey-norkin @hassount @joelsole 
Hi,
Support for Screen content tools and OBMC is added in Decoder.
Please review.

Source/Lib/Decoder/Codec/EbDecObmc.c
```diff
@@ -0,0 +1,617 @@
...
+                mask[j], src0[i * src0_stride + j], src1[i * src1_stride + j]);
+        }
+    }
+}
+
+static INLINE void build_obmc_inter_pred_above(EbDecHandle *dec_handle,
+    PartitionInfo_t *pi, BlockSize bsize, int rel_mi_col, uint8_t above_mi_width,
+    uint8_t *above_tmp_buf[MAX_MB_PLANE], int above_tmp_stride[MAX_MB_PLANE],
+    uint8_t *curr_blk_recon_buf[MAX_MB_PLANE],
+    int32_t curr_recon_strd[MAX_MB_PLANE], const int num_planes)
```
andrey-norkin: It would make sense to spell stride instead of strd in the decoder. This is mostly for the reason that encoder and common modules usually use "stride" spelling, and it seems slightly confusing to have a different convention for the decoder modules. I have noticed that some decoder modules already use strd but it could be good time to change it before more code has been submitted.

nijilk: @andrey-norkin 
Thanks for the review.
This issue has been fixed. Now decoder uses stride, which is in sync with encoder and common usage.

PR #505, opened 2019-07-31
Fix the wrong profile message

lizhong1008:
There is no main10 profile for AV1

Signed-off-by: Zhong Li <zhong.li@intel.com>


lizhong1008:
@hassount, is it ok to merge if nobody against? It is a clear wrong need to be fix.

PR #507, opened 2019-07-31
Style check: split git grep for trs, use git ls-tree, sed, don't use travis_terminate

1480c1:
About 5 seconds quicker and doesn't immediately stop the job

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,27 +100,31 @@ matrix:
       before_install: skip
       install: skip
       script:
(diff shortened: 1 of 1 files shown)


EwoutH:
@hassount Could we merge this? Would help a lot of contributors finding their style errors quicker.

1480c1:
I would also second the merging of this as it would help in cases like 
 https://github.com/OpenVisualCloud/SVT-AV1/pull/609#issuecomment-531897835 where carriage returns are exactly obvious

PR #509, opened 2019-08-01
Optimizing kernels construct_best_sorted_arrays_md_stage_2/3.

spawlows:
Kernels work 2.7 faster on msvc and 4.3 faster on linux


mihas666:
Not bad,but it's too difficult for me

Issue #332, opened 2019-06-04
Assembly implementation of cfl_predict is incorrect.


Cidana-Developers:
There are two assembly functions for cfl prediction, one is cfl_predict_hbd_avx2 and the other is  cfl_predict_lbd_avx2, Both of them are incorrect. 

Following is the declaration of cfl_predict_hbd_avx2 :
```
void cfl_predict_hbd_avx2(
    const int16_t *pred_buf_q3,
    uint16_t *pred,// AMIR ADDED
    int32_t pred_stride,
    uint16_t *dst,// AMIR changed to 8 bit
    int32_t dst_stride,
    int32_t alpha_q3,
    int32_t bit_depth,
    int32_t width,
    int32_t height)
```
If pred_buf_q3 is cleared to zeros, and pred buffer (4x4) is:
```
10 3 1 2
2 8 9 5
4 0 14 6
15 3 6 8
```
The dst buffer will become the following array after execution:
```
10 10 10 10
10 10 10 10
10 10 10 10
10 10 10 10
```
however the content of dst should be the same with pred buffer.

#331 adds the intra pred test including cfl test. Meanwhile, these two assembly functions are not used in libaom.

hassount:
@ttrigui could you please take a look at this ?

lzhangnj:
Wrong issue. The mismatch is caused by unit test code and fixed by PR #543.

PR #541, opened 2019-08-14
Fix cfl_predict avx2 mismatch

lzhangnj:
Issue #332

pred[i] should update for each pixel i.


lzhangnj:
Wrong issue and abandon this PR. The mismatch is caused by unit test code and fixed by PR #543.

PR #543, opened 2019-08-15
Fix cfl unittest mismatch issue

Cidana-Developers:
1. Populated the dst buffer using the same value instead of random
   value, because the pred buffer is supposed to be populated by
   dc prediction before cfl prediction.


Cidana-Developers:
> Related to Issue #332 and PR #541. Duplicated issue with different point of view.
> So the question is, is pred always an array of a single constant value?

In this case, yes. In svt-av1 or libaom,  intra prediction will be built using dc mode before cfl prediction.

lzhangnj:
> In this case, yes. In svt-av1 or libaom, intra prediction will be built using dc mode before cfl prediction.

Thanks for confirm! I have closed Issue #332 and abandoned PR #541.

Issue #539, opened 2019-08-13
Encoder crashed when the SpeedControlFlag is set


Cidana-Developers:
commit id: d0f64ff
build type: release
platform: ubuntu 16.04
Steps:
```
./SvtAv1EncApp -i ../../test/vectors/kirland_640_480_30.yuv -w 640 -h 480 -speed-ctrl 1 -o ../../../test/recon.yuv stat-report 1 -b ../../../test/output.ivf
-------------------------------------------
SVT-AV1 Encoder
SVT [version]:  SVT-AV1 Encoder Lib v0.6.0
SVT [build]  :  GCC 5.4.0        64 bit
LIB Build date: Aug 13 2019 17:28:34
-------------------------------------------
Number of logical cores available: 12
Number of PPCS 72
-------------------------------------------
SVT [config]: Main Profile      Tier (auto)     Level (auto)
SVT [config]: EncoderMode                                                       : 8
SVT [config]: EncoderBitDepth / EncoderColorFormat / CompressedTenBitFormat     : 8 / 1 / 0
SVT [config]: SourceWidth / SourceHeight                                        : 640 / 480
SVT [config]: FrameRate / Gop Size                                              : 30 / 32
SVT [config]: HierarchicalLevels / BaseLayerSwitchMode / PredStructure          : 4 / 0 / 2
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange                  : CQP / 50 / 33 / 0
-------------------------------------------
Encoding          Segmentation fault (core dumped)
```

DesWurstes:
Even though I'm on macOS:

(Manual) Stack trace:

[...]

PR #544, opened 2019-08-15
Fix crash issue when speed control flag is enabled

Cidana-Developers:
1. update the cdf mode in signal_derivation_pre_analysis_oq;
2. always allocate ec_ctx_array and rate_est_array since cdf_mode might change.
3. initial the avail_blk_flag in init_sq_non4_block.


Cidana-Developers:
This pr will fix issue #539.

PR #552, opened 2019-08-19
CMake: Don't set default CMAKE_BUILD_TYPE if MSVC or Xcode

1480c1:


hassount:
@mafonsoN @joelsole merging if no issues on your end

PR #554, opened 2019-08-19
Cmakelists: Cleanup the main list

1480c1:
Add option to skip using yasm, add a warning if 32-bit, move options up, and match CFLAGS from MSVC to else


CMakeLists.txt
```diff
@@ -68,26 +83,29 @@ set(flags_to_test
...
-else()
+if(MSVC)
     list(APPEND flags_to_test /MP)
     set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} /O2")
     set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} /Od")
+else()
+    list(INSERT flags_to_test 0 -Wall)
+    option(NATIVE "Build for native performance (march=native)")
+    if(NATIVE)
+        list(APPEND flags_to_test -march=native -mtune=native)
```
DesWurstes: `march=native` implies `mtune=native`, so this could be redundant (or this is to make it explicit?)

1480c1: Actually, it doesn't. But yea, explicit

1480c1: I only found out recently that sometimes mtune isn't set to march

ferdnyc: @1480c1 If `-march` ever doesn't set `-mtune` that's a bug, it's [explicitly documented](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html) that:
> Specifying <code>-march=<i>cpu-type</i></code> implies <code>-mtune=<i>cpu-type</i></code>.

(`-mtune` can subsequently be set to something _different_ than `-march`, but that's the only time they should ever differ.)

1480c1: https://lemire.me/blog/2018/07/25/it-is-more-complicated-than-i-thought-mtune-march-in-gcc/ 
https://stackoverflow.com/questions/56870475/how-to-correctly-determine-march-and-mtune-for-intel-processors

In some cases (gcc is old, cpu is not fully known to the compiler), although the march may be set to native (or whatever gcc detects) mtune might be set to generic

ferdnyc: Interesting! Thanks. Though, reading through that, it strikes me that he missed a crucial step. He just _**assumes**_ that `-mtune=native` will give him `broadwell` tuning on his (unrecognized) `skylake` processor, but never actually checked the output of `gcc -march=native -mtune=native -Q --help=target|grep -- '-mtune='` to confirm.

If the compiler doesn't _recognize_ a processor, doing fallback `-march` makes sense (as it just controls the _available_ instructions sets, but doing fallback _tuning_ may not. (This was pointed out by one of the commenters as well.) In other words, in the cases where the arch is **unknown**, `-mtune=native` may very well (and very appropriately) give you the same thing as `-mtune=generic`. He never demonstrates otherwise.

EwoutH:
@1480c1 Could you rebase to trigger the updated Travis config?

1480c1:
done

1480c1:
Rebased to master

AlexBoswellVCD:
Found a 1% speed gain averaged over modes 1,4 and 8. No memory deviations across these modes. No loss when comparing modes 4 and 8, there was a r2r issue in M1 Master using a 360p clip so we could not validate that the changes are lossless in this context.

hassount:
@kylophone any issues with this ?

PR #524, opened 2019-08-07
[WIP]Remove function tables and replace implementation choice by dsp_rtcd

JackZhouVCD:
This helps simplify the codebase by removing all function tables that are used to pick implementations (c vs simd mostly) and rewriting them in aom_dsp_rtcd.h

PR #560, opened 2019-08-20
Replace C/SIMD switch function tables with dsp_rtcd aliases

JackZhouVCD:
Reincarnate of #524.

This helps simplify the codebase by removing all function tables that are used to pick implementations (c vs simd mostly) and rewriting them in aom_dsp_rtcd.h. Also removes unused asm_type parameters in functions.

This is only updated for commit on master 6012f53c04dba7e53d6ea6728747e7be77c4741d. Please @ me or let me know when this PR goes under review so I can work on making it updated


Source/Lib/Common/ASM_AVX2/EbComputeSAD_Intrinsic_AVX2.c
```diff
@@ -3373,7 +3373,7 @@ uint32_t compute8x_m_sad_avx2_intrin(
     return (uint32_t)_mm_cvtsi128_si32(xmm0);
 }
 
-uint32_t Compute4xMSadSub_AVX2_INTRIN(
+uint32_t Compute4xMSadSub_avx2_intrin(
```
andrey-norkin: A mixture of snake_case and CamelCase. Should be snake_case.

Source/Lib/Common/Codec/EbAvcStyleMcp.c
```diff
@@ -1048,3 +1051,69 @@ void bi_pred_i_free_ref8_bit(
...
+    uint32_t dst_stride,
+    uint32_t pu_width,
+    uint32_t pu_height,
+    EbByte temp_buf,
+    EbBool skip,
+    uint32_t frac_pos,
+    uint8_t choice
+) {
+    switch (choice) {
+    case 0:
```
andrey-norkin: Does it make sense to have enum here? Numbers make it more difficult to read the code. In addition to that the input parameters choice might benefit from a more descriptive name.

Source/Lib/Common/Codec/EbMotionEstimation.c
```diff
@@ -15358,3 +15239,181 @@ EbErrorType open_loop_intra_search_sb(
...
+    const uint8_t  *ref,
+    uint32_t  ref_stride,
+    uint32_t  height,
+    uint32_t  width,
+    uint8_t choice
+) {
+    uint32_t nxm_sad = 0;
+
+    switch (choice) {
+    case 0:
```
andrey-norkin: Choice may be an enum. Also, you are calling the same functions in all the cases. You can have just one call of this functions with multiple cases, this would look cleaner.

switch (choice) {
    case 0:
    case 1:
    case 3:
...
    case 16:
        nxm_sad = fast_loop_nx_m_sad_kernel(src, src_stride, ref, ref_stride, height, width); break;
    default:
        assert(0);
    }

Source/Lib/Common/Codec/EbMotionEstimation.c
```diff
@@ -15358,3 +15239,181 @@ EbErrorType open_loop_intra_search_sb(
...
+    uint32_t  ref_stride,
+    uint32_t  height,
+    uint32_t  width,
+    uint8_t choice
+){
+    uint32_t nxm_sad = 0;
+
+    switch (choice) {
+    case 0:
+        nxm_sad = Compute4xMSadSub_avx2_intrin(src, src_stride, ref, ref_stride, height, width); break;
```
andrey-norkin: You can also group corresponding cases with the same function as in the comment above, this would look cleaner. Just remember to add a break after the function is called.

Source/Lib/Common/Codec/EbMotionEstimation.c
```diff
@@ -15358,3 +15239,181 @@ EbErrorType open_loop_intra_search_sb(
...
+    case 3:
+        nxm_sad = compute24x_m_sad_avx2_intrin(src, src_stride, ref, ref_stride, height, width); break;
+    case 4:
+        nxm_sad = compute32x_m_sad_avx2_intrin(src, src_stride, ref, ref_stride, height, width); break;
+    case 6:
+        nxm_sad = compute48x_m_sad_avx2_intrin(src, src_stride, ref, ref_stride, height, width); break;
+    case 8:
+        nxm_sad = compute64x_m_sad_avx2_intrin(src, src_stride, ref, ref_stride, height, width); break;
+    case 16:
+        nxm_sad = compute64x_m_sad_avx2_intrin(src, src_stride, ref, ref_stride, height, width); break;
```
andrey-norkin: The last two cases can be grouped.

Source/Lib/Common/Codec/EbMotionEstimation.c
```diff
@@ -15358,3 +15239,181 @@ EbErrorType open_loop_intra_search_sb(
...
+    uint32_t  ref2_stride,
+    uint32_t  height,
+    uint32_t  width,
+    uint8_t   choice){
+
+    uint32_t nxm_sad_avg = 0;
+
+    switch (choice) {
+    case 0:
+        nxm_sad_avg = combined_averaging_sad(src, src_stride, ref1, ref1_stride, ref2, ref2_stride, height, width); break;
```
andrey-norkin: All of these case statements can be grouped. Enum might be a good choice as well

Source/Lib/Common/Codec/EbPictureOperators.c
```diff
@@ -964,3 +958,123 @@ void eb_aom_yv12_copy_v_c(const Yv12BufferConfig *src_bc,
...
+
+void picture_addition_kernel_helper(uint8_t  *pred_ptr,
+    uint32_t  pred_stride,
+    int16_t *residual_ptr,
+    uint32_t  residual_stride,
+    uint8_t  *recon_ptr,
+    uint32_t  recon_stride,
+    uint32_t  width,
+    uint32_t  height,
+    uint8_t  choice){
```
andrey-norkin: If not enum then maybe a more meaningful name rather than choice would be appropriate

test/SadTest.cc
```diff
@@ -63,230 +63,230 @@ extern "C" void ext_eight_sad_calculation_32x32_64x64_c(
...
-    uint8_t *ref1_aligned_;
-    uint8_t *ref2_aligned_;
-    uint32_t sad8x8[64][8];
-    uint32_t sad16x16[16][8];
-    uint32_t sad32x32[4][8];
-};
+//namespace {
+//#define MAX_BLOCK_SIZE (128 * 256)
+//typedef std::tuple<int, int> SadSize;
+//typedef enum { REF_MAX, SRC_MAX, RANDOM, UNALIGN } TestPattern;
```
andrey-norkin: Do we need this commented code?

test/SadTest.cc
```diff
@@ -309,1149 +309,1149 @@ class SADTestBase : public ::testing::Test {
...
-INSTANTIATE_TEST_CASE_P(
-    EXTSAD, ExtSadCalculationTest,
-    ::testing::Combine(::testing::ValuesIn(TEST_PATTERNS),
-                       ::testing::ValuesIn(TEST_SAD_PATTERNS)));
-}  // namespace
+//class SADTestSubSample : public ::testing::WithParamInterface<TestSadParam>,
+//                         public SADTestBase {
+//  public:
+//    SADTestSubSample()
+//        : SADTestBase(std::get<0>(TEST_GET_PARAM(1)),
```
andrey-norkin: Could we remove the commented code?

test/compute_mean_test.cc
```diff
@@ -1,195 +1,195 @@
...
-        }
-    }
-}
-
-}  // namespace
+// /*
+//  * Copyright(c) 2019 Netflix, Inc.
+//  * SPDX - License - Identifier: BSD - 2 - Clause - Patent
+//  */
+
```
andrey-norkin: Same here

andrey-norkin:
The conflicts need to be resolved.

test/SadTest.cc
JackZhouVCD: since the functional definitions are moved, the unit tests couldn't find them anymore. I commented it out temporarily so the build passes. In the end there needs to be changes made to the unit tests to accommodate these changes

andrey-norkin: Should not fixing the corresponding unit tests be part of the PR? 
The person who makes the changes to the function definitions is likely to have the best knowledge about how to fix the Unit Tests which use those functions.

JackZhouVCD: okay, I will look into this and make all the requested changes you have listed. Thanks for reviewing my pr

PR #565, opened 2019-08-23
Remove limitation of less than 16 Sockets

### LINKED ISSUE #517: Encoder initialization fails with more than 16 sockets
The issue comes from the following line:

https://github.com/OpenVisualCloud/SVT-AV1/blob/4a12e7a45ed5471a57608d29213714aa9b186491/Source/Lib/Encoder/Codec/EbEncHandle.c#L240

On my system, `/proc/cpuinfo` returns physical ids up to 23, making it impossible to initialize the encoder as `InitThreadManagmentParams` will always return `EB_ErrorInsufficientResources`.

While 16 sockets sounds like more than enough for a physical machine, I am working on a VM which reports each core as a separate physical id.

luigi311:
closes #517 

There was a socket_id > 15 clause that would cause it to throw an error on any machine with 16 or more sockets as soon as it was about to start the encoding process. While this would never occur with physical hardware it is possible to occur with virtual machines. Virtual machines can configure their systems to do 1 socket per core or really anything.  Here is the output of my validation. 

```
luis@clearlinux~/Desktop/SVT-AV1/Bin/Release $ sudo ./SvtAv1EncApp -i ../../output.y4m -enc-mode 8 -b ../../multisocket.mkv
SVT-AV1 Encoder
SVT [version]:	SVT-AV1 Encoder Lib v0.6.0
SVT [build]  :	GCC 9.2.1 20190821 gcc-9-branch@274807	 64 bit
LIB Build date: Aug 22 2019 19:21:54
Number of logical cores available: 32
Number of PPCS 88
SVT [config]: Main Profile	Tier (auto)	Level (auto)	
SVT [config]: EncoderMode 							: 8 
SVT [config]: EncoderBitDepth / EncoderColorFormat / CompressedTenBitFormat	: 8 / 1 / 0
SVT [config]: SourceWidth / SourceHeight					: 1920 / 1080 
SVT [config]: Fps_Numerator / Fps_Denominator / Gop Size / IntraRefreshType 	: 24000 / 1001 / 16 / 1
SVT [config]: HierarchicalLevels / BaseLayerSwitchMode / PredStructure		: 4 / 0 / 2 
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange			: CQP / 50 / 33 / 0 
Encoding      1000
SUMMARY --------------------------------- Channel 1  --------------------------------
Total Frames		Frame Rate		Byte Count		Bitrate
        1000		23.98 fps		   6888015		1321.18 kbps


Channel 1
Average Speed:		9.464 fps
Total Encoding Time:	105665 ms
Total Execution Time:	106855 ms
Average Latency:	9337 ms
Max Latency:		14911 ms
Encoder finished

luis@clearlinux~/Desktop/SVT-AV1 $ lscpu
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
Address sizes:       40 bits physical, 48 bits virtual
CPU(s):              32
On-line CPU(s) list: 0-31
Thread(s) per core:  1
Core(s) per socket:  1
Socket(s):           32
NUMA node(s):        1
Vendor ID:           AuthenticAMD
CPU family:          23
Model:               1
Model name:          AMD Ryzen 7 1700 Eight-Core Processor
Stepping:            1
CPU MHz:             2993.902
BogoMIPS:            5987.80
Virtualization:      AMD-V
L1d cache:           64K
L1i cache:           64K
L2 cache:            512K
L3 cache:            16384K
NUMA node0 CPU(s):   0-31
Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm rep_good nopl cpuid extd_apicid pni pclmulqdq ssse3 fma cx16 sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm cmp_legacy svm cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw perfctr_core ssbd ibpb vmmcall fsgsbase tsc_adjust bmi1 avx2 smep bmi2 rdseed adx smap clflushopt sha_ni xsaveopt xsavec xgetbv1 virt_ssbd arat npt nrip_save
```

Signed-off-by: Luis Garcia <luigi311.lg@gmail.com>

### DIFF
--- a/Source/Lib/Encoder/Codec/EbEncHandle.c
+++ b/Source/Lib/Encoder/Codec/EbEncHandle.c
@@ -1923,7 +1923,6 @@ EB_API EbErrorType eb_deinit_handle(
     #if  defined(__linux__)
         if(lp_group != NULL) {
             EB_FREE(lp_group);
-            lp_group = NULL;
         }
     #endif
     return return_error;
--- a/Source/Lib/Encoder/Codec/EbEncHandle.c
+++ b/Source/Lib/Encoder/Codec/EbEncHandle.c
@@ -247,8 +247,8 @@ EbErrorType InitThreadManagmentParams() {
                 if (socket_id >= maxSize) {
                     maxSize = maxSize * 2;
                     lp_group = realloc(lp_group, maxSize * sizeof(processorGroup));
-                    if (lp_group == (processorGroup*) NULL) 
-                        return EB_ErrorInsufficientResources; 
+                    if (lp_group == (processorGroup*) NULL)
+                        return EB_ErrorInsufficientResources;
                 }
                 lp_group[socket_id].group[lp_group[socket_id].num++] = processor_id;
             }
--- a/Source/Lib/Encoder/Codec/EbEncHandle.c
+++ b/Source/Lib/Encoder/Codec/EbEncHandle.c
@@ -110,8 +110,8 @@ typedef struct logicalProcessorGroup {
     uint32_t num;
     uint32_t group[1024];
 }processorGroup;
-#define MAX_PROCESSOR_GROUP 16
-processorGroup                   lp_group[MAX_PROCESSOR_GROUP];
+#define INITIAL_PROCESSOR_GROUP 16
+processorGroup                  *lp_group = NULL;
 #endif
 
 /**************************************
@@ -217,11 +217,12 @@ EbErrorType InitThreadManagmentParams() {
     const char* PHYSICALID = "physical id";
     int processor_id_len = EB_STRLEN(PROCESSORID, 128);
     int physical_id_len = EB_STRLEN(PHYSICALID, 128);
+    int maxSize = INITIAL_PROCESSOR_GROUP;
     if (processor_id_len < 0 || processor_id_len >= 128)
         return EB_ErrorInsufficientResources;
     if (physical_id_len < 0 || physical_id_len >= 128)
         return EB_ErrorInsufficientResources;
-    memset(lp_group, 0, sizeof(lp_group));
+    memset(lp_group, 0, INITIAL_PROCESSOR_GROUP * sizeof(processorGroup));
 
     FILE *fin = fopen("/proc/cpuinfo", "r");
     if (fin) {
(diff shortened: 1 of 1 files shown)


luigi311:
Need to add IF statements around free and malloc to prevent issues when using multi-channel

kylophone:
> Need to add IF statements around free and malloc to prevent issues when using multi-channel

Can you just use a static array instead of a dynamic vector? Is there an upper limit to the possible number of sockets?

luigi311:
> Can you just use a static array instead of a dynamic vector? Is there an upper limit to the possible number of sockets?

Just to avoid any hardcoded limits. There is no limit on the amount of sockets so in order to avoid this issue in the future i would need to set the size to a crazy amount which is a waste of memory when i can just have it dynamically grow and only have it use a bit to much memory depending on where the socket count falls in powers of 2.

Source/Lib/Encoder/Codec/EbEncHandle.c
```diff
@@ -1845,6 +1852,14 @@ EB_API EbErrorType eb_init_handle(
     if(p_handle == NULL)
          return EB_ErrorBadParameter;
 
+    #if defined(__linux__)
```
1480c1: Would/should any of these code also apply to macOS?

luigi311: I see no reason why all of this would not work on macOS as well but i do not have a mac environment that i can test on. Everything revolving around lp_group is already limited to linux so i am not sure if that was just decided prior to macOS support or if there was a reason to limit it to linux.

1480c1: Nevermind, didn't unfold the code before and didn't notice the ifdef __linux__

kylophone:
Should be OK. Allocations are supposed to be tracked with the `EbMalloc` macro. I say "supposed to be" because I've seen in a few instances where straight `malloc` like you've done too. Could this be re-written to use the `Eb` allocation macros instead of `malloc` and `realloc`?

kylophone:
Also, another design question. If these are allocated at init time and never grow again, why does it have to be a vector that you grow with `realloc`? You should already know the maximum size at init, just allocate that amount and then be done with it.

luigi311:
Yeah i have never seen eb_malloc before so i used malloc. I can rewrite it to use eb_malloc instead since it looks like it is allocated pretty much the same way. 

>If these are allocated at init time and never grow again, why does it have to be a vector that you grow with realloc? You should already know the maximum size at init, just allocate that amount and then be done with it.

Isnt that false when utilizing multiple instances? I have never programmed with multiple instances in mind so i am not sure but in svt-hevc they made it seem like the different instances still utilized the same variables so without the if conditions on the malloc and free it would crash when they tried to encode multiple channels.
https://github.com/OpenVisualCloud/SVT-HEVC/pull/345

luigi311:
Is there no eb_realloc? I do not see any references to it in EbMalloc.h

luigi311:
All set on my end

kylophone:
Merged, thanks! Sorry about the wait.

PR #585, opened 2019-08-30
Disable dual interpolation filter

### LINKED ISSUE #584: Disable dual interpolation filter
Dual interpolation filter OFF provides better BD-Rate/Speed than disable dual interpolation filter ON

hguermaz:
## Description

Disable dual interpolation filter.

Closes #584

## Author

@hguermaz 

## Type of change

New feature/Tool

## Tests and performance
* BD-Rate to aom cpu0-2pass on the 28 360p clips:  0.504% gain
* Speed on a 4-core machine:  ~2% increase
* Memory footprint on a 4-core machine: negligeable


joelsole:
Interpolation filters might have more performance impact at higher resolutions. Do you have any numbers on the impact for high resolutions of turning these filters off?

hguermaz:
> Interpolation filters might have more performance impact at higher resolutions. Do you have any numbers on the impact for high resolutions of turning these filters off?

~0.2% gain when tested in context of rtime-m0-test. We also checked in context of aom and we did not get a gain.

joelsole:
So in this case, it would make sense to enable/disable depending on the input resolution. However, it might not be desirable to start having such type of tuning. 
Is there in the encoder any other tool that is enabled depending on the input format (resolution, bit-depth)?

hguermaz:
> So in this case, it would make sense to enable/disable depending on the input resolution. However, it might not be desirable to start having such type of tuning.
> Is there in the encoder any other tool that is enabled depending on the input format (resolution, bit-depth)?

I think the gain of shutting dual on the full list is expected to drop to ~0.2% because of many ~0% and not because of losses. We will re-run the data in context of master when convenient. I'm aware of settings depending on resolution (e.g. HME/ME, # of canidates @ MD, ..), but we might enable 1/8th for only 360p (WIP).  ATB is not supported for a 10BIT input.

PR #601, opened 2019-09-11
Add unit tests for picture addition functions

Cidana-Developers:
1. add unit tests for picture_addition_kernel{m}x{n}_{sse,sse2}_intrin,
  picture_addition_kernel{m}x{n}_av1_sse2_intrin,
  picture_addition_kernel16bit_sse2_intrin test

### DIFF
--- a/test/PictureAdditionTest.cc
+++ b/test/PictureAdditionTest.cc
@@ -1,0 +1,338 @@
+/*
+ * Copyright(c) 2019 Netflix, Inc.
+ * SPDX - License - Identifier: BSD - 2 - Clause - Patent
+ */
+
+/******************************************************************************
+ * @file PictureAdditionTest.cc
+ *
+ * @brief Unit test for Residual functions:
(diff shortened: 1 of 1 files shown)


test/PictureAdditionTest.cc
```diff
@@ -0,0 +1,338 @@
...
+
+typedef struct PictureAdditionTestParam {
+    AreaSize area_size;
+    LbdPictureAdditionFunc lbd_test_func;
+    HbdPictureAdditionFunc hbd_test_func;
+} PictureAdditionTestParam;
+typedef std::tuple<PictureAdditionTestParam, TestPattern> TestParam;
+PictureAdditionTestParam TEST_PARAMS[] = {
+    {AreaSize(4, 4),
+     &picture_addition_kernel4x4_sse_intrin,
```
ttrigui: picture_addition_kernel4x4_sse_intrin and picture_addition_kernel4x4_av1_sse2_intrin have different parameters. For one residual is 16bits and for the other it is 32bits. I think we need a c version to test it against.

PR #617, opened 2019-09-17
Change the MRP prediction references for non-reference frames

anaghdin:
## Description
Change the MRP prediction references:
In 4 Layer prediction structure: Pictures 3, 5 , 7 and 9 use 1 as the reference
In 5 Layer prediction structure: Pictures 3, 5 , 7 and 9 use 1 as the reference, 11, 13, 15 and 17 use 9 as the reference

## Authors
@anaghdin

## Type of change
Feature enhancement

## Tests and performance
Expected Average PSNR-SSIM BD-rate gain in the range of -0.3% on the AOM test set, using 4 QP values: {20, 32, 43 and 55}.

## Speed impact:
Speed drop of ~5% for M0

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -40,6 +40,8 @@ extern "C" {
 #endif
 
 #define II_COMP_FLAG 1
+#define PRED_CHANGE                  1 // Change the MRP in 4L Pictures 3, 5 , 7 and 9 use 1 as the reference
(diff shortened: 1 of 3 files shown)


joelsole:
lgtm. 
What would be the reason for the speed drop?

anaghdin:
> lgtm.
> What would be the reason for the speed drop?
The speed drop is due to the new dependency on picture 1 and 9. i.e. Pictures 3, 5, 7 and 9 have to wait for picture 1 to be completed.

EwoutH:
@hassount Could this PR get an _Q+_ label, for historical reference?

PR #620, opened 2019-09-19
Support arbitrary picture size (non multiple of 8)

kylophone:
Hi,

The following PR removes the restriction for non multiple of 8 sized pictures. Please feel free to test.

After looking at this section of code, I would say that the current buffer allocation model should be reworked at some point. The onus of allocating, padding, aligning the input buffer happens in the app. Not ideal, this should really be something that is handled by the lib.

Thanks,
Kyle

### DIFF
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -99,6 +99,9 @@ typedef struct EbSvtAv1EncConfiguration
      *
      * Default is 0. */
     uint32_t                 source_height;
+
+    uint32_t render_width, render_height;
+
     /* The frequecy of images being displayed. If the number is less than 1000,
      * the input frame rate is an integer number between 1 and 60, else the input
      * number is in Q16 format, shifted by 16 bits, where max allowed is 240 fps.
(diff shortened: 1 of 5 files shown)


ultrawide:
Hi, here are my results: 

No loss when compared against base encoder (48c155aa0756f074622c3572d93f331a5b72eea4).

Below are the speed results using QP 32 and encoding 500 frames from a 360p video
|  | Base (Reference) | Release |
| :---         |     :---:      |          ---: |
| M1 | 0.427 | 0.432 |
| M4 | 4.337 | 4.267 |
| M8 | 35.967 | 36.106 |

hassount:
@kylophone can you please add the issue that this addresses ?

ultrawide:
> No loss when compared against base encoder ([48c155a](https://github.com/OpenVisualCloud/SVT-AV1/commit/48c155aa0756f074622c3572d93f331a5b72eea4)). Using QP 32 and encoding 500 frames from a 360p video, I found a speed deviation of 0.0188% (improvement over base) averaged over m{1,4,8}.

No memory deviations found.

joelsole:
Closes #645

PR #615, opened 2019-09-17
fix leaks for encmode 0

xuguangxin:
fixes https://github.com/OpenVisualCloud/SVT-AV1/issues/600

PR #558, opened 2019-08-20
MinGW: Enable MinGW support

1480c1:
Since #291 hasn't been updated. Closes #291

Issue #538, opened 2019-08-13
Invalid bitstream generated when IntraPeriod is set


Cidana-Developers:
commit id: d0f64ffa3230ace50c70b7b42c9b25642dd8a944
build type: release
platform: ubuntu 16.04
Steps:
1. encode the bitstream with SvtAv1EncApp:
```
./SvtAv1EncApp -i ../../test/vectors/kirland_640_480_30.yuv -w 640 -h 480 -intra-period 10 -o /test/recon.yuv -stat-report 1 -b /test/output.ivf
-------------------------------------------
SVT-AV1 Encoder
SVT [version]:  SVT-AV1 Encoder Lib v0.6.0
SVT [build]  :  GCC 7.3.0        64 bit
LIB Build date: Aug 13 2019 07:35:07
-------------------------------------------
Number of logical cores available: 16
Number of PPCS 88
-------------------------------------------
SVT [config]: Main Profile      Tier (auto)     Level (auto)
SVT [config]: EncoderMode                                                       : 8
SVT [config]: EncoderBitDepth / EncoderColorFormat / CompressedTenBitFormat     : 8 / 1 / 0
SVT [config]: SourceWidth / SourceHeight                                        : 640 / 480
SVT [config]: FrameRate / Gop Size                                              : 30 / 11
SVT [config]: HierarchicalLevels / BaseLayerSwitchMode / PredStructure          : 4 / 0 / 2
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange                  : CQP / 50 / 33 / 0
-------------------------------------------
Encoding       300
SUMMARY --------------------------------- Channel 1  --------------------------------
Total Frames            Frame Rate              Byte Count              Bitrate
         300            30.00 fps                   303194              242.56 kbps

Average QP              Y-PSNR                  U-PSNR                  V-PSNR
      41.73             44.28 dB                   47.43dB              47.84dB


Channel 1
Average Speed:          81.433 fps
Total Encoding Time:    3684 ms
Total Execution Time:   4113 ms
Average Latency:        1136 ms
Max Latency:            1456 ms
Encoder finished
```
2. try to decode with dav1d:
```
./dav1d -i ../../../test/output.ivf -o decoded.yuv
dav1d 0.4.0-12-gdff0a08 - by VideoLAN
Decoded 11/273 frames (4.0%)Error decoding frame: Invalid argument
```

Cidana-Developers:
[...]

Issue #343, opened 2019-06-06
Add Constrained Quality (CQ) rate-control mode


EwoutH:
A true Constrained Quality (CQ) rate-control mode would be a very useful feature. Currently the constrained variable bitrate allows some sort of constrained mode, but not with a maximum bitrate of constant quality. It's very useful for situations where you have limited decoding or bandwidth capabilities.

The Constrained Quality mode should function identical to the Constant Quality mode, but with an upper bitrate limit. So `-q 25 -max-br 500000` for example. See also the [FFmpeg libaom AV1 Encoding Guide](https://trac.ffmpeg.org/wiki/Encode/AV1) as example.

Also, I don't think an upper quality `-max-qp` it that useful. A upper bitrate (`-max-br`) would be better (but both can be kept).

gnafuthegreat:
I think this is a good idea.  Presumably hardware decoders will only support up to a certain level, and this will probably tie into those restrictions.  It would be nice to have constant quality but still target a certain level in the long run, and at least target a max bitrate in the short term.

PR #623, opened 2019-09-20
Prep 0.7.0 tag

hassount:
Update CHANGELOG.md and version number for 0.7.0 tag


EwoutH:
Great progress, thanks for taking the feedback in #29 in account!

Could we merge some build issues before release? That will make integration with other applications significantly easier. #558 is very important for Handbrake (see https://github.com/HandBrake/HandBrake/issues/2313) and #254 and #554 should make the build process more stable. 

Also merging #620 before release will prevent a lot of bug reports.

Furthermore, I think the bugfixes #615 and #618 should be merged before a tag, and #538 should be looked into.

I would still suggest creating a [Milestone](https://github.com/OpenVisualCloud/SVT-AV1/milestones) for 0.7.0 so that the appropriate PRs and issues can be tracked. It will help greatly in the effort for #238.

1480c1:
Make sure to update the cmakelists in the decoder and encoder folders

1480c1:
Alternatively, we can make it so that either the API header is updated from cmake or cmake reads from the header

EwoutH:
Would also be super useful to get an Constrained Quality (CQ) rate-control mode in (see #343).

hassount:
@EwoutH 
Thank you for the feedback, so we have:

- [ ] #558 
- [ ] #554 
- [ ] #254 
- [ ] #620 
- [ ] #615
- [ ] #618  

#538 will probably not be fixed by 0.7. Adding milestones, will be working on moving away from trello and adding milestones.

@1480c1 
my understanding is that the SVT_VERSION_MAJOR, SVT_VERSION_MINOR, SVT_VERSION_PATCHLEVEL are shared between the encoder and the decoder and so both libraries will be updated accordingly. Can you explain further what you are suggesting to be done ? or if you have some time, suggest the change so that we are on the same page.

1480c1:
What I meant was
``` diff
diff --git a/Source/Lib/Decoder/CMakeLists.txt b/Source/Lib/Decoder/CMakeLists.txt
index edbd830d..f94b5ec4 100644
--- a/Source/Lib/Decoder/CMakeLists.txt
+++ b/Source/Lib/Decoder/CMakeLists.txt
@@ -6,7 +6,7 @@
 
 # Shared Decoder Version
 set(DEC_VERSION_MAJOR 0)
-set(DEC_VERSION_MINOR 6)
+set(DEC_VERSION_MINOR 7)
 set(DEC_VERSION_PATCH 0)
 set(DEC_VERSION ${DEC_VERSION_MAJOR}.${DEC_VERSION_MINOR}.${DEC_VERSION_PATCH})
 
diff --git a/Source/Lib/Encoder/CMakeLists.txt b/Source/Lib/Encoder/CMakeLists.txt
index 2dc14385..6a5b58ba 100644
--- a/Source/Lib/Encoder/CMakeLists.txt
+++ b/Source/Lib/Encoder/CMakeLists.txt
@@ -6,7 +6,7 @@
 
 # Shared Encoder Version
 set(ENC_VERSION_MAJOR 0)
-set(ENC_VERSION_MINOR 6)
+set(ENC_VERSION_MINOR 7)
 set(ENC_VERSION_PATCH 0)
 set(ENC_VERSION ${ENC_VERSION_MAJOR}.${ENC_VERSION_MINOR}.${ENC_VERSION_PATCH})
 ```

to update the cmakelists and so that way the pkg-config file will be properly updated

1480c1:
Alternatively, I can try to see if I can make CMake to pull the version number from EbSvtAv1.h and set the version that way, or do in reverse and have CMake generate the EbSvtAv1.h with the version number set in CMakeLists

hassount:
> Alternatively, I can try to see if I can make CMake to pull the version number from EbSvtAv1.h and set the version that way, or do in reverse and have CMake generate the EbSvtAv1.h with the version number set in CMakeLists

Thanks, that would be ideal, that said, if it takes too much time, we can change it for now and add the functionality later, what do you suggest ?

1480c1:
I would say go ahead and make the change, as it may take some time before I can fully make sure which way is better

EwoutH:
@hassount List looks good, the SVT-AV1 build system is shaping up nicely! Would be nice to get #162, #634 and #635 also in.

EwoutH:
@hassount Thanks for updating the list. GitHub has a very nice feature built in, called [Milestones](https://github.com/OpenVisualCloud/SVT-AV1/milestones), where you can assign issues and PRs you want to get fixed/implemented before a specific release. For a better overview, you could create a milestone for 0.7.0.

joelsole:
> @joelsole
> All ready to merge, can you do a quick review ?

lgtm

PR #624, opened 2019-09-20
Fix stride issue in c pack avx2 intrin

Cidana-Developers:
In c_pack_avx2_intrin, the output stride is not consistent when the output data is stored. When the input width is 32, the output stride is fixed to 8; If the input width is 64, the output stride could be out_stride or 16.
It would be good if we can use the out_stride consistently.

### DIFF
--- a/Source/Lib/Common/ASM_AVX2/EbPictureOperators_Intrinsic_AVX2.c
+++ b/Source/Lib/Common/ASM_AVX2/EbPictureOperators_Intrinsic_AVX2.c
@@ -243,8 +243,10 @@ void c_pack_avx2_intrin(
                 {
                     __m256i c0 = _mm256_loadu_si256((__m256i*)(localPtr));
                     __m256i c1 = _mm256_loadu_si256((__m256i*)(localPtr + 32));
-                    _mm256_stream_si256((__m256i*)&in_compn_bit_buffer[0], c0);
-                    _mm256_stream_si256((__m256i*)&in_compn_bit_buffer[32], c1);
+                    _mm_storeu_si128((__m128i*)  (in_compn_bit_buffer), _mm256_extractf128_si256(c0, 0));
+                    _mm_storeu_si128((__m128i*)  (in_compn_bit_buffer + out_stride), _mm256_extractf128_si256(c0, 1));
+                    _mm_storeu_si128((__m128i*)  (in_compn_bit_buffer + 2 * out_stride), _mm256_extractf128_si256(c1, 0));
+                    _mm_storeu_si128((__m128i*)  (in_compn_bit_buffer + 3 * out_stride), _mm256_extractf128_si256(c1, 1));
                     in_compn_bit_buffer += 4 * out_stride;
                 }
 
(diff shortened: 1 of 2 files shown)


ttrigui:
kernel c_pack_avx2_intrin is currently not being used. I tried to use the unit test in order to test the fix. I used the following --gtest_filter="*PackTest*" and many tests were running. Is there a way to only test c_pack_avx2_intrin ? also would you please be bale to display the function name that failed and what input ( search area, block size,...) triggered the fail along with the expected output value and the actual one ?

Source/Lib/Common/ASM_AVX2/EbPictureOperators_Intrinsic_AVX2.c
```diff
@@ -245,8 +245,10 @@ void c_pack_avx2_intrin(
                 {
                     __m256i c0 = _mm256_loadu_si256((__m256i*)(localPtr));
                     __m256i c1 = _mm256_loadu_si256((__m256i*)(localPtr + 32));
-                    _mm256_stream_si256((__m256i*)&in_compn_bit_buffer[0], c0);
-                    _mm256_stream_si256((__m256i*)&in_compn_bit_buffer[32], c1);
+                    _mm_storeu_si128((__m128i*)  (in_compn_bit_buffer), _mm256_extractf128_si256(c0, 0));
```
ttrigui: in_compn_bit_buffer holds values contiguous in memory. the assumption of using 16 as a stride holds. The suggested solution is more generic but more costly. I suggest we check on the stride if it is 16 we use the old code otherwise we use the new one.

Cidana-Developers:
> kernel c_pack_avx2_intrin is currently not being used. I tried to use the unit test in order to test the fix. I used the following --gtest_filter="_PackTest_" and many tests were running. Is there a way to only test c_pack_avx2_intrin ? also would you please be bale to display the function name that failed and what input ( search area, block size,...) triggered the fail along with the expected output value and the actual one ?

Done. Please use --gtest_filter="\*/PackTest\*" to test c_pack_avx2_intrin only.

Source/Lib/Common/ASM_AVX2/EbPictureOperators_Intrinsic_AVX2.c
Cidana-Developers: Since this is an implicit assumption in the implementation, we may add an assertion to check if the out_stride is width>>2 at the beginning of the function.

hassount:
@ttrigui have all your items been addressed for this PR ? 
@Cidana-Developers can you please add a clear description of what exactly was addressed here ?

Cidana-Developers:
> @ttrigui have all your items been addressed for this PR ?
> @Cidana-Developers can you please add a clear description of what exactly was addressed here ?

Done, sorry about that.

andrey-norkin:
@ttrigui, Does the PR look fine now?

ttrigui:
> @ttrigui, Does the PR look fine now?

yes it is fine .

PR #594, opened 2019-09-04
Speed optimization

hguermaz:
## Description
Added the following lossy speed optimizations:
- Opt 0: Use the top 4 ME candidates only @ 3x3 Unipred and 3x3 Bipred.
- Opt 1: Skip ATB if parent block does not have coeff.
- Opt 2: Use edge detection to bypass some angular modes.
- Opt 3: Bipred candidates reduction @ ME.
- Opt 4: MD candidates reduction @ MD.

Closes #593  

## Author

@NaderMahdi 
@chkngit  
@hguermaz 

## Type of change

New feature/Tool

## Tests and performance
* BD-Rate to aom cpu0-2pass on the full list: Opt 0: 0.020% loss, Opt 1: 0.023%, Opt 2: 0.029%, Opt 3: M1 & higher tool , Opt 4: M1 & higher too.l 
* Speed on a 4-core machine:  Opt 0: 30% faster, Opt 1: 0.9% faster, Opt 2: 1.5% faster, Opt 3: M1 & higher tool , Opt 4: M1 & higher too.l 
* Memory footprint on a 4-core machine: negligible

PR #626, opened 2019-09-20
Fix mismatch issue in PackUnpack unit test

Cidana-Developers:
1. limit the data range in avg and sub_avg test, only 10-bit are
   populated in a short.
2. do not test area_width of 4 since 4 is not supported in
   unpack_avg_safe_sub_avx2_intrin

### DIFF
--- a/test/PackUnPackTest.cc
+++ b/test/PackUnPackTest.cc
@@ -44,6 +44,8 @@
 #include "random.h"
 #include "util.h"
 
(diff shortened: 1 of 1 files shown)


hassount:
@Cidana-Developers  is there an issue that this fix is related to ? if so can you please add such to the description keeping a similar format to #594

Cidana-Developers:
> @Cidana-Developers is there an issue that this fix is related to ? if so can you please add such to the description keeping a similar format to #594

Will create issue and add description later.

PR #627, opened 2019-09-20
Fix bug in eb_enc_msb_un_pack2d_sse2_intrin

Cidana-Developers:
1. shift the data before pack.


hassount:
@Cidana-Developers  is there an issue that this fix is related to ? if so can you please add such to the description keeping a similar format to #594

Cidana-Developers:
> @Cidana-Developers is there an issue that this fix is related to ? if so can you please add such to the description keeping a similar format to #594

No, will create an issue first and then relate to it.

PR #638, opened 2019-09-24
Add YUV422 and YUV444 support unit tests in video source

Cidana-Developers:
1. Add YUV 422 and YUV 444 support in video source


kylophone:
Maybe this should be renamed slightly, since it doesn't add support for these pixel formats in the encoder.

PR #639, opened 2019-09-24
Add high bit-depth variance unit test

Cidana-Developers:
Added unit test and test cases for high bit-depth variance
calculation.

### DIFF
--- a/test/HbdVarianceTest.cc
+++ b/test/HbdVarianceTest.cc
@@ -1,0 +1,464 @@
+/*
+ * Copyright(c) 2019 Netflix, Inc.
+ * SPDX - License - Identifier: BSD - 2 -
+ * Clause - Patent
+ */
+
+/******************************************************************************
+ * @file HbdVarianceTest.cc
+ *
+ * @brief Unit test for HBD variance
+ * functions:
+ * - eb_aom_highbd_BD{8,10,12}_varianceW{8,16,32,64}xH{4,8,16,32,64}_sse2
+ * - eb_aom_highbd_BD{8,10,12}_getS{8,16}xS{8,16}var_sse2
+ *
+ * @author  Cidana-Wenyao, Cidana-Edmond
+ *
+ ******************************************************************************/
+#include <math.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <new>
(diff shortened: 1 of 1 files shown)


EwoutH:
@Cidana-Developers You introduced a new warning on the Visual Studio builds in AppVeyor. Could you adress it? https://ci.appveyor.com/project/OpenVisualCloud/svt-av1/builds/27647387/messages

Cidana-Developers:
Seems like it's not introduced by this pr. 

All the warnings are about redefinition of 'NON_AVX512_SUPPORT'. This is caused by the duplicate definition in CMakelists.txt and EbDefinitions.h, and the definitions are added in commit 6a897bd32534f62873bb85544d8b4b6f016c79fd and commit 
75257310ef302478d05bd2492131d4e658222095.

In 6a897bd32534f62873bb85544d8b4b6f016c79fd, cmake will check the available of avx512 and add definition of NON_AVX512_SUPPORT in the CMakelists.txt. 
In 75257310ef302478d05bd2492131d4e658222095, NON_AVX512_SUPPORT is always defined in EbDefinitions.h.

If the build machine does support NON_AVX512_SUPPORT, there is no warnings. To fix it, we may remove the definition from EbDefinitions.h. However I'm not sure if AVX512 is ready, could you confirm that?

EwoutH:
@Cidana-Developers Are you sure? The [previous build](https://ci.appveyor.com/project/OpenVisualCloud/svt-av1/builds/27645217) doesn't have the warnings, and those include the CMake changes. 

@1480c1 Any idea?

1480c1:
Currently, a lot of macros that are usually generated by a configure script are hard coded in EbDefinitions. Ideally, we would use CMake to generate all of the necessary macros (AVX512 support from the compiler, etc), but that would also take some time

Cidana-Developers:
> @Cidana-Developers Are you sure? The [previous build](https://ci.appveyor.com/project/OpenVisualCloud/svt-av1/builds/27645217) doesn't have the warnings, and those include the CMake changes.

I can not reproduce the warnings on build machine which support AVX512. I guess the build happened to be executed on a machine which does not support AVX512.

PR #640, opened 2019-09-24
Modify e2e test to increase the coverage

Cidana-Developers:
1. modify the screen coding config to cover the intrabc mode.


EwoutH:
@hassount Could this PR get a _test_ label, for historical reference

PR #670, opened 2019-10-07
 Fix AVX512 build

ttrigui:
Fix build error when enabling AVX512


mafonsoN:
The only change I see is including `synonyms.h` in `pickrst_avx512.c`? Is this correct?

ttrigui:
> The only change I see is including `synonyms.h` in `pickrst_avx512.c`? Is this correct?

yes. Correct.

PR #678, opened 2019-10-08
+ enable redundant block for M0

ttrigui:


JackZhouVCD:
As the setting for redundant block is [already present](https://github.com/OpenVisualCloud/SVT-AV1/pull/678/files#diff-c87d7ea89cc45aa201c092aea82b455eL1462) in master, this PR is not needed.

Issue #668, opened 2019-10-04
Code Cleanup


hassount:
Now that the feature implementations are almost complete, Let's use this issue to set the items to be cleaned up. If you see something, please put it here:

**Phase 1:**
- [x] remove rate control mode 1
- [x] remove use_ssd as input to md_stage_0. 
- [x] remove decoupled_fast_loop_search_method and decouple_intra_inter_fast_loop
- [x] remove lcu from naming (e.g. mode_decision_configure_lcu to mode_decision_configure_sb)
- [x] port COST_CLEAN_UP from rtime-m0-test
- [x] remove fast_candidate_intra_count
- [x] LargestCodingUnit to SuperBlock
- [x] ConstrainedIntra @chkngit :)
- [x] CodingUnit to BlkStruct / cu_ptr to blk_ptr / cu_origin_x -y to blk_origin_x-y / 
- [x] reinforce snake_casing in code
- [x] change atb to tx_size _search
- [x] remove unused functions `void eb_av1_cdef_search()` and `void av1_cdef_search16bit()` in EbCdef.c
- [x] verify all of the commandline switches are documented and has valid checks for invalid input values #250 
- [x] remove in_loop_me_flag
- [x] remove useless MR_MODE checks (where M0 is already using the same setting)
- [x] remove check_skip_sub_blks()  and skip_sub_blks signal
- [x] rename copy_rbsp_bitstream_to_payload to copy_payload
- [x] remove ConstrainedIntra (@chkngit :) )
- [x] remove the following kernels (unused code)
- aom_lpf_vertical_4_dual_sse2
- aom_lpf_vertical_8_dual_c
- aom_lpf_vertical_14_dual_sse2/aom_lpf_vertical_14_dual_c
- aom_lpf_horizontal_4_dual_sse2
- aom_lpf_horizontal_14_dual_sse2
- aom_lpf_horizontal_8_dual_sse2/aom_lpf_horizontal_8_dual_c
- aom_highbd_lpf_vertical_14_dual_sse2/aom_highbd_lpf_vertical_14_dual_avx2
- aom_highbd_lpf_vertical_8_dual_avx2/aom_highbd_lpf_vertical_8_dual_sse2
- aom_highbd_lpf_vertical_4_dual_avx2/aom_highbd_lpf_vertical_4_dual_sse2
- aom_highbd_lpf_horizontal_14_dual_sse2/aom_highbd_lpf_horizontal_14_dual_avx2
[...]

PR #681, opened 2019-10-09
encoder_user_guide.md: Fix typo for -rc value CBR

jefft0:
In the encoder user guide, the values for the `-rc` option currently say:

    0 = CQP , 1 = ABR , 2 = ABR , 3 = CVBR

This fixes the second `ABR` to `CBR` (constant bit rate).


hassount:
@jefft0 Thank you for the contribution. Actually rc 1 is deprecated and rc 2 is the ABR preset. this will be cleared up after #668 is resolved. For consistency we could remove the rc 1 from documentation for now or make it default to rc 2

jefft0:
I was trying to use constant bit rate see if there is an encoding mode where there is only one tile group after each frame header. Is there an encoding mode which does that?

hassount:
> I was trying to use constant bit rate see if there is an encoding mode where there is only one tile group after each frame header. Is there an encoding mode which does that?

@jefft0 there's no such mode currently

hassount:
Closing this for now until we cleanup the rc presets

PR #682, opened 2019-10-09
Set hbd-md true by default

joelsole:
Enabling 10-bit MD by default to get the ~6% BD-rate gains.

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -484,7 +484,7 @@ void eb_config_ctor(EbConfig *config_ptr)
     config_ptr->hme_level2_search_area_in_height_array[0]  = 1;
     config_ptr->hme_level2_search_area_in_height_array[1]  = 1;
     config_ptr->screen_content_mode                  = 2;
-    config_ptr->enable_hbd_mode_decision             = EB_FALSE;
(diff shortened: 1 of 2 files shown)


hassount:
@joelsole 
can we restrict this to M0? 
also, you might need to add this in the default settings of the library in EbEnchandle.c if you want the same behavior for ffmpeg

joelsole:
> @joelsole
> can we restrict this to M0?
> also, you might need to add this in the default settings of the library in EbEnchandle.c if you want the same behavior for ffmpeg

Thanks, Hassene. Your comments have been addressed.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -878,7 +878,7 @@ static EbErrorType VerifySettings(EbConfig *config, uint32_t channelNumber)
         fprintf(config->error_log_file, "Error instance %u: Invalid HBD mode decision flag [0 - 1], your input: %d\n", channelNumber + 1, config->target_socket);
         return_error = EB_ErrorBadParameter;
     }
-    if (config->enable_hbd_mode_decision == 1 && config->encoder_bit_depth != 10)
+    if (config->encoder_bit_depth < 10)
```
hassount: @joelsole this condition should also be in the library next to the M0 check

PR #685, opened 2019-10-09
Add eb_av1_lowbd_pixel_proj_error_avx512()

lzhangnj:
32%-49% faster than avx2.


hassount:
@lzhangnj 
Can you please rebase this so that we can merge it ?

PR #696, opened 2019-10-14
Fix a read overflow at the end of y4m headers

magwyz:
For example, after the use of strtol, tokstart is set to one character
after the last of the number. The increment part of the for-loop increases
tokstart by one. If the parsed number was at the end of the header string,
tokstart is now one character after the '\0' at the end of the string. The
loop may then continue and fail because it tries to parse uninitialized
memory.

This commit transforms the for-loop into a while-loop without increment.
It also adds an assert to check that we do not try to parse further the
end of the header string.

This fixes random premature stops of the encoder with valid y4m inputs.


JackZhouVCD:
Thanks for adding this fix! The changes are lossless with no speed impact so it looks good so far. Can you provide me with a test case for us to validate that your fix works?

magwyz:
To always trigger the parsing algorithm overflow issue, you can test the following tiny patch.
Instead of leaving the buffer data initialized, the added `memset` just initialize it with specific data so the parsing will always fail, even if your input file is perfectly fine.

```From 99c03107cbd50884f091ca62aee2d460ad444731 Mon Sep 17 00:00:00 2001
From: Adrien Maglo <adrien@videolabs.io>
Date: Tue, 15 Oct 2019 22:38:03 +0200
Subject: [PATCH] Init Y4M header memory buffer with predefined data so parsing
 will always fail

The header is read with fgets but since the parsing algorithm overflows,
it ends up trying to parse data that is not part of the header.
---
 Source/App/EncApp/EbAppInputy4m.c | 1 +
 1 file changed, 1 insertion(+)

diff --git a/Source/App/EncApp/EbAppInputy4m.c b/Source/App/EncApp/EbAppInputy4m.c
index ff7c6041..950cb8e7 100644
--- a/Source/App/EncApp/EbAppInputy4m.c
+++ b/Source/App/EncApp/EbAppInputy4m.c
@@ -29,6 +29,7 @@ char* copyUntilCharacterOrNewLine(char *src, char *dst, char chr){
 int32_t read_y4m_header(EbConfig *cfg){
     FILE *ptr_in;
     char buffer[YFM_HEADER_MAX];
+    memset(buffer, 'W', sizeof(buffer));
     char *fresult, *tokstart, *tokend, format_str[YFM_HEADER_MAX];
     uint32_t bitdepth = 8, width = 0, height = 0, fr_n = 0,
         fr_d = 0, aspect_n, aspect_d;
-- 
2.20.1```

You can also reuse the assert added in this patch. It checks that the parsing algorithm does not overflow. It is triggered in the master branch.
https://github.com/OpenVisualCloud/SVT-AV1/blob/ea04b0dc9535ff55126427dc9d66515d41e16a6f/Source/App/EncApp/EbAppInputy4m.c#L220

**Comment by JackZhouVCD** [comment]
@magwyz I'm still not too sure how to replicate the read overflow issue on my end.

It seems that nothing goes wrong when the encoding is set to finish when the clip ends (without setting the `-n` switch). `akiyo_cif.y4m` has 300 frames, when I set the encoder to encode 303 frames (`-n 303`) with the following command:

```css
.\Bin\Release\Release\SvtAv1EncApp.exe -i C:\stream\akiyo_cif.y4m -n 303
```
I get this output:

```css
SVT-AV1 Encoder
SVT [version]:  SVT-AV1 Encoder Lib v0.7.0
SVT [build]  :  Visual Studio 2017 64 bit
LIB Build date: Oct 22 2019 16:10:31
Number of logical cores available: 8
Number of PPCS 72
SVT [config]: Main Profile      Tier (auto)     Level (auto)
SVT [config]: EncoderMode                                                       : 8
SVT [config]: EncoderBitDepth / EncoderColorFormat / CompressedTenBitFormat     : 8 / 1 / 0
SVT [config]: SourceWidth / SourceHeight                                        : 352 / 288
SVT [config]: Fps_Numerator / Fps_Denominator / Gop Size / IntraRefreshType     : 30000 / 1001 / 32 / 1
SVT [config]: HierarchicalLevels / BaseLayerSwitchMode / PredStructure          : 4 / 0 / 2
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange                  : CQP / 50 / 33 / 0
Encoding       221Failed to read proper y4m frame delimeter. Read broken.
      222Failed to read proper y4m frame delimeter. Read broken.
      303
SUMMARY --------------------------------- Channel 1  --------------------------------
Total Frames            Frame Rate              Byte Count              Bitrate
         303            29.97 fps                   125824              99.56 kbps


Channel 1
Average Speed:          148.166 fps
Total Encoding Time:    2045 ms
Total Execution Time:   2283 ms
Average Latency:        468 ms
Max Latency:            644 ms
Encoder finished
```

I applied your changes and I am seeing this too. Is this what you were referring to by the read overflows?

magwyz:
@JackZhouVCD The overflow I am talking about is in the Y4M header parsing. It is not related to the `-n` argument.
The problem is that the function that parses the Y4M header processes data beyond the buffer that is read from the file.
Since the triggering of this bug depends on the state of the memory, to always trigger it, you must force the initialization of the buffer with specific data as mentioned in my previous comment.

magwyz:
I have just rebased the branch.

PR #704, opened 2019-10-16
Valgrind bug fix

ttrigui:
call non-square decision only when non-square is enabled


hassount:
@ttrigui, can we have the BDR impact of this change ?

hassount:
@ttrigui is this PR still valid ?

ttrigui:
> @ttrigui is this PR still valid ?

no.I'll close it.

PR #707, opened 2019-10-16
Two Pass Encoding

anaghdin:
## Description:
Two pass encoding. For now, the encoder is called two times and data transferred using file.
First pass collect the referenced area per SB and the second pass uses it to: 
- Frame QP assignment 
- SB QP assignment 
- Temporal filtering strength adjustment

A config parameter (-enc-mode-2p) is passed to the first pass to use the ME settings of the second pass (to achieve better bdRate)

### First pass:
SvtAv1EncApp.exe -enc-mode 5  -w 1920  -h 1080  -i -q 20  -fps-num 60000 -fps-denom 1001 -b out.bin   -n 60  -intra-period 119  **-output-stat-file out.stat**   -bit-depth 8 
### Second pass:
SvtAv1EncApp.exe -enc-mode 0  -w 1920  -h 1080  -i aspen_1080p_60f.yuv  -q 20  -fps-num 60000 -fps-denom 1001 -b out-m0.bin   -n 60  -intra-period 119   **-input-stat-file in.stat**   -bit-depth 8

## Authors
@anaghdin

## Type of change
New Feature

## Tests and performance
Expected Average Y-PSNR BD-rate gain in the range of -2.4% on the objective fast 1 test set, using 4 QP values: {20, 32, 43 and 55}.

No Speed impact.

### DIFF
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -12,8 +12,13 @@ extern "C" {
 
 #include "stdint.h"
 #include "EbSvtAv1.h"
-
+#if 1 //TWO_PASS
+#include <stdlib.h>
+#include <stdio.h>
+#endif
 #define TILES    1
+#define TWO_PASS_USE_2NDP_ME_IN_1STP                1
+#define TWO_PASS                                    1
 #define ALT_REF_OVERLAY_APP                         1
     //***HME***
 #define EB_HME_SEARCH_AREA_COLUMN_MAX_COUNT         2
@@ -39,7 +44,13 @@ typedef struct EbSvtAv1EncConfiguration
      *
      * Default is defined as MAX_ENC_PRESET. */
     uint8_t                  enc_mode;
-
+#if TWO_PASS_USE_2NDP_ME_IN_1STP
+    /* For two pass encoding, the enc_mod of the second pass is passed in the first pass.
+    * First pass has the option to run with second pass ME settings.
+    *
+    * Default is defined as MAX_ENC_PRESET. */
+    uint8_t                  snd_pass_enc_mode;
+#endif
     // GOP Structure
 
     /* The intra period defines the interval of frames after which you insert an
@@ -180,7 +191,12 @@ typedef struct EbSvtAv1EncConfiguration
     *
     * Default is 0.*/
     EbBool                   use_qp_file;
-
+#if 1 //TWO_PASS
+    /* Input stats file */
+    FILE                    *input_stat_file;
+    /* output stats file */
+    FILE                    *output_stat_file;
+#endif
     /* Enable picture QP scaling between hierarchical levels
     *
     * Default is null.*/
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -30,6 +30,10 @@
 #define OUTPUT_RECON_TOKEN              "-o"
 #define ERROR_FILE_TOKEN                "-errlog"
 #define QP_FILE_TOKEN                   "-qp-file"
+#if TWO_PASS
+#define INPUT_STAT_FILE_TOKEN           "-input-stat-file"
+#define OUTPUT_STAT_FILE_TOKEN          "-output-stat-file"
+#endif
 #define STAT_FILE_TOKEN                 "-stat-file"
 #define WIDTH_TOKEN                     "-w"
 #define HEIGHT_TOKEN                    "-h"
@@ -46,6 +50,9 @@
 #define ENCODER_COLOR_FORMAT            "-color-format"
 #define INPUT_COMPRESSED_TEN_BIT_FORMAT "-compressed-ten-bit-format"
 #define ENCMODE_TOKEN                   "-enc-mode"
+#if TWO_PASS_USE_2NDP_ME_IN_1STP
+#define ENCMODE2P_TOKEN                 "-enc-mode-2p"
+#endif
 #define HIERARCHICAL_LEVELS_TOKEN       "-hierarchical-levels" // no Eval
 #define PRED_STRUCT_TOKEN               "-pred-struct"
 #define INTRA_PERIOD_TOKEN              "-intra-period"
@@ -179,6 +186,21 @@ static void SetCfgQpFile                        (const char *value, EbConfig *cf
     if (cfg->qp_file) { fclose(cfg->qp_file); }
     FOPEN(cfg->qp_file,value, "r");
 };
+#if TWO_PASS
+static void set_input_stat_file(const char *value, EbConfig *cfg)
+{
+    if (cfg->input_stat_file) { fclose(cfg->input_stat_file); }
+    FOPEN(cfg->input_stat_file, value, "rb");
+};
+static void set_output_stat_file(const char *value, EbConfig *cfg)
+{
+    if (cfg->output_stat_file) { fclose(cfg->output_stat_file); }
+    FOPEN(cfg->output_stat_file, value, "wb");
+};
+#if TWO_PASS_USE_2NDP_ME_IN_1STP
+static void set_snd_pass_enc_mode(const char *value, EbConfig *cfg) { cfg->snd_pass_enc_mode = (uint8_t)strtoul(value, NULL, 0); };
+#endif
+#endif
 static void SetCfgStatFile(const char *value, EbConfig *cfg)
 {
     if (cfg->stat_file) { fclose(cfg->stat_file); }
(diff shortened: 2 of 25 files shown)


EwoutH:
Is scene transition detection for keyframe placement also included or planned for a future PR?

> First pass: -use-output-stat-file 1 -output-stat-file out.stat
> Second pass: -use-input-stat-file 1 -input-stat-file in.stat

This seems redundant. Can we just use `-output-stat-file file.stat` to detect a first pass and `-input-stat-file file.stat` to detect a second pass? Then we don't need `-use-output-stat-file` and `-use-input-stat-file`. Or otherwise maybe:

> First pass: -pass 1 -output-stat-file file.stat
> Second pass: -pass 2 -input-stat-file file.stat

anaghdin:
Thanks for the suggestion. It makes sense. I will modify to  "-output-stat-file file.stat to detect a first pass and -input-stat-file file.stat to detect a second pass".

scene transition detection is not included in this PR. We still need to address some related issues.

mafonsoN:
@anaghdin can the user decide at which enc-mode to run on the first pass? 

IMO, it would be better to have a single command line to run a two-pass encode by using, for example, `--passes 2`, and the encoder would take care of doing the multiple encodes strategy.

anaghdin:
@mafonsoN Yes, we can choose any mode at the first pass. Using M5 in the first pass is only 3% of M0 (the second pass)
I agree, we are working on having a single command line.

EwoutH:
@anaghdin Thanks for adopting the changes!

@mafonsoN I agree a single CLI option to run two passes is useful for the end user. However being able to run only a first or second pass or use another stat-file is useful, so let's implement both.

mafonsoN:
@EwoutH that's true. Both options is even better! 

@anaghdin did you try different enc-modes for the first pass to check how the performance varies? For example, if using enc-mode 1 instead of enc-mode 5 for the first pass produces higher compression efficiency when running enc-mode 0 for the second pass?

anaghdin:
@mafonsoN We have tried using M3-M8 as the first pass, but the difference is around 0.2%. I do not expect using slower mode as the first pass brings more gain at this time.

ultrawide:
I ran a output difference test on 360P clips in context of M0 by comparing outputs from master (09fc9ca08b76ed45ecb0a01864bef908c6d3c77d) and the head of this PR, no mismatches found. Looks okay to me.

anaghdin:
@mafonsoN @joelsole @hassount  The PR is rebased with the latest master to include the new builds. Any comments on the PR?

hassount:
@joelsole @mafonsoN 
This is up next. any issues with this ?

grill05:
The quality gain is "Y-PSNR BD-rate gain in the range of -2.4% ". From an end-user's perspective, is this significant, or not very significant ?
I'm trying to figure out if this improvement means that svt's video quality(2-pass vbr) will now be similar/better/worse compared to libaom at the same bitrate. Can anyone please explain in simplified terms.

Imroy:
I'm getting lots of "Error in freed  returnVal 1" messages in the second pass.

Source/API/EbSvtAv1Enc.h
```diff
@@ -180,7 +191,12 @@ typedef struct EbSvtAv1EncConfiguration
     *
     * Default is 0.*/
     EbBool                   use_qp_file;
-
+#if 1 //TWO_PASS
```
ultrawide: Change to #if TWO_PASS

Source/App/EncApp/EbAppConfig.c
```diff
@@ -179,6 +186,21 @@ static void SetCfgQpFile                        (const char *value, EbConfig *cf
     if (cfg->qp_file) { fclose(cfg->qp_file); }
     FOPEN(cfg->qp_file,value, "r");
 };
+#if TWO_PASS
+static void set_input_stat_file(const char *value, EbConfig *cfg)
```
ultrawide:
```suggestion
static void SetInputStatFile(const char *value, EbConfig *cfg)
```

Source/App/EncApp/EbAppConfig.c
```diff
@@ -179,6 +186,21 @@ static void SetCfgQpFile                        (const char *value, EbConfig *cf
...
+{
+    if (cfg->input_stat_file) { fclose(cfg->input_stat_file); }
+    FOPEN(cfg->input_stat_file, value, "rb");
+};
+static void set_output_stat_file(const char *value, EbConfig *cfg)
```
ultrawide:
```suggestion
static void SetOutputStatFile(const char *value, EbConfig *cfg)
```

Source/App/EncApp/EbAppConfig.c
```diff
@@ -179,6 +186,21 @@ static void SetCfgQpFile                        (const char *value, EbConfig *cf
...
+{
+    if (cfg->output_stat_file) { fclose(cfg->output_stat_file); }
+    FOPEN(cfg->output_stat_file, value, "wb");
+};
+#if TWO_PASS_USE_2NDP_ME_IN_1STP
+static void set_snd_pass_enc_mode(const char *value, EbConfig *cfg) { cfg->snd_pass_enc_mode = (uint8_t)strtoul(value, NULL, 0); };
```
ultrawide:
```suggestion
static void SetSndPassEncMode(const char *value, EbConfig *cfg) { cfg->snd_pass_enc_mode = (uint8_t)strtoul(value, NULL, 0); };
```

Source/App/EncApp/EbAppConfig.c
```diff
@@ -305,6 +327,10 @@ config_entry_t config_entry[] = {
     { SINGLE_INPUT, OUTPUT_RECON_TOKEN, "ReconFile", SetCfgReconFile },
     { SINGLE_INPUT, QP_FILE_TOKEN, "QpFile", SetCfgQpFile },
     { SINGLE_INPUT, STAT_FILE_TOKEN, "StatFile", SetCfgStatFile },
+#if TWO_PASS
+    { SINGLE_INPUT, INPUT_STAT_FILE_TOKEN, "input_stat_file", set_input_stat_file },
```
ultrawide:
```suggestion
    { SINGLE_INPUT, INPUT_STAT_FILE_TOKEN, "input_stat_file", SetInputStatFile},
```

Source/App/EncApp/EbAppConfig.c
```diff
@@ -305,6 +327,10 @@ config_entry_t config_entry[] = {
...
+    { SINGLE_INPUT, OUTPUT_STAT_FILE_TOKEN, "output_stat_file", set_output_stat_file },
```
ultrawide:
```suggestion
    { SINGLE_INPUT, OUTPUT_STAT_FILE_TOKEN, "output_stat_file", SetOutputStatFile},
```

Source/App/EncApp/EbAppConfig.c
```diff
@@ -317,6 +343,9 @@ config_entry_t config_entry[] = {
     { SINGLE_INPUT, BUFFERED_INPUT_TOKEN, "BufferedInput", SetBufferedInput },
     { SINGLE_INPUT, BASE_LAYER_SWITCH_MODE_TOKEN, "BaseLayerSwitchMode", SetBaseLayerSwitchMode },
     { SINGLE_INPUT, ENCMODE_TOKEN, "EncoderMode", SetencMode},
+#if TWO_PASS_USE_2NDP_ME_IN_1STP
+    { SINGLE_INPUT, ENCMODE2P_TOKEN, "EncoderMode2p", set_snd_pass_enc_mode},
```
ultrawide:
```suggestion
    { SINGLE_INPUT, ENCMODE2P_TOKEN, "EncoderMode2p", SetSndPassEncMode},
```

Source/API/EbSvtAv1Enc.h
ultrawide:
```suggestion
#if TWO_PASS
```

anaghdin:
@mafonsoN No the two pass command line is not ready yet. We will have another PR after.
I updated the description: "A config parameter (-enc-mode-2p) is passed to the first pass to use the ME settings of the second pass (to achieve better bdRate)"
Eventually, first pass and second pass will be combined, and they can share many encoder components (i.e. ME). So, we are adding the option for the first pass to use the second pass me.

Source/App/EncApp/EbAppConfig.c
anaghdin: We have switched to use snake_case. All the code with CamelCase will be converted to snake_case.

EwoutH:
@anaghdin Could you update the [svt-av1-encoder-user-guide.md](https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Docs/svt-av1_encoder_user_guide.md) accordingly?

PR #711, opened 2019-10-17
Fix memory leak 

ttrigui:
change the allocation of scratch_candidate_buffer to fix the memory leak


hassount:
@ttrigui 
there seems to be a segmentation fault in the binary identical test:
https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/246840463

PR #716, opened 2019-10-18
NEW_NEAR & NEAR_NEW support

### LINKED ISSUE #715: NEAR_NEW & NEW_NEAR mode support
add support for NEAR_NEW & NEW_NEAR modes

okhlif:
**Description**
NEW_NEAR & NEAR_NEW support

Closes #715 

**Author**
@chkngit
@okhlif

**Type of change**
Enhancement

**Tests and performance**
BDR gain is 0.04% [All objective-fast1 clips ]
No memory change
Speed cost is about 3% [4 core machine]

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -47,6 +47,7 @@ extern "C" {
 #define OBMC_FLAG            1 // OBMC motion mode flag
 #define OBMC_CONVOLVE        1 // to track convolve kernels changes
 
+#define INJECT_NEW_NEAR_NEAR_NEW   1   // Inject NEW_NEAR / NEAR_NEW inter prediction 
 
 
 #define II_COMP_FLAG 1
--- a/Source/Lib/Common/Codec/EbInterPrediction.c
+++ b/Source/Lib/Common/Codec/EbInterPrediction.c
@@ -7065,7 +7065,7 @@ EbErrorType inter_pu_prediction_av1(
     {
         candidate_buffer_ptr->candidate_ptr->interp_filters = 0;
     } else {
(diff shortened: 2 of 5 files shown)


joelsole:
The BD-rate cost vs speed trade-off is bad. What is the reason to inject these modes?

chkngit:
> The BD-rate cost vs speed trade-off is bad. What is the reason to inject these modes?

how can you quantify scientifically that the trade-off is good or bad? we think that for M0, a gain of 0.4% should not cost more than 4% speed drop, so for this case, it  passes this thereshold. anyway, it is good to support those modes in the code, so later we have them handy to use. we can always re-evaluate the tradeoff and consider shut the injection

okhlif:
Fixed the braces , thank you. These are missing modes that the standard supports. 
I think the trad off is still reasonable for M0 that can always be reevaluated.

joelsole:
0.4% sounds good.
The OP states 0.04%.

chkngit:
well it is actually 0.04!!

joelsole:
I agree that it is good to support these modes in the code for future optimizations.

Note that the trade-off 0.04% BD-rate with a 3% cost is equivalent to doubling the encoder complexity to get a 1% BD-rate gain. Also, 0.04% could be more noise than signal.

okhlif:
> I agree that it is good to support these modes in the code for future optimizations.
> 
> Note that the trade-off 0.04% BD-rate with a 3% cost is equivalent to doubling the encoder complexity to get a 1% BD-rate gain. Also, 0.04% could be more noise than signal.

Actually it depends on resolution, where we see 0.12% for 360p. I don't believe the gain is noise.

chkngit:
thanks for approving to get these modes supported in the code. more work is definitely needed to make the trade off better

PR #719, opened 2019-10-18
Add link to performance tracker website in README.md

joelsole:


grill05:
Those graphs are very interesting. I have a couple of suggestions

1. AV1's closest competitor is libx265 and libvpx-vp9. If BD-rate curves of SVT-AV1/LIBAOM vs libx265 (and maybe even libvpx-vp9) were included, that would be very interesting to track

2. From an user's perspective, the time-performance can matter as much as the quality performance. I understand that BD-rate curves can't measure changes in encoding speed. But maybe a separate graph for how encoding speed (for LIBAOM/SVT-AV1) is changing over time would be very useful

joelsole:
Thanks for your input.
Regarding your second point: we put a note indicating that "encoding times are not reliable", but we are planning to have a set-up with more reliable timing. Working on that.

PR #723, opened 2019-10-21
add picture operator test

Cidana-Developers:
1. add picture copy, average func test
  - picture_average_kernel_sse2_intrin
  - picture_average_kernel1_line_sse2_intrin
  - picture_copy_kernel_sse2

### DIFF
--- a/Source/Lib/Common/C_DEFAULT/EbPictureOperators_C.c
+++ b/Source/Lib/Common/C_DEFAULT/EbPictureOperators_C.c
@@ -31,6 +31,17 @@ void picture_average_kernel(
     }
 }
 
+void picture_average_kernel_1line_c(
+    EbByte   src0,
+    EbByte   src1,
+    EbByte   dst,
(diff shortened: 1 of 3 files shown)


Source/Lib/Common/C_DEFAULT/EbPictureOperators_C.c
```diff
@@ -31,6 +31,17 @@ void picture_average_kernel(
     }
 }
 
+void picture_average_kernel_1line_c(
```
andrey-norkin: This might be a useful function in the codec library. But unless we are planning to use it in the encoder, it should probably be in the test folder. @hassount, Do we have a use for this function (or do we have something similar elsewhere in the code)?

andrey-norkin: Never mind, this looks like the C function for the intrinsics, it should be where it is.

PR #725, opened 2019-10-21
[SIMD] Add combined_averaging_ssd_avx512() and eb_av1_txb_init_levels_avx512()

lzhangnj:
Add combined_averaging_ssd_avx512()

Speed comparing with original avx2:
 8x8 :  3.61x
16x16:  7.67x
32x32: 14.34x
64x64: 18.31x

Update eb_av1_txb_init_levels_avx2()

Speed comparison:
 4x 4: 2.12x
 4x 8: 2.01x
 4x16: 1.62x
 8x 4: 1.82x
 8x 8: 1.53x
 8x16: 1.47x
 8x32: 1.43x
16x 4: 1.38x
16x 8: 1.49x
16x16: 1.47x
16x32: 1.36x
32x 8: 1.52x
32x16: 1.15x
32x32: 1.24x

Add eb_av1_txb_init_levels_avx512()

About 10% to 30% faster for 16x and 32x.

### DIFF
--- a/Source/Lib/Common/ASM_AVX2/EbCombinedAveragingSAD_Intrinsic_AVX2.c
+++ b/Source/Lib/Common/ASM_AVX2/EbCombinedAveragingSAD_Intrinsic_AVX2.c
@@ -3,12 +3,11 @@
 * SPDX - License - Identifier: BSD - 2 - Clause - Patent
 */
 
-#include "EbCombinedAveragingSAD_Intrinsic_AVX2.h"
 #include "immintrin.h"
+#include "EbCombinedAveragingSAD_Inline_AVX2.h"
+#include "EbCombinedAveragingSAD_Intrinsic_AVX2.h"
 #include "EbMemory_AVX2.h"
-
-#define _mm256_set_m128i(/* __m128i */ hi, /* __m128i */ lo) \
-    _mm256_insertf128_si256(_mm256_castsi128_si256(lo), (hi), 0x1)
+#include "EbMemory_SSE4_1.h"
 
 uint32_t combined_averaging8x_msad_avx2_intrin(
     uint8_t  *src,
@@ -22,10 +21,10 @@ uint32_t combined_averaging8x_msad_avx2_intrin(
 {
     __m256i sum = _mm256_setzero_si256();
     __m128i sad;
-    uint32_t y;
+    uint32_t y = height;
     (void)width;
 
-    for (y = 0; y < height; y += 4) {
+    do {
         const __m256i s = load_u8_8x4_avx2(src, src_stride);
         const __m256i r1 = load_u8_8x4_avx2(ref1, ref1_stride);
         const __m256i r2 = load_u8_8x4_avx2(ref2, ref2_stride);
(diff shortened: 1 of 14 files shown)


JackZhouVCD:
Streams matched for all encode modes after applying this PR, and I'm seeing a -0.33% speed drop averaged over M{1,4,8}

lzhangnj:
> Streams matched for all encode modes after applying this PR, and I'm seeing a -0.33% speed drop averaged over M{1,4,8}

I suspect the removing of the static function tables caused the speed drop. I just restored it.
Could you please help retest?

Thanks!

lzhangnj:
> Yup, I reran the test and I'm seeing a 0.13% speed gain now
Great and Thanks!

EwoutH:
@JackZhouVCD Are you sure it's not just noise you're measuring? Do you have standard deviation or 95% confidence intervals?

JackZhouVCD:
> @JackZhouVCD Are you sure it's not just noise you're measuring? Do you have standard deviation or 95% confidence intervals?

@EwoutH I ran the speed benchmark 5 more times. The average of the 6 runs is 0.16% with a standard deviation of 0.303%. I'm doing as much as I can to minimize the noise of the testing system, but indeed we might be seeing some noisy data as you suggested. 

In cases like this I am simply making sure that this Pull Request isn't bringing any unexpected speed drops or output differences (of which I have already validated that the changes are lossless). After all, it is kind of difficult to be certain of how much faster AVX512 optimizations are at this time, due to hardware limitations.

Source/Lib/Common/ASM_AVX2/EbCombinedAveragingSAD_Intrinsic_AVX2.c
```diff
@@ -219,18 +223,18 @@ uint32_t combined_averaging64x_msad_avx2_intrin(
...
+    do {
         sum = CombinedAveragingSad32_AVX2(src + 0x00,
             ref1 + 0x00, ref2 + 0x00, sum);
         sum = CombinedAveragingSad32_AVX2(src + 0x20,
             ref1 + 0x20, ref2 + 0x20, sum);
         src += src_stride;
         ref1 += ref1_stride;
         ref2 += ref2_stride;
-    }
+    } while (--y);
```
mafonsoN: This do...while loop could be replaced by a simple for loop which is easier to read IMO. Also, I've noticed you use a lot of do...while() statements throughout the code. Is there a reason for this choice?

lzhangnj: The loop number here 'height' is a variable, and using for loop will have an extra check to see if height > 0 before the first loop occurs. Using do ... while() will save this check. If loop number is a constant, definitely using a for() loop is a better choice.

mafonsoN:
@lzhangnj can you rebase so that we can proceed with merging this patch?

lzhangnj:
> @lzhangnj can you rebase so that we can proceed with merging this patch?

Done. Thanks.

ttrigui:
@joelsole any issues with this PR?

PR #737, opened 2019-10-24
Add missing IFFT functions for C version with corresponding tests

tszumski:


hassount:
@ttrigui good to go ?

PR #740, opened 2019-10-24
Fix run-to-run in hbd-md 1

### LINKED ISSUE #728: hbd-md 1 debug crash
hbd-md 1 mode debug crash and run to run

okhlif:
### **Description**
1. Fix when hbd-md 1:
- run to run in release
- crash in debug  
2. created a function table :  av1_inter_prediction_function_table for 8 and 10 bit
3. Code clean up 
Fixes #728 

### **Author**
@okhlif 

### **Type of change**
Bug fix

### **Tests and performance**
BDR : ~0.9% gain on 360p objective-fast1 clips ( changed to 10bit)

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -378,7 +378,7 @@ config_entry_t config_entry[] = {
     { SINGLE_INPUT, LOCAL_WARPED_ENABLE_TOKEN, "LocalWarpedMotion", SetEnableLocalWarpedMotionFlag },
     // OBMC
     { SINGLE_INPUT, OBMC_TOKEN, "Obmc", SetEnableObmcFlag },
-	// Filter Intra
+    // Filter Intra
     { SINGLE_INPUT, FILTER_INTRA_TOKEN, "FilterIntra", SetEnableFilterIntraFlag },
     // ME Tools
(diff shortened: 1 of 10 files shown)


joelsole:
Very good clean up. 

For my local tests, the issue is independent of hbd-md (0 or 1) or single pass/two pass.  It seems to be related to 10-bit encoding. 
Applying this PR, the issue still shows up, but many frames later. No issues when applying it and obmc is off.

hassount:
@okhlif @joelsole 
Does this PR solve #728 or does not partially solve it ? if it partially does, then we should change the reference in the description to make sure such is clear

okhlif:
> Very good clean up.
> 
> For my local tests, the issue is independent of hbd-md (0 or 1) or single pass/two pass. It seems to be related to 10-bit encoding.
> Applying this PR, the issue still shows up, but many frames later. No issues when applying it and obmc is off.

I did not see such in my local test. There is still a r2r ( the same one as the 8 bit ) 
@hassount  it does solve the issue I saw in #728 (crash and significant BDR difference from run to run) due to uninitialized variable.

joelsole:
We could get this PR in soon - at it is good clean-up and blocks other 10-bit development. 
However, @hassount , shall we address first the drop in performance in yesterday's regression tests?

PR #746, opened 2019-10-26
Add combined test mode for two pass.

jsunintel:
Usage: SvtAv1EncApp -i /work/clips/matrix_1920x1080_250i.yuv -n 30 -w 1920 -h 1080 -enc-mode 5 -secondary-enc-mode 0 -b out.ivf

### DIFF
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -210,9 +210,6 @@ typedef struct EbSvtAv1EncConfiguration
      * Default is defined as MAX_ENC_PRESET. */
     uint8_t                  enc_mode2p;
 #endif
-
-    /* Stat data generated in the 1st pass and used in the 2nd */
-    uint8_t                  *stat_buffer;
 #endif
 
     /* Enable picture QP scaling between hierarchical levels
@@ -532,6 +529,16 @@ typedef struct EbSvtAv1EncConfiguration
     EB_API EbErrorType eb_svt_get_recon(
         EbComponentType      *svt_enc_component,
         EbBufferHeaderType   *p_buffer);
+
+    /* OPTIONAL: Fill buffer with stat data.
+     *
+     * Parameter:
+     * @ *svt_enc_component  Encoder handler.
+     * @ *p_buffer           Output buffer. */
+    EB_API EbErrorType eb_svt_get_stat(
+        EbComponentType      *svt_enc_component,
+        EbBufferHeaderType   *p_buffer,
+        EbBool               finished);
 
     /* STEP 6: Deinitialize encoder library.
      *
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -29,6 +29,11 @@ extern "C" {
 #define EB_BUFFERFLAG_HAS_TD        0x00000004  // signals that the packet contains a TD
 #define EB_BUFFERFLAG_IS_ALT_REF    0x00000008  // signals that the packet contains an ALT_REF frame
 #define EB_BUFFERFLAG_ERROR_MASK    0xFFFFFFF0  // mask for signalling error assuming top flags fit in 4 bits. To be changed, if more flags are added.
+
+#if 1 //TWO_PASS
+#define STAT_BUFFER_UNIT 11680
+#endif
+
 // Will contain the EbEncApi which will live in the EncHandle class
 // Only modifiable during config-time.
 typedef struct EbSvtAv1EncConfiguration
@@ -205,6 +210,13 @@ typedef struct EbSvtAv1EncConfiguration
 *
 * Default is 0.*/
     EbBool                   use_output_stat_file;
+
+    /* The enc mode for the second pass */
+    int8_t                   secondary_enc_mode;
+
+    /* Stat data generated in the 1st pass and used in the 2nd */
+    uint8_t                  *stat_buffer;
+
 #endif
     /* Enable picture QP scaling between hierarchical levels
     *
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -189,10 +189,17 @@ typedef struct EbSvtAv1EncConfiguration
     EbBool                   use_qp_file;
 
 #if 1 //TWO_PASS
-    /* The two passes switch */
+    /* The two passes switch
+     * 1 for single pass, first or second pass
+     * 2 for two passes
+     * Default is 1 */
     uint8_t                  passes;
 
-    /* Specify which pass */
+    /* Specify which pass
+     * 1 for first pass
+     * 2 for second pass
+     * 0 for not-specified
+     * Default is 0 */
     uint8_t                  pass;
 
 #if TWO_PASS_USE_2NDP_ME_IN_1STP
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -195,9 +195,6 @@ typedef struct EbSvtAv1EncConfiguration
     /* Specify which pass */
     uint8_t                  pass;
 
-    /* First pass file */
-    FILE                     *fpf;
-
 #if TWO_PASS_USE_2NDP_ME_IN_1STP
      /* A preset defining the quality vs density tradeoff point that the second pass
      * encoding is to be performed at. 0 is the highest quality mode, 3 is the highest
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -47,15 +47,6 @@ typedef struct EbSvtAv1EncConfiguration
      * Default is defined as MAX_ENC_PRESET. */
     uint8_t                  enc_mode;
 
-#if TWO_PASS_USE_2NDP_ME_IN_1STP
-     /* A preset defining the quality vs density tradeoff point that the second pass
-     * encoding is to be performed at. 0 is the highest quality mode, 3 is the highest
-     * density mode.
-     *
-     * Default is defined as MAX_ENC_PRESET. */
-    uint8_t                  enc_mode2p;
-#endif
-
     // GOP Structure
 
     /* The intra period defines the interval of frames after which you insert an
@@ -198,26 +189,28 @@ typedef struct EbSvtAv1EncConfiguration
     EbBool                   use_qp_file;
 
 #if 1 //TWO_PASS
-    /* Input stats file */
-    FILE                    *input_stat_file;
-    /* output stats file */
-    FILE                    *output_stat_file;
-    /* Get stats from the input_stat_file
-*
-* Default is 0.*/
-    EbBool                   use_input_stat_file;
-    /* Write stats to the input_stat_file
-*
-* Default is 0.*/
-    EbBool                   use_output_stat_file;
+    /* The two passes switch */
+    uint8_t                  passes;
 
-    /* The enc mode for the second pass */
-    int8_t                   secondary_enc_mode;
+    /* Specify which pass */
+    uint8_t                  pass;
+
+    /* First pass file */
+    FILE                     *fpf;
+
+#if TWO_PASS_USE_2NDP_ME_IN_1STP
+     /* A preset defining the quality vs density tradeoff point that the second pass
+     * encoding is to be performed at. 0 is the highest quality mode, 3 is the highest
+     * density mode.
+     *
+     * Default is defined as MAX_ENC_PRESET. */
+    uint8_t                  enc_mode2p;
+#endif
 
     /* Stat data generated in the 1st pass and used in the 2nd */
     uint8_t                  *stat_buffer;
-
 #endif
+
     /* Enable picture QP scaling between hierarchical levels
     *
     * Default is null.*/
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -29,11 +29,12 @@
 #define ERROR_FILE_TOKEN                "-errlog"
 #define QP_FILE_TOKEN                   "-qp-file"
 #if 1 //TWO_PASS
-#define INPUT_STAT_FILE_TOKEN           "-input-stat-file"
-#define OUTPUT_STAT_FILE_TOKEN          "-output-stat-file"
-#define USE_INPUT_STAT_FILE_TOKEN       "-use-input-stat-file"
-#define USE_OUTPUT_STAT_FILE_TOKEN      "-use-output-stat-file"
-#define SECONDARY_ENCMODE_TOKEN         "-secondary-enc-mode"
+#define PASSES_TOKEN                    "-passes"
+#define PASS_TOKEN                      "-pass"
+#define FPF_TOKEN                       "-fpf"
+#if TWO_PASS_USE_2NDP_ME_IN_1STP
+#define ENCMODE2P_TOKEN                 "-enc-mode-2p"
+#endif
 #endif
 #define WIDTH_TOKEN                     "-w"
 #define HEIGHT_TOKEN                    "-h"
@@ -49,9 +50,6 @@
 #define ENCODER_COLOR_FORMAT            "-color-format"
 #define INPUT_COMPRESSED_TEN_BIT_FORMAT "-compressed-ten-bit-format"
 #define ENCMODE_TOKEN                   "-enc-mode"
-#if TWO_PASS_USE_2NDP_ME_IN_1STP
-#define ENCMODE2P_TOKEN                 "-enc-mode-2p"
-#endif
 #define HIERARCHICAL_LEVELS_TOKEN       "-hierarchical-levels" // no Eval
 #define PRED_STRUCT_TOKEN               "-pred-struct"
 #define INTRA_PERIOD_TOKEN              "-intra-period"
@@ -167,22 +165,12 @@ static void SetCfgQpFile                        (const char *value, EbConfig *cf
     FOPEN(cfg->qp_file,value, "r");
 };
 #if 1 //TWO_PASS
-static void set_input_stat_file(const char *value, EbConfig *cfg)
+static void SetPasses                           (const char *value, EbConfig *cfg) {cfg->passes = (uint8_t)strtoul(value, NULL, 0);};
+static void SetPass                             (const char *value, EbConfig *cfg) {cfg->pass = (uint8_t)strtoul(value, NULL, 0);};
+static void SetFpf                              (const char *value, EbConfig *cfg)
 {
-    if (cfg->use_input_stat_file) {
-        if (cfg->input_stat_file) { fclose(cfg->input_stat_file); }
-        FOPEN(cfg->input_stat_file, value, "rb");
-    }
+    EB_STRNCPY(cfg->fpf_name, sizeof(cfg->fpf_name), value, sizeof(cfg->fpf_name));
 };
-static void set_output_stat_file(const char *value, EbConfig *cfg)
-{
-    if (cfg->use_output_stat_file) {
-        if (cfg->output_stat_file) { fclose(cfg->output_stat_file); }
-        FOPEN(cfg->output_stat_file, value, "wb");
-    }
-};
-static void set_use_input_stat_file(const char *value, EbConfig *cfg) { cfg->use_input_stat_file = (EbBool)strtol(value, NULL, 0); };
-static void set_use_output_stat_file(const char *value, EbConfig *cfg) { cfg->use_output_stat_file = (EbBool)strtol(value, NULL, 0); };
 #if TWO_PASS_USE_2NDP_ME_IN_1STP
 static void SetencMode2p                        (const char *value, EbConfig *cfg) {cfg->enc_mode2p = (uint8_t)strtoul(value, NULL, 0);};
 #endif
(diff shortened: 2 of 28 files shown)


andrey-norkin:
ffmpeg patch seems failing though

jsunintel:
> 
> 
> ffmpeg patch seems failing though

Hi Andrey, I checked the ffmpeg building failure and it seems that had nothing to do with SVT AV1:

libavcodec/libavcodec.a(hq_hqa.o): In function `hq_decode_block':

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:74: undefined reference to `ff_hq_quants'

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:83: undefined reference to `ff_hq_ac_skips'

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:86: undefined reference to `ff_hq_ac_syms'

libavcodec/libavcodec.a(hq_hqa.o): In function `hq_decode_frame':

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:129: undefined reference to `ff_hq_profile'

libavcodec/libavcodec.a(hq_hqa.o): In function `hq_decode_block':

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:72: undefined reference to `ff_hq_quants'

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:83: undefined reference to `ff_hq_ac_skips'

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:86: undefined reference to `ff_hq_ac_syms'

libavcodec/libavcodec.a(hq_hqa.o): In function `hq_decode_frame':

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:126: undefined reference to `ff_hq_profile'

libavcodec/libavcodec.a(hq_hqa.o): In function `hq_hqa_decode_init':

/home/travis/build/OpenVisualCloud/SVT-AV1/ffmpeg/libavcodec/hq_hqa.c:371: undefined reference to `ff_hq_init_vlcs'

collect2: error: ld returned 1 exit status

Makefile:108: recipe for target 'ffmpeg_g' failed

anaghdin:
Hi, Could you update the description with the new CLI for both 1 pass and 2 pass? Also, please update the user guide accordingly.

anaghdin:
@hassount could you please review this PR?

Source/API/EbSvtAv1Enc.h
```diff
@@ -193,19 +189,25 @@ typedef struct EbSvtAv1EncConfiguration
...
-    FILE                    *output_stat_file;
-    /* Get stats from the input_stat_file
-*
-* Default is 0.*/
-    EbBool                   use_input_stat_file;
-    /* Write stats to the input_stat_file
-*
-* Default is 0.*/
-    EbBool                   use_output_stat_file;
+    /* The two passes switch */
```
hassount: need a default value

Source/API/EbSvtAv1Enc.h
```diff
@@ -193,19 +189,25 @@ typedef struct EbSvtAv1EncConfiguration
...
+    uint8_t                  passes;
+
+    /* Specify which pass */
```
hassount: need more description on pass and passes for the user, also what is the default value

Source/App/EncApp/EbAppMain.c
```diff
@@ -146,6 +147,61 @@ int32_t main(int32_t argc, char* argv[])
             if (configs[0]->target_socket != -1)
                 AssignAppThreadGroup(configs[0]->target_socket);
 
+ #if 1 //TWO_PASS
```
hassount: I don't believe this section should be set as loose code in the main file, I suggest putting it into a function, add a comment on what we are trying to do as it may not be conventional, then moving it to the EbAppConfig.c

Source/App/EncApp/EbAppMain.c
```diff
@@ -146,6 +147,61 @@ int32_t main(int32_t argc, char* argv[])
...
+                        configs[0]->recon_file = NULL;
+                        configs[0]->bitstream_file = NULL;
+
+                        printf("\n[1st Pass]:\n\n");
+                    } else {
+                        configs[0]->pass = 2;
+
+                        configs[0]->enc_mode = configs[0]->enc_mode2p;
+
+                        fseek(configs[0]->input_file, 0, SEEK_SET);
```
hassount: please check if this works with large file sizes > 2GB on windows

Source/App/EncApp/EbAppMain.c
```diff
@@ -293,6 +349,22 @@ int32_t main(int32_t argc, char* argv[])
...
             }
 #endif
+
+#if 1 //TWO_PASS
+                if (combined_stat_test && !pass)
+                    return_errors[0] = de_init_encoder(appCallbacks[0], 0);
+
+                if (separated_first_pass) {
+                    if (configs[0]->fpf == NULL) {
+                        printf("Invalid first pass file to write\n");
```
hassount: this should be handled in the EbAppConfig.c and not here and should be checked at the stage of the VerifySettings

Source/App/EncApp/EbAppMain.c
```diff
@@ -293,6 +349,22 @@ int32_t main(int32_t argc, char* argv[])
...
+                    } else {
+                        size_t size = fwrite(configs[0]->stat_buffer, STAT_BUFFER_UNIT, configs[0]->frames_to_be_encoded, configs[0]->fpf);
```
hassount: the write op to the file should be handled similarly to the write stream or write recon files which are combined in the EbAppProcessCmd.c

Source/Lib/Common/Codec/EbEntropyCodingProcess.c
```diff
@@ -524,20 +524,21 @@ static EbBool UpdateEntropyCodingRows(
...
-        printf("Error in fseek  returnVal %i\n", fseek_return_value);
-    fwrite(&stat_struct,
-        sizeof(stat_struct_t),
-        (size_t)1,
-        sequence_control_set_ptr->static_config.output_stat_file);
-    eb_release_mutex(sequence_control_set_ptr->encode_context_ptr->stat_file_mutex);
+    eb_block_on_mutex(sequence_control_set_ptr->encode_context_ptr->stat_mutex);
+
+    if (sequence_control_set_ptr->static_config.stat_buffer)
+        memcpy(sequence_control_set_ptr->static_config.stat_buffer + (long)(ref_poc * STAT_BUFFER_UNIT),
```
hassount: the stat buffer is of a static size STAT_BUFFER_UNIT, so an assert would need to be added to make sure we don't for any reason write over bounds

Source/Lib/Common/Codec/EbPredictionStructure.c
```diff
@@ -1873,18 +1873,27 @@ EbErrorType prediction_structure_group_ctor(
...
 #if MRP_M1
+    PredictionStructureConfigEntry *cur_four_level_hierarchical_pred_struct = NULL;
+    PredictionStructureConfigEntry *cur_five_level_hierarchical_pred_struct = NULL;
+    EB_MALLOC(PredictionStructureConfigEntry*, cur_four_level_hierarchical_pred_struct, sizeof(four_level_hierarchical_pred_struct), EB_N_PTR);
+    EB_MALLOC(PredictionStructureConfigEntry*, cur_five_level_hierarchical_pred_struct, sizeof(five_level_hierarchical_pred_struct), EB_N_PTR);
+    EB_MEMCPY(cur_four_level_hierarchical_pred_struct, four_level_hierarchical_pred_struct, sizeof(four_level_hierarchical_pred_struct));
+    EB_MEMCPY(cur_five_level_hierarchical_pred_struct, five_level_hierarchical_pred_struct, sizeof(five_level_hierarchical_pred_struct));
+    PredictionStructureConfigArray[4-1].entry_array = cur_four_level_hierarchical_pred_struct;
+    PredictionStructureConfigArray[5-1].entry_array = cur_five_level_hierarchical_pred_struct;
+
```
hassount: Can you please explain how this part has to do with the 2-pass PR?

Source/API/EbSvtAv1Enc.h
```diff
@@ -193,19 +189,25 @@ typedef struct EbSvtAv1EncConfiguration
...
+     /* A preset defining the quality vs density tradeoff point that the second pass
+     * encoding is to be performed at. 0 is the highest quality mode, 3 is the highest
+     * density mode.
+     *
+     * Default is defined as MAX_ENC_PRESET. */
+    uint8_t                  enc_mode2p;
 #endif
+
+    /* Stat data generated in the 1st pass and used in the 2nd */
+    uint8_t                  *stat_buffer;
```
hassount: passing a large buffer in the stat config creates a limitation here which is, every buffer would need to know the number of frames of the encode. Such might not be a known piece of info for the library. The way this should be handled is that every frame contains a small buffer with stats for that frame only. In the EbBufferHeaderType, there's a section for pieces of info sent for every picture // pic info.

So the way this would work is that, for the 1-pass data is sent-out in that structure from the library and written by the application in the appropriate file. for the 2nd pass, the data is read from the file per picture and sent to the library in for every picture in the same structure.

Source/App/EncApp/EbAppProcessCmd.c
```diff
@@ -1141,11 +1141,15 @@ AppExitConditionType ProcessOutputStreamBuffer(
...
 
     // System performance variables
-    static int32_t         frame_count                = 0;
+    static int32_t         frame_count;
 
     // Local variables
     uint64_t                finishsTime     = 0;
     uint64_t                finishuTime     = 0;
+
+    if (config->performance_context.frame_count == 0)
```
hassount: this would break the multi-instance support. the total frame number is handled by the static variable  frame_count. so if it's reset everytime an instance starts encoding, the app would loose track of the sent frames for the other instances

Source/Lib/Common/Codec/EbEntropyCodingProcess.c
hassount: All of this logic, along with the mutex logic will go away when the data is sent by frame.

Source/Lib/Common/Codec/EbResourceCoordinationProcess.c
```diff
@@ -573,21 +573,21 @@ static void CopyInputBuffer(
...
 {
-    eb_block_on_mutex(sequence_control_set_ptr->encode_context_ptr->stat_file_mutex);
+    eb_block_on_mutex(sequence_control_set_ptr->encode_context_ptr->stat_mutex);
 
-    int32_t fseek_return_value = fseek(sequence_control_set_ptr->static_config.input_stat_file,(long) picture_control_set_ptr->picture_number* sizeof(stat_struct_t), SEEK_SET);
-
-    if (fseek_return_value != 0) {
-        printf("Error in fseek  returnVal %i\n", fseek_return_value);
+    if (sequence_control_set_ptr->static_config.stat_buffer) {
+        if (picture_control_set_ptr->picture_number < sequence_control_set_ptr->static_config.frames_to_be_encoded) {
```
hassount: this is not supposed to be known by the library. in the context of a pipe encode or an ffmpeg stream this info is unknown.
The data as explained above should be sent by frame

Source/API/EbSvtAv1Enc.h
jsunintel: Will do it.

Source/API/EbSvtAv1Enc.h
jsunintel: Will do it.

Source/App/EncApp/EbAppMain.c
jsunintel: Normally we run the encoding for only once, while for the combined test mode, we have to run it twice. I don't think that for  stategy could be put into a function.

Source/App/EncApp/EbAppMain.c
jsunintel: It shall be OK, for here the code tries to seek the file to its beginning and the length is not too large to get handled on Windows. I saw in ReadInputFrames, it also did such an operation.

                if (readSize != headerPtr->n_filled_len) {
                    fseek(input_file, 0, SEEK_SET);
                    headerPtr->n_filled_len = 0;

But to be safer, I am changing it to fseeko64.

Source/App/EncApp/EbAppMain.c
jsunintel: It's already checked in VerifySettings. I will remove the double-check.

Source/Lib/Common/Codec/EbEntropyCodingProcess.c
jsunintel: Didn't get it. What does "the data is sent by frame" mean please?

Source/Lib/Common/Codec/EbPredictionStructure.c
jsunintel: Such as in the case enc_mode 5 then enc_mode 0, the first pass of enc_mode 5 modifies the values in four_level_hierarchical_pred_struct and five_level_hierarchical_pred_struct, those global structures' values are changed and never can get restored, then for the second pass, it works with dirty structures, whose values are not the ones after initialization, and outputs wrong pred_struct, and the referencing info is different and wrong.

Source/API/EbSvtAv1Enc.h
jsunintel: The large buffer is not passed, but only the address of it. 
We can't do that per frame passing, for the sequence of frame processing in Lib is unknown to App. For GOP 30, in one type of cmdline, the sequence of calling read_stat and write_stat is like 0, 25, 1, 2, ..., etc, for another, it might in another sequence.

Source/App/EncApp/EbAppProcessCmd.c
jsunintel: It's reset only when the second pass in combined test mode is started. Without this, the static frame_count keeps incrementing, and the counting becomes wrong:
For the 1st pass:
0...30
For the 2nd Pass:
31...60
And the speed computed is also wrong.

jsunintel:
> 
> 
> Hi, Could you update the description with the new CLI for both 1 pass and 2 pass? Also, please update the user guide accordingly.

The description is updated. There is 0 about 2 passes in the user guide yet. Could I do the guide updates when the 2 pass interfaces are fully decided? I think those are still under improvement.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -847,13 +832,64 @@ static EbErrorType VerifySettings(EbConfig *config, uint32_t channelNumber)
...
+    }
+
+    if (config->passes == 2) {
+        if (config->pass != 0) {
+            fprintf(config->error_log_file, "Error instance %u: Not allowed to specify pass for combined test mode\n", channelNumber + 1);
+            return_error = EB_ErrorBadParameter;
+        }
+
+        if (config->enc_mode >= 4) {
+            fprintf(config->error_log_file, "Warning instance %u: The enc mode %u is too large. Run as single pass\n", channelNumber + 1, config->enc_mode);
```
hassount: suggest changing: "The enc mode %u is too large" to "two pass mode is not supported for enc mode %u"

Source/Lib/Common/Codec/EbEntropyCodingProcess.c
```diff
@@ -524,20 +524,49 @@ static EbBool UpdateEntropyCodingRows(
...
+    EncodeContext *encode_context_ptr = sequence_control_set_ptr->encode_context_ptr;
+    EbObjectWrapper *wrapper_ptr = NULL;
+    EbBufferHeaderType *buffer_ptr = NULL;
+
+    if (sequence_control_set_ptr->static_config.pass != 1) {
+        printf("Invalid pass to write stat\n");
+        return;
+    }
+
+    if (ref_poc >= sequence_control_set_ptr->static_config.frames_to_be_encoded) {
```
hassount: frames_to_be_encoded should not be used inside the library. it will be cleaned up .

Source/Lib/Common/Codec/EbEntropyCodingProcess.c
```diff
@@ -524,20 +524,49 @@ static EbBool UpdateEntropyCodingRows(
...
+    eb_get_empty_object(
+        sequence_control_set_ptr->encode_context_ptr->statistics_output_fifo_ptr,
+        &wrapper_ptr);
+
+    buffer_ptr = (EbBufferHeaderType*)wrapper_ptr->object_ptr;
+
+    ++encode_context_ptr->total_number_of_stat;
+
+    if (encode_context_ptr->total_number_of_stat
+        == sequence_control_set_ptr->static_config.frames_to_be_encoded)
```
hassount: same here, what happens when this is 0 ?

Source/Lib/Common/Codec/EbResourceCoordinationProcess.c
```diff
@@ -569,25 +569,34 @@ static void CopyInputBuffer(
...
-        sizeof(stat_struct_t),
-        (size_t) 1,
-        sequence_control_set_ptr->static_config.input_stat_file);
+
+    if (n_stat_filled_len == 0)
+        return;
+
+    eb_block_on_mutex(sequence_control_set_ptr->encode_context_ptr->stat_mutex);
+
+    if (picture_control_set_ptr->picture_number < sequence_control_set_ptr->static_config.frames_to_be_encoded)
```
hassount: same comment for frames_to_be_encoded

Source/App/EncApp/EbAppMain.c
hassount: > I don't think that for() stategy could be put into a function.

Why not ?

Source/Lib/Common/Codec/EbSystemResourceManager.c
```diff
@@ -670,6 +670,34 @@ EbErrorType eb_get_full_object(
     return return_error;
 }
 
+EbErrorType eb_get_full_object_timeout(
```
hassount: In my opinion this is not a solution we can work with as it is a polling API that will consume a bunch of cycles.
what is the impact of not getting those last few pieces ?

hassount: @chkngit @anaghdin

Source/App/EncApp/EbAppConfig.c
jsunintel: OK

Source/App/EncApp/EbAppConfig.c
```diff
@@ -847,13 +832,64 @@ static EbErrorType VerifySettings(EbConfig *config, uint32_t channelNumber)
...
+            config->passes = 1;
+        } else {
+            if (config->fpf) {
+               fclose(config->fpf);
+               config->fpf = NULL;
+            }
+
+            if (strlen(config->fpf_name) == 0)
+                strcpy(config->fpf_name, "./stat.bin");
```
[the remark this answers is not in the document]
jsunintel: The whole memory optimization is based on a stat file existing. Always write to that file in 1st pass and read from it in 2nd pass. For combined test mode, the -fpf is optional. If users don't specify the file name, we use the default file name.

Source/Lib/Common/Codec/EbEntropyCodingProcess.c
jsunintel: I will update here when it's actually cleaned up. The last sent input frame has no input frame in it for real, but only as the flag of end of stream. This line is for that.

Source/Lib/Common/Codec/EbEntropyCodingProcess.c
jsunintel: The last one tends not be the last POC. The same,  I will update here when it's actually cleaned up, trying to figure something else.

jsunintel: And because the last pieces are missed, we never reach this point for now. It's OK to keep it, until the frames_to_be_encoded is cleaned up.

Source/Lib/Common/Codec/EbSystemResourceManager.c
jsunintel: This optimization should not make the current two_pass work in a worse way. With my solution, the encoding results are equivalent. I think that's important.
The only bad income is extra 1s. For the app finished event is triggered by timeout, which costs 1s.

Source/Lib/Common/Codec/EbPredictionStructure.c
hassount: @chkngit @lijing0010 
Any thoughts on this ?

Source/App/EncApp/EbAppProcessCmd.c
jsunintel: Yes, I tested it and that's why I made such a modification. With my modification, all the cases work.

jsunintel:
@hassount @lijing0010 what's current status for this? :-)

hassount:
> @hassount @lijing0010 what's current status for this? :-)

There are a few items here which would need to be resolved here:
- I have just realized that this PR is to the rtime-m0-test, we usually only make PRs to the master. This will need to change to go to the master now that the master is up to date with the rtime branch and has surpassed it when it comes to performance. 
- A few conversations would need to be resolved. Some with Feedback from @anaghdin @chkngit to close on

Ultimately, this PR would need to be reviewed again once made to the master as it is now quite different than the rtime branch. So I suggest we close this PR and open another one that's based on master while addressing all of the outstanding discussion items

hassount:
@jsunintel 
any chance we can get this rebased this week ?

PR #753, opened 2019-10-30
Document TargetBitRate

dprestegard:
The default is crazy high!


hassount:
> The default is crazy high!

would you like to suggest a different value ? might need to be done per resolution because 7Mbps is not too high for 4k

1480c1:
> > The default is crazy high!
> 
> would you like to suggest a different value ? might need to be done per resolution because 7Mbps is not too high for 4k

Perhaps he thinks it's 700 mbps

dprestegard:
Indeed, I did think it's 700 Mbps. Looks like I need to update the parameter description then.

dprestegard:
7 Mbps is a reasonable default.

hassount:
> Indeed, I did think it's 700 Mbps. Looks like I need to update the parameter description then.

Bringing a good point here though, too many zeros, we might need to go to kbps instead of bps

dprestegard:
I do personally prefer Kbps, it improves readability and is what x264 and x265 CLIs do :)

dprestegard:
I'm a git newbie unfortunately. Sorry for the mess.

PR #763, opened 2019-11-01
Fix the r2r which is caused by tx search in encdec

lijing0010:


lijing0010:
it will cause r2r in M5

JackZhouVCD:
I just checked through all of the clips from the objective-fast-1 test set running M5 and the existing r2r issue is addressed by this PR. Thanks for adding this fix!

lijing0010:
add macro protection and rebase to latest

joelsole:
@lijing0010 could you squash the two commits and we can push this fix to master. Thanks.

lijing0010:
@joelsole , rebased to the latest and squash into one commit

PR #764, opened 2019-11-01
Fix temporal filtering for non-key Intra frames

anaghdin:
Description:
Fix a bug related to temporal filtering of non-key Intra frames. Temporal Filtering now enable for all Intra frame

Authors
@anaghdin

Type of change
Bug Fix

Tests and performance
Expected Average Y-PSNR BD-rate gain in the range of -3% on the jctvc list with Intra refresh enabled, using 4 QP values: {20, 32, 43 and 55}.


EwoutH:
That's a huge improvement for a bug fix, good find!

Is there a reason this is behind a switch though? Shouldn't it just be always enabled if it improved BDR without speed impact?

anaghdin:
@EwoutH Thanks. We usually keep the new features under a Macro, which is ON by default, to make sure there is no conflict/issue with other features. This approach helps us debugging especially when there are many features being developed in parallel. After a while, we do a round of clean up and remove the macros. Obviously, it has its pros and cons.

EwoutH:
Sounds logical, thanks

Issue #734, opened 2019-10-23
EncMode 3 and 4 BDrate regression


EwoutH:
Between 1f82662 and 0175b58 I noticed a BDrate regressiom in EncMode 4 and 3. EncMode 5 to 8 weren't impacted as far as I can see, I didn't test mode 2 to 0 (due to compute time limitations) so they could also be impacted.

See these runs on AWCY: [EncMode 3](https://beta.arewecompressedyet.com/?job=SVT-AV1-enc-mode-3%402019-10-06&job=SVT-AV1-enc-mode-3%402019-10-22) and [EncMode 4](https://beta.arewecompressedyet.com/?job=SVT-AV1-enc-mode-4%402019-10-06&job=SVT-AV1-enc-mode-4%402019-10-22).

Here the BDrate change between aad466d and 0175b58
![Screenshot_470](https://user-images.githubusercontent.com/15776622/67384492-8875df00-f591-11e9-99df-94dfba6cc726.png)

EwoutH:
Another serious regression in EncMode3 between 0175b58 and 7844f0b: https://beta.arewecompressedyet.com/?job=SVT-AV1-enc-mode-3%402019-10-22&job=SVT-AV1-enc-mode-3%402019-10-25

EwoutH:
@hassount Could you give a status update on this issue?

EwoutH:
@hguermaz @lijing0010 Could you detail in what portion #768 fixes this regression?

EwoutH:
Since 2019-10-06 EncMode 3 [regressed](https://beta.arewecompressedyet.com/?job=SVT-AV1-enc-mode-3%402019-10-06&job=SVT-AV1-enc-mode-3%402019-11-19) only more. The [largest regression](https://beta.arewecompressedyet.com/?job=SVT-AV1-enc-mode-3%402019-10-22&job=SVT-AV1-enc-mode-3%402019-10-25) seems between 0175b58 and 7844f0b, with BDR increases between 15% and 20%. Between d8e468f and 3ea18c2 EncMode 3 performance also [regressed](https://beta.arewecompressedyet.com/?job=SVT-AV1-enc-mode-3%402019-11-05&job=SVT-AV1-enc-mode-3%402019-11-19) 2-3% depending on metrics.

[...]

PR #768, opened 2019-11-04
Fix the corruption in M3(resolution like 1080P)

lijing0010:
In M3, sb size may be 128x128 for resolution >= 1080i.
while init_sq_non4_block() is only designed for sb size of 64x64

### DIFF
--- a/Source/Lib/Common/Codec/EbPictureDecisionProcess.c
+++ b/Source/Lib/Common/Codec/EbPictureDecisionProcess.c
@@ -823,11 +823,16 @@ EbErrorType signal_derivation_multi_processes_oq(
         else if (picture_control_set_ptr->enc_mode <= ENC_M2)
             picture_control_set_ptr->pic_depth_mode = PIC_ALL_DEPTH_MODE;
 
+        //Jing: TODO:
+        //In M3, sb_sz may be 128x128, and init_sq_non4_block only works for 64x64 sb size
         else if (picture_control_set_ptr->enc_mode <= ENC_M3)
             if (picture_control_set_ptr->slice_type == I_SLICE)
                 picture_control_set_ptr->pic_depth_mode = PIC_ALL_C_DEPTH_MODE;
             else
-                picture_control_set_ptr->pic_depth_mode = PIC_SQ_NON4_DEPTH_MODE;
+                if (sequence_control_set_ptr->seq_header.sb_size == BLOCK_128X128)
+                    picture_control_set_ptr->pic_depth_mode = PIC_SQ_DEPTH_MODE;
+                else
+                    picture_control_set_ptr->pic_depth_mode = PIC_SQ_NON4_DEPTH_MODE;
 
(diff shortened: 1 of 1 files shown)


lijing0010:
may be able to fix the M3 part in #734

hguermaz:
When pic_depth_mode is set to PIC_SQ_NON4_DEPTH_MODE, 128x128 should not be tested @ MD. So now block injection does not seem to be controlled by pic_depth_mode only.

Found that after ADD_MDC_REFINEMENT_LOOP adoption, pic_depth_mode controls the injection for only I_SLICE or when mdc_depth_level is set to MAX_MDC_LEVEL, but otherwise open_loop_partitioning_pass() controls the injection, and open_loop_partitioning_pass() might inject 128x128.

So a better fix for now will be to modify all_cu_init derivation; EbBool all_cu_init = (picture_control_set_ptr->parent_pcs_ptr->pic_depth_mode <= PIC_SQ_DEPTH_MODE **||  mdc_depth_level < MAX_MDC_LEVEL**).

However we might need to add a pic_depth_mode mode that controls mdc_depth_level to have block injection controlled by 1 unique internal signal.

NaderMahdi:
Although I don't think this is the right way to fix the issue because you are changing the operating mode of the pic_depth_mode which can lead to a different behavior of the speed/bdrate for M3 preset. However, you can have it as a temporally fix until we open the gate for the generalized open_loop_partitioning, so called MDC, to be operational in M1-M8 encoder presets which will fix the problem since it handles all SB sizes and all partition sizes.

lijing0010:
@NaderMahdi you are right, changing the pic_depth_mode will affect MDC and the overall speed/bdrate. @hguermaz , I'm a bit confused. 
For M3 1080P case, the super block size will be set to 128x128, and for non-I slice, pic_depth_mode will be set to PIC_SQ_NON4_DEPTH_MODE.
And in MDC_kernel, forward_sq_non4_blocks_to_md will also generate 128x128 block candidate.
So what is the expected behavior? init_sq_nsq_block will init all 4421 blocks while init_sq_non4_block only 1101. It looks like init_sq_non4_block should be combined with sb_sz==64, not 128x128. why it's related to MDC block injection?

hguermaz:
@lijing0010 you are right. I actually checked the settings of another branch when I 1st replied to your PR, and not the master (my mistake). In fact MDC will be soon be used for all presets, but that's not yet the case in the master.

lijing0010:
@hguermaz OK, so we can leave the bug here for some time

lijing0010:
@hguermaz , I may misunderstand your comments. 
Do you mean you will have another PR which can resolve this problem or not?

hguermaz:
@lijing0010 I do not have PR that solves this problem, however @hassount  might have 1 (i.e. M1 beyond settings PR). But since the M1 & beyond settings PR will not be ready soon, then I think you can go ahead and merge this PR.

PR #769, opened 2019-11-04
Interpolation filter search for 10-bit (#769)

joelsole:
## Description

Add interpolation filter search for 10-bit
* Clean-up of some of the related function calls
* Consolidate the SSE computation

Associated to issue #622

## Type of change

Added feature

## Tests and performance

BD-rate gains for 7 360p 10-bit test sequences is -0.6%  (enc-mode 0); BD-VMAF -1.06%.

Tiny change for 8-bit sequences due to the reuse of existing SSE computation (-0.001%). Encoding time slightly reduced -0.3% (enc-mode 0), though not reliable.

PR #770, opened 2019-11-04
hbd clean up

okhlif:
### Description
Hbd clean up to support three mode for 10bit samples
   - EB_8_BIT_MD    = 8bit mode decision
   - EB_10_BIT_MD   = 10bit mode decision **(default)**
   - EB_DUAL_BIT_MD = Auto: 8bit & 10bit mode decision

Add 8bit IFS path in 10bit MD
### Author
@okhlif 
### Type of change
Cleanup
Enhancement
### Tests and performance
EB_10_BIT_MD  & EB_8_BIT_MD unchanged
Add IFS 8 bit under EB_DUAL_BIT_MD
~ 2.8% speed gain for 10bit 360p
~ BDR over 7 clips 360p-10bit -0.11% (PSNR) -0.04% (SSIM)
~ Memory increase ~1%

### DIFF
--- a/Source/Lib/Common/Codec/EbEncDecProcess.c
+++ b/Source/Lib/Common/Codec/EbEncDecProcess.c
@@ -101,7 +101,7 @@ EbErrorType enc_dec_context_ctor(
 #endif
     EbBool                  is16bit,
     EbColorFormat           color_format,
-    EbBool                  enable_hbd_mode_decision,
+    uint8_t                 enable_hbd_mode_decision,
     uint32_t                max_input_luma_width,
     uint32_t                max_input_luma_height)
 {
@@ -278,7 +278,7 @@ static void ResetEncDec(
         &context_ptr->full_chroma_lambda,
         (uint8_t)picture_control_set_ptr->parent_pcs_ptr->enhanced_picture_ptr->bit_depth,
         context_ptr->qp_index,
-        picture_control_set_ptr->hbd_mode_decision);
+        context_ptr->md_context->hbd_mode_decision);
     // Reset MD rate Estimation table to initial values by copying from md_rate_estimation_array
     if (context_ptr->is_md_rate_estimation_ptr_owner) {
         EB_FREE(context_ptr->md_rate_estimation_ptr);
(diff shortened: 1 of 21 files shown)


okhlif:
@joelsole I made changes to support 3 modes. Please let me know your feedback
     * 0 = 8bit mode decision
     * 1 = 10bit mode decision
     * 2 = Auto: 8bit & 10bit mode decision

joelsole:
Interesting -  I wonder about the benefit of introducing IFS_8BIT_MD because the reference frame might be the 8-bit or the 10-bit version depending on the part of MD being executed: what are the pros and cons?

There will be some conflicts with #769. We'll solve those when we get there.

okhlif:
> Interesting - I wonder about the benefit of introducing IFS_8BIT_MD because the reference frame might be the 8-bit or the 10-bit version depending on the part of MD being executed: what are the pros and cons?
> 
> There will be some conflicts with #769. We'll solve those when we get there.

We always have the reference frame and input frame in both 8-bit and 10-bit. For example IFS, even the whole MD path is 10-bit, IFS can still reference 8-bit and use 8-bit input. I need to update lambda update in this case ( not yet done) 
This can bring better BDRare/speed trad-off especially the memory overhead is ~1%

kylophone:
> @joelsole I made changes to support 3 modes. Please let me know your feedback
> * 0 = 8bit mode decision
> * 1 = 10bit mode decision
> * 2 = Auto: 8bit & 10bit mode decision

Maybe use an `enum` for these.

okhlif:
@joelsole @chkngit  .Can you please review the changes

Source/Lib/Common/Codec/EbEncDecProcess.c
```diff
@@ -1684,7 +1684,9 @@ void* enc_dec_kernel(void *input_ptr)
             segmentRowIndex = segment_index / segments_ptr->segment_band_count;
             segmentBandIndex = segment_index - segmentRowIndex * segments_ptr->segment_band_count;
             segmentBandSize = (segments_ptr->lcu_band_count * (segmentBandIndex + 1) + segments_ptr->segment_band_count - 1) / segments_ptr->segment_band_count;
-
+//#if HBD_CLEAN_UP // context_ptr
```
hassount: still needed ?

hassount:
@okhlif 
This LGTM except for comment that might need to be removed.
This PR is up next as soon as it's ready

Source/Lib/Common/Codec/EbEncDecProcess.c
okhlif: Removed thank you

okhlif:
> @okhlif
> This LGTM except for comment that might need to be removed.
> This PR is up next as soon as it's ready

Thank you, comment addressed. Ready to go form my end.

PR #772, opened 2019-11-06
Fix the r2r case

lijing0010:
cand_buff_indices[CAND_CLASS_0] may not be initialized in certain case


JackZhouVCD:
@lijing0010 thanks for adding this fix! Do you know which cases of r2r you are targeting with this fix?

lijing0010:
@JackZhouVCD You can reproduce it in M2 with clip "vidyo1"

JackZhouVCD:
> @JackZhouVCD You can reproduce it in M2 with clip "vidyo1"

Thanks! I tested your changes and found that there are no more r2r's seen when encoding the clip in M2

hguermaz:
This piece of code will be very soon removed from the code, and replaced by a mechanism that checks all CLASSES @ the output of md-stage-0. Therefore, to avoid conflicts and save time, there is no need to merge this PR now.

lijing0010:
close this one since it will be useless

hguermaz:
Useless after PR 783 (already merged)

PR #781, opened 2019-11-11
Fix issue in unit test: ResidualSumTest.MatchTest/4,13 and 14

wallerjiaotencent:
Fix issues:
1. The data of this ResidualSum avx2 funtion is out of bounds in avx2 register when size = 8 ;
2. The ResidualSum avx2 function has no implementation code when size = 64;


wallerjiaotencent:
BTW: I have not found anywhere calling this function. Why?

hassount:
@wallerjiaotencent could it be used only in the decoder ?

PR #786, opened 2019-11-13
Implementation missing C kernels and fix UT AVCStyleMcpTest

spawlows:
Implementation missing C kernels: 
compute_mean_squared_values_c,
compute_mean_c,
avc_style_luma_interpolation_filter 

Fix UT: AVCStyleMcpTest*


Source/Lib/Common/ASM_SSE2/EbAvcStyleMcp_Intrinsic_SSE2.c
```diff
@@ -43,164 +43,62 @@ void picture_average_kernel_sse2_intrin(
...
-
-                _mm_storeu_si128((__m128i*) dst, xmm_avg1);
-                _mm_storel_epi64((__m128i*) (dst + 16), xmm_avg2);
-                _mm_storeu_si128((__m128i*) (dst + dst_stride), xmm_avg3);
-                _mm_storel_epi64((__m128i*) (dst + dst_stride + 16), xmm_avg4);
-
-                src0 += src0_stride << 1;
-                src1 += src1_stride << 1;
-                dst += dst_stride << 1;
+    __m128i xmm_avg1, xmm_avg2;
```
ttrigui: what is the reason behind changing this code? what happens when area_width is 24?

spawlows: The reason of this changes is that picture_average_kernel_sse2_intrin is used in avc_style_luma_interpolation_filter_pos[?]_ssse3() with not supported area_width like  88, 120, 136, 184.... All width sizes biggest that 64 was be calculated like 64. This the reason why UT failed. When area_width is 24, then width 16 (or multiple of 16) will be calculated by loop
```
for (y = 0; y + 15 < area_width; y += 16) {
```
and then rest of width 8 will be calculated inside
```
if (area_width & 8) {
```
and if some width equal 4 left (for example for 28 = 16+8 +4) will be calculated in:
```
if (area_width & 4) {
```

spawlows:
@hassount @ttrigui @magwyz 

Hi, I need to revert changes from PR #827 which remove "dead code", to resolve conflicts. I need this code, to implement C version of avc_style_luma_interpolation_filter_helper_ssse3.
At this moment master uses only ssse3 kernel, even if you force C optimizations. 

As an option, I can move files EbAvcStyleMcp.h and EbAvcStyleMcp.c from COMMON_CODEC to COMMON_C_DEFAULT.

magwyz:
@spawlows Sure, feel free to revert if this code is actually needed.

Source/Lib/Common/ASM_SSE2/EbAvcStyleMcp_Intrinsic_SSE2.c
ttrigui: ok I see thx

hassount:
@andrey-norkin @joelsole 
can we get this PR reviewed. 
@spawlows can you please rebase and squash the commits into 1

spawlows:
> @spawlows can you please rebase and squash the commits into 1

Done.

PR #792, opened 2019-11-15
Fix : decoder 12-bit output mismatch

nijilk:
Fixed decoder mismatch for 12-bit streams.
The buggy intrinsic module is replaced by C variant for 12 bit streams.


nijilk:
@joelsole @hassount 
Please review

joelsole:
All 45 12-bit dav1d streams are passing now.

PR #795, opened 2019-11-15
eb_deinit_handle: free lp_group, before eb_decrease_component_count

xuguangxin:
else it will report memory leak

### DIFF
--- a/Source/Lib/Encoder/Codec/EbEncHandle.c
+++ b/Source/Lib/Encoder/Codec/EbEncHandle.c
@@ -1846,16 +1846,13 @@ EB_API EbErrorType eb_deinit_handle(
         return_error = eb_av1_enc_component_de_init(svt_enc_component);
 
         free(svt_enc_component);
+#if  defined(__linux__)
+        EB_FREE(lp_group);
+#endif
         eb_decrease_component_count();
     }
     else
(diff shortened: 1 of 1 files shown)


kylophone:
Looks good. Thanks!

luigi311:
I noticed you removed the if(lp_group != NULL) line, did you test multi encode instances? That was added in svt-hevc to prevent memory leak/crashing when running multiple encoder instances.
https://github.com/OpenVisualCloud/SVT-HEVC/commit/4f4fd5915fc03bb5fed490233638abc85664dcba

xuguangxin:
Hi @luigi311 ,
Thanks for the review. 

free will do null check for us, it's no need double-check.

> The free() function frees the memory space pointed to by ptr, which must have been returned by a previous call to malloc(), calloc() or realloc(). Otherwise, or if free(ptr) has already been called before, undefined behavior occurs. **If ptr is NULL, no operation is performed.**

and EB_FREE will set the pointer to null
https://github.com/OpenVisualCloud/SVT-AV1/blob/3b4e627d4294511f903b068e1b063a1b042a3073/Source/Lib/Common/Codec/EbMalloc.h#L82-L87

so before and after the patch, the behavior is similar.

PR #796, opened 2019-11-15
Fix issue in unit test: DeblockTest

spawlows:
Code SSE2 use in different way array *_blimit1 in relation to C, and SSE2 calculated 2 times more elements that they should.


ttrigui:
> _No description provided._

Please add a description.

PR #802, opened 2019-11-18
Fix mismatch between C and AVX2 kernel get_eight_horizontal_search_po…

spawlows:
Fix mismatch output between C and AVX2 encoding

AVX2 kernel did not respect when value xMv overflowed and reduced  to type uint16 

```
 int16_t xMv, yMv;
...
  p_best_sad32x32[3] = sad32x32_3;
            xMv = _MVXT(mv) + (int16_t)xSearchIndex * 4;
            yMv = _MVYT(mv);
            p_best_mv32x32[3] = ((uint16_t)yMv << 16) | ((uint16_t)xMv);
```


ttrigui:
> Fix mismatch output between C and AVX2 encoding

@spawlows please add a little description about the problem

Issue #302, opened 2019-05-29
Implement library logging function, svt_log()


kylophone:
We need a proper log function complete with log levels. Also, the lib should log to stderr by default (not stdout!) since we may want to have piped output. All calls to `printf()`and `fprintf()` should be replaced with this logging function. This is a relatively easy task, see `av_log()` in ffmpeg for inspiration.

1480c1:
I was considering to do this to address #231, at least the log to stderr

joelsole:
That would be great, thanks. 

Harmonization is needed, as there are several definitions with printf: SVT_LOG, H0, PRINT

Issue #701, opened 2019-10-15
Support output to stdout for improved pipe support {feature request}


grill05:
Currently, when using ffmpeg with SvtAv1EncApp with pipes, one does

`ffmpeg -i input.mkv -an -s 424x240 -f yuv4mpegpipe -pix_fmt yuv420p  - | SvtAv1EncApp -i stdin -n 10 -w 424 -h 240  -fps 24 -rc 3 -tbr 75000 -enc-mode 5  test.webm`

for producing output to a file test.webm
The problem here is that one cannot encode audio simultaneously. If Svt could output to stdout, one could do 

`ffmpeg -i input.mkv -an -s 424x240 -f yuv4mpegpipe -pix_fmt yuv420p  - | SvtAv1EncApp -i stdin -n 10 -w 424 -h 240  -fps 24 -rc 3 -tbr 75000 -enc-mode 5  - | ffmpeg -i - -i input.mkv -map 0:v -c:v copy -map 1:a -c:a libopus output.webm`

and in this case, output.webm would have both audio and video.
I know that there is a ffmpeg plugin (libsvt), but I cant get that to compile, so I have to use SvtAv1EncApp via pipes. 

It would be nice if svt could output to stdout (like rav1e and aomenc can)

1480c1:
Related to https://github.com/OpenVisualCloud/SVT-AV1/issues/302

Duplicate of https://github.com/OpenVisualCloud/SVT-AV1/issues/231

EwoutH:
@hassount Could we get this in for the [0.7.5 release](https://github.com/OpenVisualCloud/SVT-AV1/milestone/1)? Would be a great addition to enlarge the user base (and thus testing and bug reports).

Also a useful step for #259.

hassount:
@EwoutH getting it in 0.7.5 is probably not going to be possible as we won't have resources to support this feature. Though, this is high level enough to be open for contributions.

j-mas:
@hassount@EwoutH
I think this is the most important feature for us users in the nearest future!

grill05:
From a user perspective, I'd say 2-pass vbr mode is absolutely essential, while crf-mode and output to stdout would be very nice to have. 
That is because the encoder output looks very bad (visually) for high motion scenes right now (at low-ish bit rates). Whenever there is high motion/scene-change the encoder produces very visible pixellation (at low bit rates). Hopefully 2-pass vbr should help with that. 
[...]

PR #809, opened 2019-11-20
add svt log system

xuguangxin:
this will remove all printf to stdout, fixes https://github.com/OpenVisualCloud/SVT-AV1/issues/302

as applciation user,you have three options in env variables
SVT_LOG: control the log level
SVT_LOG_FILE: log file for svt library
SVT_APP_LOG_FILE: log file for svt applications
here is sample commandline:
```
1. SVT_LOG=4 SVT_LOG_FILE=lib.log SVT_APP_LOG_FILE=app.log SvtEncApp -i test.y4m -b test.ivf
2. SVT_LOG=4 SVT_LOG_FILE=lib.log SVT_APP_LOG_FILE=app.log ffmpeg -i test.y4m -c:v libsvt_av1 test.mp4
```

as developer, you can use following macros
```
SVT_LOG
SVT_DEBUG
SVT_INFO
SVT_WARN
SVT_ERROR
SVT_FATAL
```
please check EbMalloc.c for for usage

### DIFF
--- a/Source/App/DecApp/EbDecAppMain.c
+++ b/Source/App/DecApp/EbDecAppMain.c
@@ -46,7 +46,7 @@ int init_pic_buffer(EbSvtIOFormat *pic_buffer, CLInput *cli,
         pic_buffer->y_stride = cli->width;
         break;
     default:
-        printf("Unsupported colour format. \n");
+        fprintf(stderr, "Unsupported colour format. \n");
         return 0;
     }
     pic_buffer->width = cli->width;
@@ -74,7 +74,7 @@ int read_input_frame(DecInputContext *input, uint8_t **buffer, size_t *bytes_rea
             buffer_size);
         break;
     default:
-        printf("Unsupported bitstream type. \n");
+        fprintf(stderr, "Unsupported bitstream type. \n");
         return 0;
     }
 }
@@ -125,7 +125,7 @@ void write_frame(EbBufferHeaderType *recon_buffer, CLInput *cli) {
 }
 
 static void show_progress(int in_frame, uint64_t dx_time) {
-    printf("\n%d frames decoded in %" PRId64 " us (%.2f fps)\r",
+    fprintf(stderr, "\n%d frames decoded in %" PRId64 " us (%.2f fps)\r",
         in_frame, dx_time,
         (double)in_frame * 1000000.0 / (double)dx_time);
 }
@@ -170,8 +170,8 @@ int32_t main(int32_t argc, char* argv[])
     size_t bytes_in_buffer = 0, buffer_size = 0;
 
     // Print Decoder Info
-    printf("-------------------------------------\n");
-    printf("SVT-AV1 Decoder\n");
+    fprintf(stderr, "-------------------------------------\n");
+    fprintf(stderr, "SVT-AV1 Decoder\n");
 
     // Initialize config
     if (!config_ptr)
@@ -210,7 +210,7 @@ int32_t main(int32_t argc, char* argv[])
         ((EbSvtIOFormat *)recon_buffer->p_buffer)->cb = (uint8_t*)malloc(size >> 2);
         ((EbSvtIOFormat *)recon_buffer->p_buffer)->cr = (uint8_t*)malloc(size >> 2);
         if (!init_pic_buffer((EbSvtIOFormat*)recon_buffer->p_buffer, &cli, config_ptr)) {
-            printf("Decoding \n");
+            fprintf(stderr, "Decoding \n");
             EbAV1StreamInfo *stream_info = (EbAV1StreamInfo*)malloc(sizeof(EbAV1StreamInfo));
             EbAV1FrameInfo *frame_info = (EbAV1FrameInfo*)malloc(sizeof(EbAV1FrameInfo));
 
@@ -252,7 +252,7 @@ int32_t main(int32_t argc, char* argv[])
             }
             if (fps_summary || fps_frm) {
                 show_progress(in_frame, dx_time);
-                printf("\n");
+                fprintf(stderr, "\n");
             }
 
             if (enable_md5) {
@@ -273,7 +273,7 @@ int32_t main(int32_t argc, char* argv[])
         free(buf);
     }
     else
-        printf("Error in configuration. \n");
+        fprintf(stderr, "Error in configuration. \n");
     return_error |= eb_dec_deinit_handle(p_handle);
 
 fail:
--- a/Source/App/DecApp/EbDecParamParser.c
+++ b/Source/App/DecApp/EbDecParamParser.c
@@ -32,7 +32,7 @@ static void set_num_thread(const char *value, EbSvtAv1DecConfiguration *cfg) { c
 static void set_num_pframes(const char *value, EbSvtAv1DecConfiguration *cfg) {
     cfg->num_p_frames = strtoul(value, NULL, 0);
     if (cfg->num_p_frames != 1) {
-        printf("Warning : Multi frame parallelism not supported. Setting parallel frames to 1. \n");
+        fprintf(stderr, "Warning : Multi frame parallelism not supported. Setting parallel frames to 1. \n");
         cfg->num_p_frames = 1;
     }
 };
@@ -98,7 +98,7 @@ EbErrorType read_command_line(int32_t argc, char *const argv[],
                 config_strings[cmd_token_cnt] = argv[++token_index];
         }
         else {
-            printf(" Invalid CLI: %s \n", argv[token_index]);
+            fprintf(stderr, " Invalid CLI: %s \n", argv[token_index]);
             return EB_ErrorBadParameter;
         }
     }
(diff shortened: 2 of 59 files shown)


1480c1:
Why use environment variables instead of setting the variables via command line args or others?

1480c1:
> this will remove all printf to stdout, fixes
> #302

Also change this to one line like
```
this will remove all printf to stdout, fixes #302
```

That way you can auto close the issue
https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

xuguangxin:
@1480c1 ,  
thanks for the review. 

> That way you can auto close the issue

fixed

> Why use environment variables instead of setting the variables via command line args or others?

It's easy to add a debug command line for SvtEncApp. But if we want integrate it to frameworks like ffmpeg or gstreamer, it's hard to add svt specific debug options.

thanks

kylophone:
Thanks!

A few nits:
- You wrote an duplicate loggers, an app logger and a lib logger. Just the lib logger is sufficient.
- Like others have said, setting this with an environment variable is not great.
- The macros hiding the actual `svt_log()` function are unnecessary IMHO.

kylophone:
Please also add a `SVT_LOG_QUIET` to turn logging off altogether.

xuguangxin:
Thanks for review.
> Thanks!
> 
> A few nits:
> 
> * You wrote duplicate loggers, an app logger and a lib logger. Just the lib logger is sufficient.


In this case, how we handle app-level printf?
> * Like others have said, setting this with an environment variable is not great.


Could you explain more about the drawback of an environment variable? If we change to comandline params, we may need to add a special debug command line to ffmpeg/gstremaer plugins. It may not acceptable to ffmpeg/gstreamer upstream. Also, we need to expose set_svt_log_level and set_svt_log_file things as public API. 


> * The macros hiding the actual `svt_log()` function are unnecessary IMHO.

It has the following advantages:

- It will let the caller focus on the output content, they do not need care about log level and tag params
- It's more extendable. If needed, we may change the tag from string to structure, so we can support module based debug level like [GST_DEBUG](https://gstreamer.freedesktop.org/documentation/tutorials/basic/debugging-tools.html?gi-language=c)
- If we want to disable all log functions in compile-time, call svt_log directly will leave an empty function call.  But if we use macros, you can simply define macros to empty, it's no runtime overhead.

xuguangxin:
> Please also add a `SVT_LOG_QUIET` to turn logging off altogether.

thanks for the suggestion.
this is a compile-time flag, right. do we need to add some cmake option to enable it?

xuguangxin:
@kylophone , @1480c1 
is https://github.com/OpenVisualCloud/SVT-AV1/pull/809#issuecomment-557368903 addressed all your concerns?

@kylophone SVT_LOG_QUIET added please help review it's you wanted or not.
thanks

xuguangxin:
@kylophone , @1480c1
any feedback on this?
thanks

xuguangxin:
@1480c1 thanks for the review. 
> something like -logfile SVT_AV1_$(date +%H%M%S).log and -loglevel quiet and -loglevel warn

we can do this for env too, just use

```
SVT_LOG_FILE=SVT_AV1_$(date +%H%M%S).log SVT_LOG=0 SvtEncApp xxxx
or
SVT_LOG_FILE=SVT_AV1_$(date +%H%M%S).log SVT_LOG=0 ffmpeg xxxx
```

and the env version is a unified approach across all frameworks.


> Could we not just add the lib logger header into the Source/API folder somewhere and link the app printfs to the lib logger? Since svt_log is already non-static so we need to expose set_svt_log_level and set_svt_log_file things as public API

If you and @kylophone preferred to use the command line, I have no objection. But we need @hassount @kylophone agrees to add svt_av1_log svt_av1_set_log _file and svt_av1_set_level as public API.

> Suggestion: change svt_log to svt_av1_log to prevent conflicts with the other svt encoders if they also add a log system similar to this one

sure I will do this.

1480c1:
> we can do this for env too, just use
> 
> ```
> SVT_LOG_FILE=SVT_AV1_$(date +%H%M%S).log SVT_LOG=0 SvtEncApp xxxx
> or
> SVT_LOG_FILE=SVT_AV1_$(date +%H%M%S).log SVT_LOG=0 ffmpeg xxxx
> ```
> 
> and the env version is a unified approach across all frameworks.

To preface anything else, in my personal opinion, any programs that requires the user to either read or remember and then use any environment variable past a certain length, with the exception of build systems, compilers, linkers, and certain others, is bad design.

For some, like gstreamer, I do understand using using a `GST_DEBUG` variables since they have multiple command line applications and multiple interfaces to their library and facilitate communication between multiple other libraries and it would be hard to set a common flag. 

However, in the case of SVT-AV1, we neither support multiple command line applications nor application interfaces such as python etc and are primarily a library before a complete application, and introducing a usage of an environment variable not commonly needed will only introduce clutter for the users and those wanting to develop SVT-AV1 with other tools

> > Could we not just add the lib logger header into the Source/API folder somewhere and link the app printfs to the lib logger? Since svt_log is already non-static so we need to expose set_svt_log_level and set_svt_log_file things as public API
> 
> If you and @kylophone preferred to use the command line, I have no objection. But we need @hassount @kylophone agrees to add svt_av1_log svt_av1_set_log _file and svt_av1_set_level as public API.

@hassount, should we put the logger header in the API folder and export it like others for the app to link to, or do something similar to the string functions and keep a duplicate in the app folder and common/codec folder?

hassount:
@1480c1 
My personal opinion is that an API call for the library should be restricted to the in / out buffers and configuration of the library. anything else should live independently in the application. @kylophone any thoughts here ?

kylophone:
> @kylophone , @1480c1
> any feedback on this?
> thanks

Thanks for the ping. Been away for a few days...

- Please do not set the `log_level` with an environment variable.

- For the app, either use `fprintf(stderr, ...)` or add your lib logger to the public API. 

And to address the following, see the x264 encoder in libavcodec. This is a pretty common pattern.

> > Why use environment variables instead of setting the variables via command line args or others?
> 
> It's easy to add a debug command line for SvtEncApp. But if we want integrate it to frameworks like ffmpeg or gstreamer, it's hard to add svt specific debug options.

https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/libx264.c#L110-L123

Let me know if there are any questions.

xuguangxin:
@kylophone thanks for replay. 

>And to address the following, see the x264 encoder in libavcodec. This is a pretty common pattern.
this needs a public API for the log call back. 
https://github.com/mstorsjo/x264/blob/master/x264.h#L348
Do you think it's ok to add 
`svt_av1_set_log(log_call_back, default_level) `
to public API.

hassount:
if the log API is the controversial one, can we keep it as an #define for now so that we can get this PR in before the code freeze ? @kylophone would this be ok ?

xuguangxin:
@hassount thanks for the reminder, do not worry, this pr is easy to rebase. 
I am more worry about https://github.com/OpenVisualCloud/SVT-AV1/pull/825.
could you help review it and make sure it merged before code freeze?

thanks

xuguangxin:
hmm, add a log callback is not so easy task. We can't use a global variable for this. It will have a problem when we have multi instances.
 If we want to add callback, we need to add log related things in EbSvtAv1EncConfiguration, and pass the call back to function who need the log.

for example, everyone calls x264_log. need a per instance [pf_log](https://github.com/ShiftMediaProject/x264/blob/79de3131a5f72e3ab8c17c1d03f653ed24f06c5d/common/common.c#L41)
If we use this pattern, we need to change signature for functions like https://github.com/OpenVisualCloud/SVT-AV1/blob/48142688b1db2492660652a5d1bd1758ba95c51a/Source/Lib/Common/Codec/EbModeDecision.c#L572
to pass per instance callback to them. It's a huge task and it will change many lines of code.

hassount:
@xuguangxin 
can we stay out of the call back situation and create a #define inside the code until we figure out the best way of proceeding ? 

I'd like the get the rest of this PR in.

xuguangxin:
@hassount , you mean the ifdef to disable all log outs?
yes, we have this https://github.com/OpenVisualCloud/SVT-AV1/pull/809/files#diff-a9e517f4f9cdb52366bf759a23564948R23

hassount:
ok, and can we use the fprintf as @kylophone has suggested ?

xuguangxin:
ok, let me recapture this:
1. keep using log functions for the library.
2. use fprintf for applications.
Is it correct?

hassount:
> ok, let me recapture this:
> 
> 1. keep using log functions for the library.
> 2. use fprintf for applications.
>    Is it correct?

correct

xuguangxin:
@hassount rebased. please help review.
thanks

hassount:
> @hassount rebased. please help review.
> thanks

@xuguangxin thank you, so it seems that now we have the following at the app:
- Info level print outs
- Error level print outs

For the app only and now that we have the error output using the fprintf into the err stream, can we just use printf for the regular output and remove the common APP svt log as it seems as if it's not being used ?

1480c1:
Could we keep using fprintf to stderr so that way we can use stdout for piping SvtAv1EncApp's output into something else in the future?

xuguangxin:
@hassount , 
@1480c1 is right, we need to remove all printf to support pipe. see https://github.com/OpenVisualCloud/SVT-AV1/issues/701#issuecomment-543861389 for details
thanks

1480c1:
Nice

xuguangxin:
thanks all help on this

1480c1:
Thank you and have a happy holiday (If you celebrate one)

PR #810, opened 2019-11-20
EbPictureControlSet: fix rst_frame leak

xuguangxin:
steps to reproduce:
run Bin/Debug/SvtAv1EncApp -i test.y4m -enc-mode 7 -b test.ivf

you will see memory leak at EbPictureBufferDesc.c:L354


xuguangxin:
yes, you are right. 
commit log changed.

PR #813, opened 2019-11-21
Added check on input/output stat file 

ttrigui:
Added -pass to indicate first or second pass
Check on input/output stat file : output for first pass and output for second pass


hassount:
#746 is taking care of this

Issue #710, opened 2019-10-17
Using enc mode 5 or lower + -intra-period 31 or greater (24 fps source) results in stall at empty buffer allocate semaphore


bitbytebit-cr:
When encoding with mode 5 or below + an intra period of 31 frames or more (24 fps video) the SvtAv1EncApp sits without any CPU usage indefinitely stalled on the first/second frame (Processes 1 frame) here as captured with GDB...

```
[centos@ip-172-30-2-6 perceptual_hash_encoder]$ gdb
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-114.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
(gdb) file SvtAv1EncApp
Reading symbols from /usr/local/bin/SvtAv1EncApp...done.
(gdb) run -i input.yuv -rc 2 -tbr 3000000 -intra-period 48 -fps-num 24000 -fps-denom 1001 -lp 8 -w 1920 -h 1080 -n 4000 -b out.ivf -enc-mode 4
Starting program: /usr/local/bin/SvtAv1EncApp -i input.yuv -rc 2 -tbr 3000000 -intra-period 48 -fps-num 24000 -fps-denom 1001 -lp 8 -w 1920 -h 1080 -n 4000 -b out.ivf -enc-mode 4
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
-------------------------------------------
SVT-AV1 Encoder
SVT [version]:  SVT-AV1 Encoder Lib v0.7.0
SVT [build]  :  GCC 5.4.0        64 bit
LIB Build date: Oct 17 2019 17:21:39
-------------------------------------------
Number of logical cores available: 8
Number of PPCS 87
-------------------------------------------
SVT [config]: Main Profile      Tier (auto)     Level (auto)
SVT [config]: EncoderMode                                                       : 4
SVT [config]: EncoderBitDepth / EncoderColorFormat / CompressedTenBitFormat     : 8 / 1 / 0
[...]

PR #816, opened 2019-11-22
Turn OFF NON_KF_INTRA_TF_FIX pending fixing issue 710

ttrigui:
## Description
NON_KF_INTRA_TF_FIX causing a crash when the intra peirod is not aligned due to the non-availability of altref pictures in the picture decision queue.
Turning off the feature until the issue is addressed.

## Related Issue
#710


joelsole:
what is the reason to disable the fix?

ttrigui:
> what is the reason to disable the fix?

There is a crash triggered by this feature. The crash was reported in this issue #710

joelsole:
Thanks for the explanation.

PR #817, opened 2019-11-22
Travis: Fix python, add clang-10

### LINKED ISSUE #812: MacOS clang build is broken on Travis
I suspect the environment Travis is running in has changed.

Example: https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/258874055

```
The command "pip install --user cpp-coveralls" failed and exited with 127 during .
```

1480c1:
Not sure why valgrind job is still failing,

Fixes #812 
@kylophone @EwoutH


hassount:
> Not sure why valgrind job is still failing,

might be timing out because it's set to -enc-mode 2

1480c1:
> > Not sure why valgrind job is still failing,
> 
> might be timing out because it's set to -enc-mode 2

Should I set it to a higher number, or don't mess with it for now?

hassount:
> Should I set it to a higher number, or don't mess with it for now?

yeah let's set it to -enc-mode 4 maybe, that would be quite faster and we can check if that works. the way it is is kind of useless right now.

kylophone:
Can you rebase and squash? There's some stuff in the diff that I don't think you meant to change.

1480c1:
If you are talking about the `cleanup:` commits, they are there to fix the style check job since it hasn't actually been working

hassount:
@1480c1 
I think we should reduce the amount commits maybe 3 to 4 steps would be enough ?

hassount:
@kylophone 
apart from the commits, does everything look good here ?

.travis.yml
```diff
@@ -182,12 +196,31 @@ matrix:
...
+      stage: coveralls+valgrind
       env: COVERALLS_PARALLEL=true build_type=debug CMAKE_EFLAGS="-DCOVERAGE=ON"
       script:
         - *base_script
-        - *coveralls_script
-      after_script: *after_coveralls_script
+        - &coveralls_script |
+          pip3 install --user cpp-coveralls
+          export PATH=/Users/travis/Library/Python/3.7/bin:${PATH}
+          SvtAv1EncApp -enc-mode 8 -w 720 -h 486 -fps 60 -i akiyo_cif.y4m -n 150 -b test1.ivf
```
hassount: no need for width and height as it is a y4m

hassount:
One thing that I do not understand is that why would we still have trailing spaces when the ci checks all code for that :/

kylophone:
> @1480c1
> I think we should reduce the amount commits maybe 3 to 4 steps would be enough ?

Yeah, that sounds about right.

.travis.yml
1480c1: Removed unnecessary width and heights from SvtAv1EncApp commandlines

.travis.yml
```diff
@@ -96,21 +103,24 @@ matrix:
       before_install: skip
       install: skip
       script:
-        - git remote rm upstream 2> /dev/null || true
-        - git remote add upstream https://github.com/OpenVisualCloud/SVT-AV1.git
-        - git fetch -q upstream master
-        - git grep -InP --heading "\t" -- . ':!third_party/**/*' && false || true
-        - git grep -InP --heading "\r" -- . ':!third_party/**/*' && false || true
-        - git grep -InP --heading " $" -- . ':!third_party/**/*' && false || true
```
1480c1: > One thing that I do not understand is that why would we still have trailing spaces when the ci checks all code for that :/

Logical errors here where we usually do `! git grep -InP --heading " $" -- . ':!third_party/**/*'` but for yaml, `!` is a special thing so the `!` isn't put into bash. Solved by using `- |` to make the following parts literal

hassount:
@1480c1 
so ready to go ?

1480c1:
yep

PR #818, opened 2019-11-22
Improve High-Precision MV

anaghdin:
**Description:**
Improves the high precision MV performance by:
- Shutting it based on the qindex
- Rate estimation fixes
- Candidate injection fixes

**Authors**
@anaghdin

**Type of change**
Bug Fix, enhancement

**Tests and performance**
Expected Average PSNR-SSIM BD-rate gain in the range of -0.4% (8 bit) and -0.6%(10bit) over 360P clips in aom list, using 5 QP values: {20, 32, 43, 55 and 63}.


joelsole:
What is the performance for other resolutions and 10b? 
I guess there is no encoding time impact - can you confirm?

ultrawide:
Hi Joel, I found a gain of 0.61% for 10bit 360P clips in context of two pass M5M0. There is no impact for high resolution clips.

PR #832, opened 2019-11-26
Update cost calculation for cdef filtering

ultrawide:
## Description 
- Update rate estimation, cost calculation equation, and lambda to match lambda from MD

## Type of change

- Optimization / enhancement

## Performance 

Tested on Objective-fast1 test-set in context of M5M2 where there was a gain of 0.1%. Going to re-run a simulation over M0.

## Author
@NaderMahdi @anaghdin


ultrawide:
Please do not merge this pull request yet. There is a simulation running over M5M0 to verify the improvements over the aomlist.

ultrawide:
Using a difference of difference comparison, the M5M0 simulation yielded a 0.1% gain.

PR #840, opened 2019-11-28
Fix mismatch between C and AVX2 kernel get_eight_horizontal_search_po…

spawlows:
…int_results_32x32_64x64_pu_avx2_intrin

Different output when few best_mv64x64 have that same value.

For AVX2 version, kernel find best_mv64x64 in different order.
Not see any drop performance.


ttrigui:
@spawlows the speed evaluation was in context of the unit test?

spawlows:
Yes, in the context of the unit test, so i expected that encoder especially should not change performance.

PR #842, opened 2019-11-28
Code cleanup for cdef filtering

### LINKED ISSUE #833: Remove unused functions in EbCdef.c
There are functions in EbCdef.c which are not called within the solution, namely `void eb_av1_cdef_search()` and `void av1_cdef_search16bit()`. It would be good to reduce dead code which may cause potential misunderstanding and administrative overhead.

ultrawide:
This pull request aims to improve code quality related to cdef filtering and to refactor code so it adheres to STYLE.md. Feedback or suggestions are appreciated! 

- remove unused functions `void eb_av1_cdef_search()` and `void av1_cdef_search16bit()`
- remove unused macro _CDEF_BLOCK_H (1)
- remove commented out code

closes #833

### DIFF
--- a/Source/Lib/Common/Codec/EbCdef.c
+++ b/Source/Lib/Common/Codec/EbCdef.c
@@ -322,8 +322,8 @@ void eb_cdef_filter_fb(uint8_t *dst8, uint16_t *dst16, int32_t dstride, uint16_t
     }
     if (pli == 1 && xdec != ydec) {
         for (bi = 0; bi < cdef_count; bi++) {
-            /*static*/ const int32_t conv422[8] = { 7, 0, 2, 4, 5, 6, 6, 6 };
-            /*static*/ const int32_t conv440[8] = { 1, 2, 2, 2, 3, 4, 6, 0 };
+            const int32_t conv422[8] = { 7, 0, 2, 4, 5, 6, 6, 6 };
+            const int32_t conv440[8] = { 1, 2, 2, 2, 3, 4, 6, 0 };
             by = dlist[bi].by;
             bx = dlist[bi].bx;
             dir[by][bx] = (xdec ? conv422 : conv440)[dir[by][bx]];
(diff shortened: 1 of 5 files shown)


Source/Lib/Common/Codec/EbCdef.c
```diff
@@ -797,7 +781,7 @@ void av1_cdef_frame16bit(
     uint16_t*  reconBufferCb = (uint16_t*)recon_picture_ptr->buffer_cb + (recon_picture_ptr->origin_x / 2 + recon_picture_ptr->origin_y / 2 * recon_picture_ptr->stride_cb);
     uint16_t*  reconBufferCr = (uint16_t*)recon_picture_ptr->buffer_cr + (recon_picture_ptr->origin_x / 2 + recon_picture_ptr->origin_y / 2 * recon_picture_ptr->stride_cr);
 
-    const int32_t num_planes = 3;// av1_num_planes(cm);
+    const int32_t num_planes = 3;
```
mafonsoN: The number of planes is not always 3, in case of monochrome (which we might add support for soon). So it is better to use av1_num_planes() to get this value, which is already used in some parts of the code.

ultrawide: Thanks for the suggestion. To make the changes, we need to define `av1_num_planes()` in a common header that both the encoder and decoder share. At the moment `av1_num_planes()` is defined in 'EbDecInverseQuantize.h'.

mafonsoN: True. What about moving it to EbDefinitions.h?

ultrawide: For sure, moving it to EbDefinitions.h works. There is also a header called EbUtility.h. I think separating definitions and utility functions would be better going forward, but may be out of scope for this pull request.

hassount:
This might conflict with #864

ultrawide:
No problem!

hassount:
@ultrawide  can you please rebase and squash to not have non-building commits in the master?

PR #845, opened 2019-11-29
Single av1_make_masked_inter_predictor function for 8 and 10 bit streams

magwyz:
This removes av1_make_masked_inter_predictor_hbd and should have no impact
on the bit stream.

### DIFF
--- a/Source/Lib/Common/Codec/EbInterPrediction.c
+++ b/Source/Lib/Common/Codec/EbInterPrediction.c
@@ -1884,73 +1884,6 @@ void av1_make_masked_inter_predictor(
         blk_geom->bsize, bheight, bwidth, conv_params, bitdepth);
 
 }
-#if INTER_INTER_HBD
-void av1_make_masked_inter_predictor_hbd(
-    uint16_t                  *src_ptr,
-    uint32_t                   src_stride,
-    uint16_t                  *dst_ptr,
-    uint32_t                   dst_stride,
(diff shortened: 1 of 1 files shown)


okhlif:
lgtm

JackZhouVCD:
I was testing PR #853 and I found that the changes from this PR introduced another crash using the same inputs after their fix. 

Namely: applying #853 on e54f59e (the commit before this PR was merged) addressed the crash that it was trying to fix, but applying #853 with 25b73da (the commit of this PR) the bug is back again.

Command used to reproduce the crash:
```haskell
 ./SvtAv1EncApp -i speed_720p_10bit.yuv -w 424 -h 240 -bit-depth 10 -rc 2 -tbr 200000 -irefresh-type 2 -enc-mode 5 -fps 24.0
```
It seems that the stream used for encoding does not matter as long as it's 10bit and the encoding is configured for a 424x240 stream.

@magwyz can you please look into this?

magwyz:
@JackZhouVCD Sure.

magwyz:
@JackZhouVCD I have just tested #853 on top of master and it does not crash for me.
Can you reproduce the crash?
This PR only modified EbInterPrediction.c and the stack trace of #853 does not mention this file. The crash is in intra prediction.

JackZhouVCD:
@magwyz I see, I tried again and I wasn't able to reproduce the crash either; I must have had an issue with my setup yesterday, false alarm. Sorry about that!

Issue #708, opened 2019-10-17
VBV Max Bitrate and Buffer size {enhancement}


n808:
Max Bitrate is defined in the specification's Levels table. However, if I wanted to change this to a lower number, I have so far been unable to find where to modify the source code to influence max bitrate. Any suggestions? 

With regards to the VBV max buffer size, it appears possible to modify the code as suggested below to achieve this.

```diff
diff --git a/Source/API/EbSvtAv1Enc.h b/Source/API/EbSvtAv1Enc.h
index ab32bed..5c8d38a 100644
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -276,6 +276,13 @@ typedef struct EbSvtAv1EncConfiguration
      *
      * Default is 7000000. */
     uint32_t                 target_bit_rate;
+
+    /* VBV Buffer size */
+    uint32_t                 vbv_bufsize;
+
+    /* VBV Max Rate */
+    uint32_t                 vbv_maxrate;
+    
     /* Maxium QP value allowed for rate control use, only applicable when rate
      * control mode is set to 1. It has to be greater or equal to minQpAllowed.
      *
diff --git a/Source/App/EncApp/EbAppConfig.c b/Source/App/EncApp/EbAppConfig.c
index 99520ae..57b7cdc 100644
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -99,6 +99,8 @@
 #define TILE_COL_TOKEN                   "-tile-columns"
 
 #define SCENE_CHANGE_DETECTION_TOKEN    "-scd"
+#define VBV_MAXRATE                     "-vbv-maxrate"
+#define VBV_BUFSIZE                     "-vbv-bufsize"
 #define INJECTOR_TOKEN                  "-inj"  // no Eval
 #define INJECTOR_FRAMERATE_TOKEN        "-inj-frm-rt" // no Eval
 #define SPEED_CONTROL_TOKEN             "-speed-ctrl"
@@ -225,6 +227,8 @@ static void SetRateControlMode                  (const char *value, EbConfig *cf
 static void SetTargetBitRate                    (const char *value, EbConfig *cfg) {cfg->target_bit_rate = strtoul(value, NULL, 0);};
[...]

PR #849, opened 2019-12-02
Adds parameter vbv-bufsize to CVBR mode

klillevo:
This is part of a suggested fix for issue #708, and adds a parameter -vbv-bufsize that if provided changes the vbv buffer size in CVBR mode (rc-mode 3) from 1 second of target bitrate to the parameter's value.  I see from the trello page that there are plans to eventually make the CBR mode VBV compliant, so perhaps this parameter is not the right direction to go. Please provide suggestions if this is the case.


klillevo:
Thanks - I updated my branch to resolve the changes to rate control on master:
1) CVBR mode is now mode 2 (not 3).
2) vbv-bufsize is now also in kbps, to match the same change to target bitrate

EwoutH:
Would be nice to get this in before 0.8.0!

klillevo:
> Could you add a line to the [svt-av1-encoder-user-guide](https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Docs/svt-av1_encoder_user_guide.md)?

I apologize for the delayed reply. That's something I should have included in the original pull request, but I can add that to the documentation (via a new pull request).

Issue #767, opened 2019-11-02
Encoder gets stuck when VBR


mtangtencent:
I was running the encoder (master branch d3880e0c2af53) with the following command line
./SvtAv1EncApp.exe -i BasketballDrill_832x480_50.yuv -w 832 -h 480 -b test.ivf -enc-mode 8 -n 300 -rc 2 -tbr 1000000 -intra-period -1
The encoder got stuck with the log ending at
`Encoding      ` 
There is no number printed after "Encoding", which is the position where the number of frames that have been encoded is printed.

grill05:
I have observed the same issue on a xeon testbed, for enc-modes 3 and below. When the encoder is stuck, the cpu-usage goes to 0%, but SvtAv1EncApp doesn't crash

hassount:
-intra-period -1 would probably be the issue. can you set it to a value for now to make sure it works ?

hassount:
this is related to #710

JackZhouVCD:
closing this issue as it is resolved by #854

PR #854, opened 2019-12-03
fix deadlock in vbr mode:

jerrymiao-tencent:
1、setting feedback_arrived flag should not be in RateControl process，but in PictureManager process before using it.
2、modifying the look_ahead_period, when both intra_period and look_ahead_period are equal to -1.

### DIFF
--- a/Source/Lib/Common/Codec/EbRateControlProcess.c
+++ b/Source/Lib/Common/Codec/EbRateControlProcess.c
@@ -4478,23 +4478,7 @@ void* rate_control_kernel(void *input_ptr)
 
             parentpicture_control_set_ptr = (PictureParentControlSet  *)rate_control_tasks_ptr->picture_control_set_wrapper_ptr->object_ptr;
             sequence_control_set_ptr = (SequenceControlSet *)parentpicture_control_set_ptr->sequence_control_set_wrapper_ptr->object_ptr;
-            if (sequence_control_set_ptr->static_config.rate_control_mode) {
-                ReferenceQueueEntry           *reference_entry_ptr;
-                uint32_t                          reference_queue_index;
-                EncodeContext             *encode_context_ptr = sequence_control_set_ptr->encode_context_ptr;
-                reference_queue_index = encode_context_ptr->reference_picture_queue_head_index;
-                // Find the Reference in the Reference Queue
-                do {
-                    reference_entry_ptr = encode_context_ptr->reference_picture_queue[reference_queue_index];
-                    if (reference_entry_ptr->picture_number == parentpicture_control_set_ptr->picture_number) {
-                        // Set the feedback arrived
-                        reference_entry_ptr->feedback_arrived = EB_TRUE;
-                    }
 
-                    // Increment the reference_queue_index Iterator
-                    reference_queue_index = (reference_queue_index == REFERENCE_QUEUE_MAX_DEPTH - 1) ? 0 : reference_queue_index + 1;
-                } while ((reference_queue_index != encode_context_ptr->reference_picture_queue_tail_index) && (reference_entry_ptr->picture_number != parentpicture_control_set_ptr->picture_number));
-            }
             // Frame level RC
             if (sequence_control_set_ptr->intra_period_length == -1 || sequence_control_set_ptr->static_config.rate_control_mode == 0) {
                 rate_control_param_ptr = context_ptr->rate_control_param_queue[0];
--- a/Source/Lib/Encoder/Codec/EbEncHandle.c
+++ b/Source/Lib/Encoder/Codec/EbEncHandle.c
@@ -1899,7 +1899,7 @@ void SetDefaultConfigurationParameters(
 static uint32_t compute_default_look_ahead(
     EbSvtAv1EncConfiguration*   config){
     int32_t lad = 0;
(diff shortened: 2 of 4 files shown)


JackZhouVCD:
Thanks for the fix! Can you look into addressing the build warnings?

https://ci.appveyor.com/project/tianjunwork/svt-av1/builds/29276547#L537
https://travis-ci.com/OpenVisualCloud/SVT-AV1/jobs/262563742#L761

closes #767

Source/Lib/Common/Codec/EbRateControlProcess.c
```diff
@@ -4486,21 +4486,6 @@ void* rate_control_kernel(void *input_ptr)
...
-                do {
-                    reference_entry_ptr = encode_context_ptr->reference_picture_queue[reference_queue_index];
-                    if (reference_entry_ptr->picture_number == parentpicture_control_set_ptr->picture_number) {
-                        // Set the feedback arrived
-                        reference_entry_ptr->feedback_arrived = EB_TRUE;
-                    }
-
-                    // Increment the reference_queue_index Iterator
-                    reference_queue_index = (reference_queue_index == REFERENCE_QUEUE_MAX_DEPTH - 1) ? 0 : reference_queue_index + 1;
-                } while ((reference_queue_index != encode_context_ptr->reference_picture_queue_tail_index) && (reference_entry_ptr->picture_number != parentpicture_control_set_ptr->picture_number));
```
anaghdin: The feedback_arrived was intentionally added in ratecontrol to update the previous stats. Is this one causing the deadlock?

Source/Lib/Encoder/Codec/EbEncHandle.c
```diff
@@ -1899,7 +1899,7 @@ void SetDefaultConfigurationParameters(
 static uint32_t compute_default_look_ahead(
     EbSvtAv1EncConfiguration*   config){
     int32_t lad = 0;
-    if (config->rate_control_mode == 0)
+    if (config->rate_control_mode == 0 || config->intra_period_length < 0)
```
anaghdin: what is the reason for changing this?

jerrymiao-tencent: Make sure look_ahead_period is non-negative，even if intra_period_length equal to -1.

Source/Lib/Common/Codec/EbRateControlProcess.c
jerrymiao-tencent: The feedback_arrived was added in ratecontrol to update the previous stats.  But using the feedback_arrived flag is in picture manager process.  That's out of sync.  If ratecontrol kernel thread was not scheduled by the CPU in time , it would be late to update the feedback_arrived,  while the picture manager was waiting for next frame to come in,  and did not process the available frame in time.

Source/Lib/Encoder/Codec/EbEncHandle.c
anaghdin: i understand, but why not to 0?

Source/Lib/Common/Codec/EbRateControlProcess.c
anaghdin: Yes, makes sense. But since the feedback is about number of bits, you need to add the feedback_arrived under EB_PIC_FEEDBACK (from packetization) case not EB_PIC_REFERENCE (from Restoration for the reconstructed picture) in picturemanager. You might need to change the condition in packetization for sending the feedback. Thanks.

anaghdin: I think the problem has been addressed in SVT-HEVC.

Source/Lib/Encoder/Codec/EbEncHandle.c
jerrymiao-tencent: sorry, do you mean the condition should be "config->intra_period_length <= 0 " ?  
If intra_period_length is equal to 0,  it means that's all the intra frame.  I don't think look ahead is needed in this case.

Source/Lib/Common/Codec/EbRateControlProcess.c
jerrymiao-tencent: > Yes, makes sense. But since the feedback is about number of bits, you need to add the feedback_arrived under EB_PIC_FEEDBACK (from packetization) case not EB_PIC_REFERENCE (from Restoration for the reconstructed picture) in picturemanager. You might need to change the condition in packetization for sending the feedback. Thanks.

Yes.  thanks for reminding me.

Source/Lib/Encoder/Codec/EbEncHandle.c
anaghdin: Thanks for making the changes. I meant why not setting the look ahead to zero?

Source/Lib/Common/Codec/EbRateControlProcess.c
anaghdin: thanks for making the changes

Source/Lib/Encoder/Codec/EbEncHandle.c
jerrymiao-tencent: as far as I know, setting look ahead to zero is not yet supported.

PR #855, opened 2019-12-03
Enable AVX2 kernel eb_av1_fwd_txfm2d_32x16() to calculate DCT_DCT

spawlows:
Extend UT to cover kernel eb_av1_fwd_txfm2d_32x16 for DCT_DCT calculation.

I see increase performance of encoder  1280x720 10bits in mode 0 around 1.2%

### DIFF
--- a/test/TxfmCommon.h
+++ b/test/TxfmCommon.h
@@ -170,14 +170,45 @@ static INLINE int get_txfm1d_size(TxfmType txfm_type) {
 }
 
 static INLINE bool is_txfm_allowed(TxType tx_type, TxSize tx_size) {
-    /* According to 5.11 */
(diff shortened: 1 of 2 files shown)


test/TxfmCommon.h
```diff
@@ -170,14 +170,45 @@ static INLINE int get_txfm1d_size(TxfmType txfm_type) {
...
+        case TX_32X16:
+        case TX_64X64:
+        case TX_8X32:
+        case TX_32X8: {
+            int types[] = {DCT_DCT, IDTX, -1};
+            support_types = types;
+            break;
+        }
+    }
+
```
mafonsoN: For clarity, might be good to add the default case as well instead of assigning support_types before, in which case all types are supported.

Issue #852, opened 2019-12-03
Set enable_adaptive_quantization = 1 would cause assert failed in the encoder with debugging build


wallerjiaotencent:
I tried to set enable_adaptive_quantization = 1; when debug build, the encoder throw out "debug assert failed" message; when release build, the encoder can run normally and output playable bitstream.
It looks like some variables have not been initialized. The debugging build is on windows platform.

wallerjiaotencent:
The debugging build is on windows platform.

JackZhouVCD:
Thanks for reporting this issue! Can you share with us the command and perhaps the patch (as plaintext) you used that generated the assertion fail?

wallerjiaotencent:
The following command line triggers this issue:
SvtAv1EncApp.exe -i FourPeople_1280x720_60.yuv -w 1280 -h 720 -b D:\testSequences\test.ivf -enc-mode 8 -n 200 -rc 0 -q 40 -intra-period 31 -adaptive-quantization 1

JackZhouVCD:
> The following command line triggers this issue(with commit [e54f59e](https://github.com/OpenVisualCloud/SVT-AV1/commit/e54f59eb9041c28dd97497d5d8fbe67f40c7606f)):
> SvtAv1EncApp.exe -i FourPeople_1280x720_60.yuv -w 1280 -h 720 -b test.ivf -enc-mode 8 -n 200 -rc 0 -q 40 -intra-period 31 -adaptive-quantization 1

I see, thanks!

I tried to run with this command as well, same issue happens with `dark720p` clip. I found the following failed assert:

```css
Assertion failed: x < max, Source\Lib\Common\Codec\EbEntropyCoding.c, line 5864
```

I did a bit to digging and found that `cu_ptr->segment_id` was uninitialized. Setting it to 0 resolved the issue but the resulting bitstream has VERY prominent blocking artifacts.

@anaghdin can you please look into this?

anaghdin:
@JackZhouVCD Thanks for checking it. I recall it was working. Could you do a regression test to figure out which commit broke it?

JackZhouVCD:
> . I recall it was working. Could you do a regression test to figure out which commit broke it?

For sure, actually this issue happens even in the initial commit 47e6f527

anaghdin:
> > . I recall it was working. Could you do a regression test to figure out which commit broke it?
> 
[...]

PR #863, opened 2019-12-06
fix assert failed when setting adaptive_quantization = 1 with debugging build

wallerjiaotencent:
This issue #852 is caused by some variables uninitialized.


anaghdin:
I am not sure this is the only issue. cu_ptr->segment_id should be set properly later. Would segments work properly after adding the fix?

wallerjiaotencent:
After adding the fix, I have tested tens of cases, all things looks fine.

anaghdin:
> e

OK thanks. looks good to me

PR #869, opened 2019-12-09
Add 10bit test to Travis

### LINKED ISSUE #850: [Task] add 10 bit encoding test to Travis
A trial encoding of a sample 10-bit sequence needs to be added to the Travis CI for the Linux/Mac/Windows builds and for the binary identical test.

Related to #622.

mafonsoN:
### Description
- A 10-bit sequence, Chimera-Aerial 480x270, is now available in video.tar.gz. Thanks @1480c1 for copying it to the mirror
- Add 10-bit sequence to the Travis build tests (enc-mode 0, 3 frames, and enc-mode 8, 120 frames)
- Add 8-bit to Travis build tests (enc-mode 0, 3 frames)
- Add 10-bit sequence to binary identical tests (enc-mode 0 and 8)

Closes #850

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,10 @@ script:
     cmake $TRAVIS_BUILD_DIR -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release} ${CMAKE_EFLAGS}
     cmake -j $(if [ $TRAVIS_OS_NAME = osx ]; then sysctl -n hw.ncpu; else nproc; fi) --build . &&
     sudo cmake --build . --target install && cd $TRAVIS_BUILD_DIR
-  - SvtAv1EncApp -enc-mode 8 -i akiyo_cif.y4m -n 150 -b test1.ivf
+  - &8bit_test |
+    SvtAv1EncApp -enc-mode 0 -i akiyo_cif.y4m -n 3 && SvtAv1EncApp -enc-mode 8 -i akiyo_cif.y4m -n 150
+  - &10bit_test |
+    SvtAv1EncApp -enc-mode 0 -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m -n 3 && SvtAv1EncApp -enc-mode 8 -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m -n 150
 before_cache:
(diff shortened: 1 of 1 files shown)


mafonsoN:
@hassount let me know if you think we should add enc-mode 4 test as well.

.travis.yml
```diff
@@ -78,7 +78,10 @@ script:
     cmake $TRAVIS_BUILD_DIR -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release} ${CMAKE_EFLAGS}
     cmake -j $(if [ $TRAVIS_OS_NAME = osx ]; then sysctl -n hw.ncpu; else nproc; fi) --build . &&
     sudo cmake --build . --target install && cd $TRAVIS_BUILD_DIR
+  - SvtAv1EncApp -enc-mode 0 -i akiyo_cif.y4m -n 3 -b test1.ivf
   - SvtAv1EncApp -enc-mode 8 -i akiyo_cif.y4m -n 150 -b test1.ivf
+  - SvtAv1EncApp -enc-mode 0 -i Chimera-Aerial_480x270_2997fps_10bit_420_150frames.y4m -n 3 -b test1.ivf
+  - SvtAv1EncApp -enc-mode 8 -i Chimera-Aerial_480x270_2997fps_10bit_420_150frames.y4m -n 150 -b test1.ivf
```
1480c1: Perhaps we don't need the `-b test1.ivf` argument on these since we aren't using the remaining file?

Would not outputting the file have any benefits or detriments to testing in our case since we aren't testing decodability in the base jobs?

mafonsoN: True. For these ones, it should be fine to remove the output bitstream. 
I don't believe we have any test for decodability yet. But it should be added. Perhaps in another PR though. Also, a rec/dec mismatch test.

mafonsoN:
> Thank you both, I think we should start with this for now then add -enc-mode 4 and decode tests in another PR

Okay. 

You can see in the checks that the macOS test is failing due to a seg fault.

hassount:
> You can see in the checks that the macOS test is failing due to a seg fault.

@mafonsoN It seems that the 10-bit yuv is not a multiple of 8 resolution where we still have issues. can we change the y4m for 10bit to be a multiple of 8, that way we can at least try to track the crash back because right now after some point we can't.
seems that this issue is too old though

PR #872, opened 2019-12-09
Implementation of function av1_k_means_dim1, av1_k_means_dim2 for AVX2

spawlows:
Add AVX2 implementation for kernels av1_k_means_dim1(), av1_k_means_dim2() and similar kernels 
av1_calc_indices_dim1() av1_calc_indices_dim2().

Based on command line below, we have 23% performance improvement.
 ```-enable-altrefs 0 -pred-struct 0 -bit-depth 8 -b out.ivf  -enc-mode 8 -w 1920 -h 1080 -i 1080p.yuv -palette 1 -scm 1 ```

### DIFF
--- a/Source/Lib/Common/ASM_AVX2/palette_avx2.c
+++ b/Source/Lib/Common/ASM_AVX2/palette_avx2.c
@@ -1,0 +1,438 @@
+/*
+* Copyright(c) 2019 Intel Corporation
+* SPDX - License - Identifier: BSD - 2 - Clause - Patent
+*/
+
+#include <immintrin.h>
+#include "EbDefinitions.h"
+
+#define DIVIDE_AND_ROUND(x, y) (((x) + ((y) >> 1)) / (y))
+
+static INLINE unsigned int lcg_rand16(unsigned int *state) {
+    *state = (unsigned int)(*state * 1103515245ULL + 12345);
+    return *state / 65536 % 32768;
+}
+
+/* That same calcualtion as: av1_calc_indices_dist_dim1_avx2(),
+   but not calculate sum at the end. */
+void av1_calc_indices_dim1_avx2(const int* data,
+    const int* centroids,
+    uint8_t* indices, int n, int k) {
+
+    int i = 0;
+    int results[MAX_SB_SQUARE];
+    memset(indices, 0, n * sizeof(uint8_t));
+
+    __m256i centroids_0 = _mm256_set1_epi32(centroids[0]);
+    for (i = 0; i < n; i += 8) {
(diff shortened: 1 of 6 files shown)


grill05:
> Based on command line below, we have 23% performance improvement.

@spawlows  Just to clarify, the 23% improvement is in the total run-time of that command, or just in the time spent in the kernel ?

spawlows:
Performance change for total run-time, on Windows:

Mode | Master FPS | Changes FPS | speed %
-- | -- | -- | --
0 | 0.141 | 0.149 | 5.4
1 | 0.332 | 0.3575 | 7.1
2 | 0.449 | 0.481 | 6.7
3 | 0.985 | 1.121 | 12.1
4 | 2.031 | 2.28 | 10.9
5 | 2.387 | 2.7425 | 13.0
6 | 2.768 | 3.3315 | 16.9
7 | 3.535 | 4.4135 | 19.9
8 | 4.061 | 5.33 | 23.8

This changes work only with params: ```-palette 1 -scm 1```

On Linux we do not see any difference in execution time. From speed test on Windows and Linux kernels work few times faster, so this still reduce CPU consumption. I suppose that on Linux there are difference in threads scheduling and calculation of those kernels are not bottleneck.

spawlows:
When i run this with one core ```-lp 1``` situation is similar, on Windows we have speed benefits, on linux not. So probably this is some differences between VS and gcc

spawlows:
Performance change for total run-time, on Windows, and Linux:

Mode | Master FPS | Changes FPS | speed %
-- | -- | -- | --
0 | 0.141 | 0.149 | 5.4
1 | 0.332 | 0.3575 | 7.1
2 | 0.449 | 0.481 | 6.7
3 | 0.985 | 1.121 | 12.1
4 | 2.031 | 2.28 | 10.9
5 | 2.387 | 2.7425 | 13.0
6 | 2.768 | 3.3315 | 16.9
7 | 3.535 | 4.4135 | 19.9
8 | 4.061 | 5.33 | 23.8



After investigation why there is no improvement on linux, I find out that I wrongly executed tests.
For clarification:
"SvtAv1EncApp" use library from build directory even if you copy all the files to another one. So when I build with and without the changes (and copy binaries to another directory), encoder loaded  "libSvtAv1Enc.so.0" from build directory. To fix this issue i would recommend to remove files in Bin/(Release/Debug) before tests. "SvtAv1EncApp" is looking for "libSvtAv1Enc.so.0" in previous build path and then in current path. That might be centos-related issue.

hassount:
@mafonsoN any issues with this ?

PR #874, opened 2019-12-10
[SIMD] Update eb_av1_wiener_convolve_add_src AVX2 and add AVX512

lzhangnj:
Speed gains of new AVX2 vs old AVX2 with size 64x64:
Tap 7: 10%
Tap 5: 27%
Tap 3: 95%

Speed gains of AVX512 vs new AVX2 with size 64x64:
Tap 7: 50%
Tap 5: 50%
Tap 3: 60%


lzhangnj:
Hi @ttrigui I think I met a Visual Studio compiler bug, and had to work around by reordering the instructions, and couldn't get a solid solution. I put notes in the AVX2 and AVX512 functions. Could you please help test on different platforms to verify the whole encoder won't crash?

Thanks!

ttrigui:
> Hi @ttrigui I think I met a Visual Studio compiler bug, and had to work around by reordering the instructions, and couldn't get a solid solution. I put notes in the AVX2 and AVX512 functions. Could you please help test on different platforms to verify the whole encoder won't crash?
> 
> Thanks!

Thank you @lzhangnj for the comment. It looks like all the checks have passed. but we will certainly do more testing.

hassount:
@joelsole any issues with this ?

EwoutH:
On 8-bit 1080p content, I actually measured a [performance decrease](https://docs.google.com/spreadsheets/d/1p3PJQMkyhIrXEL6MBBQiwTnWmGyHY_zuw7_XAD0def4/edit#gid=262152911) when applying this PR compared to the current master branch (27b95fb):

| Settings    | Master | PR 874 |       |
|-------------|-------:|-------:|------:|
| -enc-mode 8 | 62,136 | 57,999 | -6,7% |
| -enc-mode 6 | 25,815 | 24,907 | -3,5% |
| -enc-mode 4 |  7,814 |  7,654 | -2,0% |
| -enc-mode 2 |  1,103 |  1,075 | -2,5% |
| -enc-mode 0 |  0,486 |  0,474 | -2,5% |

It could be because of reduced clock speeds AVX-512 is causing. Maybe someone could measure this on a system with fixed clock speeds?

EwoutH:
Using `-asm avx2` results are better, but inconclusive (due to measurement noise).

| Settings    | Master | PR 874 |       |
|-------------|-------:|-------:|------:|
| -enc-mode 8 | 61,038 | 61,963 |  1,5% |
| -enc-mode 6 | 25,989 | 26,050 |  0,2% |
| -enc-mode 4 |  7,855 |  8,005 |  1,9% |
| -enc-mode 2 |  1,108 |  1,102 | -0,5% |
| -enc-mode 0 |  0,489 |  0,486 | -0,6% |

@hassount @kylophone Do you think you could get me access to a machine with fixed clock speeds independent of SIMD level that's used? I want to get measurement noise below 0.001 (0.1%), but that's almost impossible on Cloud instances.

When I get that done, I will see if I can integrate it in the CI so that we can measure performance accurately for each PR.

hassount:
@JackZhouVCD 
can you run a quick test on this to make sure we're good, please make sure to fill up the cpu

JackZhouVCD:
> @JackZhouVCD
> can you run a quick test on this to make sure we're good, please make sure to fill up the cpu

Yup, for sure! I'm seeing a 1.71% speed gain in M0 on 360p

hassount:
thnx @JacksonDagger 
@joelsole please let us know if we can move forward with this

PR #875, opened 2019-12-10
reduce_inter_modes

NaderMahdi:
Reduce the number of candidates with new mv based on the weight computed in mv pred. When it is ON, 3% speed gain for 0.085% Bdrate loss is reported for M0.

### DIFF
--- a/Source/Lib/Common/Codec/EbModeDecision.c
+++ b/Source/Lib/Common/Codec/EbModeDecision.c
@@ -995,7 +995,47 @@ EbBool mrp_is_already_injected_mv_bipred(
 int8_t ALLOW_REFINEMENT_FLAG[BIPRED_3x3_REFINMENT_POSITIONS] = {  1, 0, 1, 0, 1,  0,  1, 0 };
 int8_t BIPRED_3x3_X_POS[BIPRED_3x3_REFINMENT_POSITIONS] = { -1, -1, 0, 1, 1, 1, 0, -1 };
 int8_t BIPRED_3x3_Y_POS[BIPRED_3x3_REFINMENT_POSITIONS] = { 0, 1, 1, 1, 0, -1, -1, -1 };
+#if REDUCE_INTER_MODES
+// Whether this reference motion vector can be skipped, based on initial
+// heuristics.
+static bool ref_mv_idx_early_breakout(
+    PictureControlSet            *picture_control_set_ptr,
+    ModeDecisionContext          *context_ptr,
+    const int8_t                 inter_direction,
+    const int8_t                 ref_idx0,
+    const int8_t                 ref_idx1,
+    const int8_t                 ref_mv_idx,
+    PredictionMode               pred_mode) {
 
+    int  is_comp_pred = inter_direction == 2 ? 1 : 0;
+    if (inter_direction == 0) {
+        if (picture_control_set_ptr->parent_pcs_ptr->reduce_inter_modes && ref_mv_idx > 0) {
+            MvReferenceFrame ref_frame_type = svt_get_ref_frame_type(REF_LIST_0, ref_idx0);
+            if (ref_frame_type == LAST2_FRAME ||
+                ref_frame_type == LAST3_FRAME ) {
+                const int has_nearmv = have_nearmv_in_inter_mode(pred_mode) ? 1 : 0;
+                int32_t weight = context_ptr->md_local_cu_unit[context_ptr->blk_geom->blkidx_mds].ed_ref_mv_stack[ref_frame_type][ref_mv_idx + has_nearmv].weight;
+                if (weight < REF_CAT_LEVEL) {
+                    return true;
+                }
+            }
+        }
+    }else if (inter_direction == 1) {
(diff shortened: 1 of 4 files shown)


mafonsoN:
Hi @NaderMahdi, can you clarify what test set these results are based on?

Source/Lib/Common/Codec/EbModeDecision.c
```diff
@@ -995,7 +995,47 @@ EbBool mrp_is_already_injected_mv_bipred(
...
+    ModeDecisionContext          *context_ptr,
+    const int8_t                 inter_direction,
+    const int8_t                 list_idx0,
+    const int8_t                 list_idx1,
+    const int8_t                 ref_idx0,
+    const int8_t                 ref_idx1,
+    const int8_t                 ref_mv_idx,
+    PredictionMode               pred_mode) {
+    //int  is_comp_pred = inter_direction == 2 ? 1 : 0;
+    if (inter_direction == 0) {
```
mafonsoN: Can you use UNI_PRED_LIST_0 and UNI_PRED_LIST_1 instead of 0 and 1, respectively?

Source/Lib/Common/Codec/EbModeDecision.c
```diff
@@ -995,7 +995,47 @@ EbBool mrp_is_already_injected_mv_bipred(
...
+            if (ref_frame_type == ALTREF2_FRAME ) {
+                const int has_nearmv = have_nearmv_in_inter_mode(pred_mode) ? 1 : 0;
+                int32_t weight = context_ptr->md_local_cu_unit[context_ptr->blk_geom->blkidx_mds].ed_ref_mv_stack[ref_frame_type][ref_mv_idx + has_nearmv].weight;
+                if (weight < REF_CAT_LEVEL) {
+                    return true;
+                }
+            }
+        }
+    }
+    return false;
```
mafonsoN: What about inter_direction == BI_PRED?

Source/Lib/Common/Codec/EbModeDecision.c
```diff
@@ -995,7 +995,47 @@ EbBool mrp_is_already_injected_mv_bipred(
...
+    PictureControlSet            *picture_control_set_ptr,
+    ModeDecisionContext          *context_ptr,
+    const int8_t                 inter_direction,
+    const int8_t                 list_idx0,
+    const int8_t                 list_idx1,
+    const int8_t                 ref_idx0,
+    const int8_t                 ref_idx1,
+    const int8_t                 ref_mv_idx,
+    PredictionMode               pred_mode) {
+    //int  is_comp_pred = inter_direction == 2 ? 1 : 0;
```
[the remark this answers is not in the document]
NaderMahdi: No we don't need it. Removed.

Source/Lib/Common/Codec/EbModeDecision.c
NaderMahdi: Done

Source/Lib/Common/Codec/EbModeDecision.c
NaderMahdi: Not considered for this commit.

Source/Lib/Common/Codec/EbPictureDecisionProcess.c
```diff
@@ -1411,6 +1411,9 @@ EbErrorType signal_derivation_multi_processes_oq(
         // GM_DOWN                                    Downsampled resolution with a downsampling factor of 2 in each dimension
         // GM_TRAN_ONLY                               Translation only using ME MV.
         picture_control_set_ptr->gm_level = GM_FULL;
+#endif
+#if REDUCE_INTER_MODES
+        picture_control_set_ptr->reduce_inter_modes = 0;
```
hassount: need a configuration flag for this option with AUTO_MODE, ON, and OFF
also some comments explaining this feature

joelsole:
The slope is ~0.03, which is 3x the minimum threshold discussed
[here](https://github.com/OpenVisualCloud/SVT-AV1/pull/882#issuecomment-565134177)
It might be worth tuning it before merging.

hassount:
> The slope is ~0.03, which is 3x the minimum threshold discussed [here](https://github.com/OpenVisualCloud/SVT-AV1/pull/882#issuecomment-565134177).
> It might be worth tuning it before merging.

good point, I believe it's off by default though and it's there to be used for other presets

joelsole:
> good point, I believe it's off by default though and it's there to be used for other presets

I see. So, what is the performance for the presets for which this is intended to be used?

A suggestion would be to test for reduce_inter_modes being enabled at the beginning of the ref_mv_idx_early_breakout function and return accordingly, so the dependency is more explicit.

hassount:
> I see. So, what is the performance for the presets for which this is intended to be used?

This code is being added to allow the team to use it during the preset optimizations, it is not targeted for any preset yet until being optimized further. Otherwise we can remove it.

EwoutH:
Could you resolve the conflicts and rebase?

hassount:
Can we please retest this PR for performance?

NaderMahdi:
Closing as not applicable anymore

PR #883, opened 2019-12-11
Enhanced M0 Settings

hguermaz:
## Description

Updated M0 settings:

- Optimized independent chroma search for all layers.
- Conservative coeff-based NSQ cands reduction.
- Shut coeff-based skip tx size search, 
- Warped for all layers.
- SUB-SAD as ME search method for non-SC only.

solves #884
## Author(s)

@PhoenixWorthVCD 
@hguermaz 

## Type of change

Enhancement

## Tests and performance
8BIT data generated for full-list objective-1-fast dataset in context of enc-mode 0. Speed data is generated on the full-list objective-1-fast dataset in context of enc-mode 0 for 5QP in term of cycles on AWS ec2 48 vCPU instances.

| Cycle Increase | BD-Rate Gain|
| -- | --
| 5.5% | 0.20%|

10BIT and jctvc 240p, 480p and 720p clips in context of enc-mode 0. Speed data is generated on the same list for 5QP in term of cycles on AWS ec2 48 vCPU instances.

| Cycle Increase | BD-Rate Gain|
| -- | --
| 8.5% | 0.24%|

Evaluation performed in context of QPS/QPM OFF.

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -33,6 +33,7 @@
 extern "C" {
 #endif
 
+#define ENHANCED_M0_SETTINGS         1 // Updated M0 settings(optimized independent chroma search for all layers, conservative coeff - based NSQ cands reduction, shut coeff - based skip tx size search, warped for all layers, SUB - SAD as ME search method for non - SC only)
 #define MULTI_PASS_PD                1 // Multi-Pass Partitioning Depth (Multi-Pass PD) performs multiple PD stages for the same SB towards 1 final Partitioning Structure. As we go from PDn to PDn + 1, the prediction accuracy of the MD feature(s) increases while the number of block(s) decreases
 
 #define HBD_CLEAN_UP                 1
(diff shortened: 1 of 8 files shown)


joelsole:
- What is the reason to no follow the usual pattern of adding candidates in the injection (EbModeDecision)? 
- Some explanation could be added to the thresholds such as md_stage_2_class_prune_th = 25
- Just to clarify: this PR makes encoder faster AND performing better? We should have a clear sign convention for this.

hguermaz:
- This PR does not inject additional candidate(s) to md_stage_0 (). search_best_independent_uv_mode() has been refactored/cleaned-up towards a fast-loop/full-loop structure, but this path is still independent from the MD Staging path (main MD path).
-  I agree. Actually there is short explanation: //Remove class if deviation to the best is higher than TH_C (but TH_C should be replaced by md_stage_2_class_prune_th).  We will fix this text and similar typo(s).
- Sorry for the confusion. Actually this PR is causing a speed drop since speed deviations are negatives. But to emphasize that the # of cycles actually increased we will update the performance summary section.

hassount:
@hguermaz can you please rebase and squash

JackZhouVCD:
> * Sorry for the confusion. Actually this PR is causing a speed drop since speed deviations are negatives. But to emphasize that the # of cycles actually increased we will update the performance summary section.

If that is the case, we should remove the `S+` label from this PR

PR #887, opened 2019-12-12
Update auto max partitioning to be performed only for M1 and higher

JackZhouVCD:
As auto max partitioning is a inference based feature using a neural network, it may not provide ideal predictions all the time and thus might impact quality. We are going to move this feature away from M0 to preserve stream quality and have this feature's shortcuts benefit higher enc-modes


hguermaz:
@JackZhouVCD Actually the MR_MODE check is useless. Could you please remove it ?

joelsole:
lgtm - could you ask a brief description of the PR, including the motivation?

JackZhouVCD:
> lgtm - could you add a brief description of the PR, including the motivation?

yup, done!

PR #894, opened 2019-12-13
Initial version of decoder with multi-thread support

nijilk:
Support for MT added for stages except LR filtering, Motion filed projection and Frame padding.

### DIFF
--- a/Source/Lib/Common/Codec/EbSystemResourceManager.c
+++ b/Source/Lib/Common/Codec/EbSystemResourceManager.c
@@ -280,6 +280,39 @@ static EbErrorType EbMuxingQueueAssignation(
     return return_error;
 }
 
+static EbErrorType EbMuxingQueueAssignationNonBlocking(
+    EbMuxingQueue *queue_ptr)
+{
+    EbErrorType return_error = EB_ErrorNone;
+    EbFifo *processFifoPtr;
+    EbObjectWrapper *wrapper_ptr;
+
+    // while loop
+    while ((EbCircularBufferEmptyCheck(queue_ptr->object_queue) == EB_FALSE) &&
+        (EbCircularBufferEmptyCheck(queue_ptr->process_queue) == EB_FALSE))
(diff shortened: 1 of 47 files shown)


Source/Lib/Common/Codec/EbSystemResourceManager.c
```diff
@@ -278,6 +278,39 @@ static EbErrorType EbMuxingQueueAssignation(
...
+        EbFifoPushBack(
+            processFifoPtr,
+            wrapper_ptr);
+
+        // Post the semaphore
+        eb_post_semaphore(processFifoPtr->counting_semaphore);
+    }
+
+    return return_error;
+}
```
hassount: @chkngit  ok with this ?

hassount: @nijilk 
these new APIs do not seem to be called anywhere in the code added, during the code cleanup this might get picked up as dead code and so it might be removed.

hassount:
LGTM, apart from the comment above

Source/Lib/Common/Codec/EbSystemResourceManager.c
nijilk: > 
> 
> @nijilk
> these new APIs do not seem to be called anywhere in the code added, during the code cleanup this might get picked up as dead code and so it might be removed.

@hassount 
These are used in decoder queue management.
We modified the existing APIs to make it more suitable for light weight frame row work jobs of decoder.

hassount: Sounds good, please rebase the branch and squash all commits into 1 so that we can merge this PR

nijilk: > 
> 
> Sounds good, please rebase the branch and squash all commits into 1 so that we can merge this PR

@hassount 
Sure.
Will do that.

PR #897, opened 2019-12-16
Actions: Move style check job over to gh actions

1480c1:
Runs in 6 seconds instead of 26 it seems. Idk how trustable those numbers are.

Frees up travis a bit to not have so many stages

@EwoutH and @kylophone, please check

### DIFF
--- a/.github/workflows/style-check.yml
+++ b/.github/workflows/style-check.yml
@@ -1,0 +1,43 @@
+name: Style Check
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      - name: Tab check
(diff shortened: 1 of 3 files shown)


1480c1:
I am looking up to see how to work with a feedback of where the issue are, but in so far, it seems it may be restricted to javascript based actions for now

EwoutH:
Looks good overall. My only concern is Travis now running the full stack while there are styling errors, I don't know if that's desired behavior.

1480c1:
I think that may be fine since gh actions will fail the pr, but we can still test if the code compiles and runs correctly in travis. That way the author can test if their code if fine, and then fixup style issues afterwards once the PR is finally ready

hassount:
are we good to go with this ?

1480c1:
I have no more plans to add to this, so this is good from my side

PR #903, opened 2019-12-17
Fix valgrind issue in eb_av1_selfguided_restoration_avx2()

spawlows:
Fix one of valgrind errors from #733

Kernel eb_av1_selfguided_restoration_avx2() read neighboring pixels from right side.
Initialize memory before use.

### DIFF
--- a/Source/Lib/Common/ASM_AVX2/selfguided_avx2.c
+++ b/Source/Lib/Common/ASM_AVX2/selfguided_avx2.c
@@ -58,6 +58,7 @@ static AOM_FORCE_INLINE void integral_images(const uint8_t *src,
     const uint8_t *srcT = src;
     int32_t *CT = C + buf_stride + 1;
     int32_t *DT = D + buf_stride + 1;
+    const __m256i zero = _mm256_setzero_si256();
 
     memset(C, 0, sizeof(*C) * (width + 8));
     memset(D, 0, sizeof(*D) * (width + 8));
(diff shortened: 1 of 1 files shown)


Source/Lib/Common/ASM_AVX2/selfguided_avx2.c
```diff
@@ -216,6 +224,12 @@ static AOM_FORCE_INLINE void integral_images_highbd(const uint16_t *src,
             x += 8;
         } while (x < width);
 
+        /* Used in calc_abc and calc_abc_fast, when calc out of right border */
```
ttrigui: Can this init loop be combined with the memset instead ?
Please modify the comment by replacing calc_abc and calc_abc_fast by calc_ab and calc_ab_fast .

spawlows: This loop is not for continuous memory. Init memory at the end of lines. 

Example with memset:
```
   for (int line = 0; line <8; ++line) {
            memset(CT + width + line * buf_stride, 0, sizeof(*CT) * (8));
            memset(DT + width + line * buf_stride, 0, sizeof(*DT) * (8));
        }
```
I can change this code, but potentially AVX2 code should work faster.

ttrigui: let's leave the avx2 code then.

PR #905, opened 2019-12-17
v0.8.0 Prep [No performance PRs after this PR, only cleanups and bugfixes]

hassount:


joelsole:
What is the expected duration of the clean-up/bugfix period?

hassount:
> What is the expected duration of the clean-up/bugfix period?

The thought is for the initial phase of Jan 6th, then we can start taking PRs in again, does that sound good ?

joelsole:
Yes, thanks @hassount

PR #910, opened 2019-12-18
eb_av1_highbd_quantize_fp_avx2: fix stack overwrite

xuguangxin:
it's introduced by https://github.com/OpenVisualCloud/SVT-AV1/commit/f29b8124c938c370c1ed0036be48776209becee1

PR #911, opened 2019-12-18
Fix quant_hbd and macos 10b crash

joelsole:
- As in #910, fix stack overwrite eb_av1_highbd_quantize_fp_avx2
- Disallow macOS 10-b test to fail
   - Needed to set 'lp' to 1 for it to NOT fail (to be changed when that is fixed)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,7 @@ script:
   - &8bit_test |
     SvtAv1EncApp -enc-mode 0 -i akiyo_cif.y4m -n 3 && SvtAv1EncApp -enc-mode 8 -i akiyo_cif.y4m -n 150
   - &10bit_test |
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -80,7 +80,7 @@ script:
   - &8bit_test |
     SvtAv1EncApp -enc-mode 0 -i akiyo_cif.y4m -n 3 && SvtAv1EncApp -enc-mode 8 -i akiyo_cif.y4m -n 150
   - &10bit_test |
-    SvtAv1EncApp -enc-mode 0 -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m -n 3 && SvtAv1EncApp -enc-mode 8 -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m -n 150
```
hassount: is this intentional

hassount: nvm I see why

xuguangxin:
@joelsole 
yes, this fixed the issue. 
thanks

PR #912, opened 2019-12-19
Remove SIMD unused kernels

OmranAbdelkafi:
## Description
Removed unused kernels

## Related issue
#668

## Performance
No impact


Source/Lib/Common/Codec/EbProductCodingLoop.c
```diff
@@ -5081,18 +5081,6 @@ void full_loop_core(
             &cb_coeff_bits,
             &cr_coeff_bits,
             context_ptr->blk_geom->bsize);
-
-        candidate_buffer->cb_distortion[DIST_CALC_RESIDUAL] = cbFullDistortion[DIST_CALC_RESIDUAL];
```
hassount: Is this not being used?

1480c1:
@hassount @OmranAbdelkafi, this PR removed the contents of some files, but left the empty file in tree. Is that intended?

hassount:
> @hassount @OmranAbdelkafi, this PR removed the contents of some files, but left the empty file in tree. Is that intended?

@1480c1 this is a first phase, we still have a few other PRs coming over the next couple of days, we’ll do a sweep as soon as we’re done. That said, this was not intended.

PR #914, opened 2019-12-19
Github Actions CI: Add build matrix

ePirat:
Add a build matrix to test compilation on various compilers.

Additionally set an explicit name for the build machine, so that it is clear on which OS version it runs.

### DIFF
--- a/.github/workflows/svt-av1.yml
+++ b/.github/workflows/svt-av1.yml
@@ -7,7 +7,7 @@ jobs:
 
     strategy:
       matrix:
-        compiler: ['gcc-7', 'gcc-8', 'gcc-9']
+        compiler: ['gcc-7', 'gcc-8', 'gcc-9', 'clang-6']
         include:
           - compiler: gcc-7
             packages: gcc-7 g++-7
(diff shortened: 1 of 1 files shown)


hassount:
@ePirat Thank you for the contribution. This LGTM

@1480c1 @EwoutH @kylophone any issues with this, the plan is to expand on github actions and hook up external instances

1480c1:
Looks good to me



> @1480c1 @EwoutH @kylophone any issues with this, the plan is to expand on github actions and hook up external instances

That would be great, that means we can finally control the CPU and ram and test further than just two video files

EwoutH:
LGTM, nice work

hassount:
@kylophone merging assuming no issues

1480c1:
> the plan is to expand on github actions and hook up external instances

Will this apply to the other SVT encoders or just SVT-AV1 for now?

hassount:
> > the plan is to expand on github actions and hook up external instances
> 
> Will this apply to the other SVT encoders or just SVT-AV1 for now?

This is planned for svt-av1 for now.

PR #919, opened 2019-12-21
Appveyor: Port changes from SVT-HEVC and fix Bin/*/*/ folder

1480c1:
@hassount, should appveyor produce static lib+exe or dll+exe?

I personally prefer static lib and exe due to dependency issues with windows, but I've also heard that some people like dll+exe since they prefer to just only replace the dll instead of caring about changes to the exe.

Adds a ffmpeg integration build on windows too


hassount:
@1480c1 
sure static builds are good. Would you be able to add an ffmpeg build check on Appveyor ?

1480c1:
Already added

1480c1:
I am planning to work conflicts with libaom to be able to re-enable tests and so ffmpeg won't conflict starting tomorrow

hassount:
@1480c1 thank you for the PR, it's great to see that we can have some testing for windows, a few questions:
- It seems that the test projects are not being built anymore, would we be able to add them back? 
- it seems that the decoder builds are not running anymore, would we be able to add them back?

1480c1:
> * It seems that the test projects are not being built anymore, would we be able to add them back?

Currently, it's being blocked by the symbol conflicts with libaom, right now working on fixing that, but I can switch to dynamic libs and re-enable testing for now



> * it seems that the decoder builds are not running anymore, would we be able to add them back?

I forgot to add the decoder lib and exec, going to now

hassount:
@1480c1 
Thnx for addressing these items, would you be able to rebase it to make sure it works well with the current master after #917

1480c1:
image

1480c1:
Right now updating Visual Studio to try to debug

1480c1:
Planning to move most of this to Github action later

hassount:
ready to go ? if so can you please rebase to make sure the new Appveyor storage limit has been increased ?

1480c1:
It seems the appveyor limit's been raised, ready from my side

1480c1:
@hassount, should I make it so that artifacts are only uploaded on master branch or zip up the files into one zip then send as archive?

hassount:
> @hassount, should I make it so that artifacts are only uploaded on master branch or zip up the files into one zip then send as archive?
> 
> Alternatively, I can make it so that for all builds other than tagged releases, the artifacts will be zip files, with releases containing the regular files instead

we probably only need them for the master branch and tags, so might be best to keep it at that. trying to get an increase of storage from Appveyor in the meantime

1480c1:
Done

1480c1:
Crap, I left a fixup commit

hassount:
> Crap, I left a fixup commit

they're both building it's ok

PR #926, opened 2019-12-27
use optimized version for convolve intraBC

ttrigui:
use optimized code for eb_av1_convolve_2d_sr, eb_av1_convolve_x_sr and eb_av1_convolve_y_sr


hassount:
@ttrigui can you please squash the commits so that we can merge this ?

Issue #873, opened 2019-12-09
CI: GitHub Actions tracking issue


EwoutH:
A tracking issue for the development of the GitHub Actions Workflows. Take on any issue from the to-do list you like!

**Done**
 - [x] Initial workflow [#871, @EwoutH]
 - [x] Move Style check from Travis to Actions [#897, @1480c1]
 - [x] Create jobs for different compilers (GCC 7, GCC 8, ~~GCC 9~~, Clang) [#914, @ePirat]
 - [x] Implement quick binary check (check for bitstream changes between PR and master) [#947, @ultrawide]
 - [x] Run on different platforms (~~Ubuntu~~, macOS, ~~Windows~~) [#931, @1480c1]
 - [x] Implement caching
 - [x] Implement compiling the FFmpeg and GStreamer patches (#983, @luigi311)
 - [x] Implement Unit Tests (#967, @Cidana-Developers)

**To-do**
 - [ ] Run with different SIMD levels enabled (sse2, ssse3, sse4.1, avx2, avx512)

**Under consideration**
 - Gather and output full encoding quality and performance metrics (see [visual metrics](https://github.com/google/compare-codecs/blob/master/lib/visual_metrics.py), lots of CPU needed)
 - Deploy pre-compiled builds to GitHub (currently AppVeyor does this on tags/releases)
 - Run style test (currently Travis)
 - Run valgrind (currently Travis)

It's possible to create more than one workflow, so maybe we should create workflows for compile, test, etc. and run those on different platforms. GitHub Actions is brand new and all major functions are covered by Travis, so feel free to experiment. 

I found [Workflow syntax for GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions) quite useful.

If you have anything that should be added to the list please comment!

1480c1:
> * Run with different SIMD levels enabled (sse2, ssse3, sse4.1, avx2, avx512)

https://github.com/OpenVisualCloud/SVT-AV1/pull/785

> * Implement caching

https://github.com/actions/cache


I particularly still can't figure out github cache yet

ultrawide:
[...]

PR #931, opened 2020-01-02
Actions/windows: Add MSVC job

1480c1:
Still haven't added any tests or anything else yet

https://github.com/OpenVisualCloud/SVT-AV1/issues/873

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -1,0 +1,34 @@
+name: MSVC builds
+on: [push, pull_request]
+
+jobs:
+  build:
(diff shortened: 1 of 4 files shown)


1480c1:
I think not wip anymore

hassount:
thnx @1480c1 would this be an initial phase and we'd then port the ffmpeg build / windows testing later from Appveyor ?

1480c1:
Yeah, right now I think it might be better to have basic compilation test of msvc and nasm for now until we can figure out if we can do any kind of caching and speed ups

PR #934, opened 2020-01-03
Added -vbv-bufsize to encoder doc. 

klillevo:
This should have been included with the vbv-bufsize code. 
Also corrected minor issue with TargetBitRate not having been updated to new RateControl modes.


hassount:
@anaghdin just making sure the description is accurate

PR #942, opened 2020-01-08
windows: fix assert in GetFileType

xuguangxin:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfiletype
GetFileType needs a HANDLE not a FILE*.


OmranAbdelkafi:
Run some BD-rate/Speed tests : everything look fine

PR #944, opened 2020-01-09
Fix mismatch between AVX512 and AVX2/C eb_av1_inv_txfm2d_add_64x16() kernel

tszumski:
Modify UT to cover mismatch
Fix for linux Release UT that cause: segmentation fault

I do not see any performance change, compared to previous implementation


hassount:
@tszumski, can we rebase this please?

hassount:
@spawlows Would you be able to rebase this branch ?

PR #947, opened 2020-01-10
Add binary identical test for Github Actions

ultrawide:
This pull request addresses one of the items in #873 - Implement quick binary check for bitstream changes between PR and master

### DIFF
--- a/.github/workflows/svt-av1.yml
+++ b/.github/workflows/svt-av1.yml
@@ -44,3 +44,64 @@ jobs:
       run: |
         ./Build/linux/build.sh release
         ./Bin/Release/SvtAv1EncApp -help
+    - name: Compress SVT-AV1 binaries
+      if: matrix.compiler == 'gcc-9'
+      run: tar czf ./svtav1.tar.gz ./Bin/Release/*
+    - name: Upload SVT-AV1 binaries
+      if: matrix.compiler == 'gcc-9'
+      uses: actions/upload-artifact@v1
+      with:
+        name: svtav1
+        path: ./svtav1.tar.gz
+
+  binary-identical-test:
+    runs-on: ubuntu-18.04
+    needs: [build]
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y gcc-9 g++-9 nasm yasm cmake
+    - uses: actions/checkout@v2
+      with:
+        ref: master
+    - name: Setup SVT-AV1 (master)
+      run: |
+        ./Build/linux/build.sh release
+        mkdir $HOME/master
+        mv -t $HOME/master/ ./Bin/Release/*
+    - name: Download SVT-AV1 binaries
+      uses: actions/download-artifact@v1
+      with:
+        name: svtav1
+    - name: Setup SVT-AV1 (current)
+      run: tar xf svtav1/svtav1.tar.gz
+    - name: Download videos
+      run: |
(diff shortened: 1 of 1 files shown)


1480c1:
Perhaps it might be good to either separate out the diff commands or add some echos so that way we know which one actually failed

.github/workflows/svt-av1.yml
```diff
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
```
ePirat: This effectively reverts my change I had done to change this to `ubuntu-18.04`.

ePirat: I don't really mind, if others think this is better for some reason, but it should at least be split into a different commit, I believe.

.github/workflows/svt-av1.yml
```diff
@@ -44,3 +44,60 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y gcc-9 g++-9 nasm yasm cmake
+    - uses: actions/checkout@v2
+      with:
+        ref: master
```
ePirat: I think it would make sense to only run `binary-identical-test` for PRs as when running for master it compares against master, which probably is not useful as it will always be identical?

.github/workflows/svt-av1.yml
ultrawide: Thanks for the input. I didn't realize that this was previously changed to `ubuntu-18.04`. I was thinking that ubuntu-latest would be more robust since we can test against the latest ubuntu build as future releases become available. It would decrease the overhead of updating this parameter in the future. 

I prefer `ubuntu-latest`, but I'm open to whatever is best.

ultrawide: I will revert back to `ubuntu-18.04 `.

.github/workflows/svt-av1.yml
ultrawide: Unless I'm mistaking my interpretation of `on push`, I'm thinking it would be useful for a user to get a quick confirmation that there are no mismatches in other development branches. The `binary-identical-test` routine takes ~3 minutes to run.

ePirat: Indeed, but maybe we can exclude SVT-AV1/master then as comparing against that has no benefit?

.github/workflows/svt-av1.yml
ePirat: The reason for not using latest is that once github changes that, it would silently start using a new version of ubuntu that could cause the CI to fail due to whatever changes that ubuntu version has, and it would be confusing to figure out that its unrelated to other changes in the repo.

ePirat: (This is for the build, for the comparison feel free to use latest)

.github/workflows/svt-av1.yml
ultrawide: Thanks for the suggestion. I believe there is a benefit to ensure that the same bit stream (output) is being generated for master vs. master.

ePirat: Ok, maybe I dont really understand what you are doing here then 😅

ultrawide: No worries :) 

I'll look into it some more. My intention was to port over the same behavior from Travis CI.

hassount:
Are we all good on this PR? can we merge it ?

hassount:
@ePirat @ultrawide good to go  ?

ultrawide:
Yes, I am good with this.

ePirat:
@hassount Looks good to me.

Issue #940, opened 2020-01-08
segmentation fault when enc_mode 3 with vbr enabled


zpdongtencent:
my command line is as following:
./SvtAv1EncApp -i ~/videos/vidyo4_720p_60fps_60f.y4m -enc-mode 3 -rc 1 -b output.ivf

my svt version was v0.8.0 (af16f70)

OmranAbdelkafi:
Fixed at Pull Request #948

PR #948, opened 2020-01-10
fix rate control crash

OmranAbdelkafi:
## Description 

Fixed crash for -rc 1 due to wrong super block size

## Related issue 

issue #940


Source/App/EncApp/EbAppInputy4m.c
```diff
@@ -241,8 +241,7 @@ int32_t read_y4m_frame_delimiter(EbConfig *cfg) {
         return EB_ErrorNone;
     }
 
-    if (EB_STRCMP((const char *)buffer_y4m_header, "FRAME\n") !=0
-        && EB_STRCMP((const char *)buffer_y4m_header, "FRAME\r\n") != 0) {
+    if (EB_STRCMP((const char *)buffer_y4m_header, "FRAME\n") != 0) {
```
hassount: this part would have to be reverted

Issue #567, opened 2019-08-25
libsvtav1 conflicts with other libraries in ffmpeg. Does not run unless they are disabled.


moisespr123:
Hi,

I've been trying to get `libsvtav1` to run on ffmpeg built using the `media-autobuild-suite` build script. 

While it does compiles, it turns out that `libsvtav1` conflicts with `libopencore-amrwb`, `libsvthevc` and `libaom`. Disabling these libraries makes `libsvtav1` encode a video successfully.

## Conflict with `libopencore-amrwb`:
```
H:/repos/media-autobuild_suite/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: H:/repos/media-autobuild_suite/local64/lib/libSvtAv1Enc.a(EbIntraPrediction.c.obj): warning: common of `pred' overridden by definition from H:/repos/media-autobuild_suite/msys64/mingw64/lib/libopencore-amrwb.a(dec_gain2_amr_wb.o)
collect2.exe: error: ld returned 1 exit status
```

## Conflict with `libsvthevc`:
```
H:/repos/media-autobuild_suite/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: H:/repos/media-autobuild_suite/local64/lib/libSvtAv1Enc.a(EbModeDecisionConfiguration.c.obj):EbModeDecisionConfiguration.c:(.text+0x0): multiple definition of `MdcRefinement'; H:/repos/media-autobuild_suite/local64/lib/libSvtHevcEnc.a(EbModeDecisionConfiguration.c.obj):EbModeDecisionConfiguration.c:(.text+0x0): first defined here
[...]

PR #951, opened 2020-01-12
[RFC] consistent namespacing for public API functions

elenril:
sending this PR primarily for gathering feedback on API namespacing, I do not
actually expect it to be merged in its current form.

SVT-AV1 currently uses `eb_` as the prefix for public functions, but I see
multiple problems with it:
-    it is used across several separate projects, causing symbol conflicts
-    it is used for both public and private functions, making it harder to
     distinguish between them
-    as far as I can tell, it's used mainly for historical reasons and it does
     not carry much meaning for the API users

Since the library API is not stable yet, these problems can still be addressed
without breaking compatibility. It thus seems reasonable to pick a prefix that
would be
-    meaningful for the end-users
-    unique for SVT-AV1
-    used for public functions only (it would also be good to pick a consistent
     prefix to be used for non-static private functions, to avoid conflicts with
     static linking, but that's a separate question)
-    ideally short, so it's easy to write
and officially document it as the reserved namespace for this library.

I am proposing `svt_av1_` in this PR. It satisfies all the points above, except
possibly the last one. `svtav1_` is also possible, but starts looking like
character soup, which is IMO not worth it for saving one character.

Thoughts? Opinions? Bikeshed colorscheme preferenes? Do you even see this as
a worthwhile issue to address?

### DIFF
--- a/Source/API/EbSvtAv1Dec.h
+++ b/Source/API/EbSvtAv1Dec.h
@@ -132,7 +132,7 @@ typedef struct EbSvtAv1DecConfiguration {
      * @ *p_app_data     Callback data.
      * @ *config_ptr     Pointer passed back to the client during callbacks, it will be
      *                   loaded with default parameters from the library. */
-EB_API EbErrorType eb_dec_init_handle(EbComponentType **p_handle, void *p_app_data,
+EB_API EbErrorType svt_av1_dec_init_handle(EbComponentType **p_handle, void *p_app_data,
                                       EbSvtAv1DecConfiguration *config_ptr);
 
 /* STEP 2: Set configuration parameters.
@@ -140,7 +140,7 @@ EB_API EbErrorType eb_dec_init_handle(EbComponentType **p_handle, void *p_app_da
      * Parameter:
      * @ *svt_dec_component             Decoder handle.
      * @ *pComponentParameterStructure  Decoder and buffer configurations will be copied to the library. */
-EB_API EbErrorType eb_svt_dec_set_parameter(
+EB_API EbErrorType svt_av1_dec_set_parameter(
     EbComponentType *svt_dec_component,
     EbSvtAv1DecConfiguration *
         pComponentParameterStructure); // pComponentParameterStructure contents will be copied to the library
@@ -149,7 +149,7 @@ EB_API EbErrorType eb_svt_dec_set_parameter(
      *
      * Parameter:
      * @ *svt_dec_component  Decoder handle. */
-EB_API EbErrorType eb_init_decoder(EbComponentType *svt_dec_component);
+EB_API EbErrorType svt_av1_dec_init(EbComponentType *svt_dec_component);
 
 /*!\brief STEP 4: Decodes a frame with associated data. The data in *data
      * should belong to one frame, possibly with sequence header and metadata.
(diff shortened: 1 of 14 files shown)


hassount:
@elenril 
Thank you for the PR and the detailed explanation. 
I agree that changing the prefixes and standardizing them would definitely be beneficial. and doing so now is definitely the best time. 
A few things to keep in mind:
- we probably need to define a prefix for the functions coming from libaom.
- we also need to keep in mind the other svt libraries and work with the maintaining teams with this get done to port the same logic over. 

Now, there will be a major PR moving files to the enc / dec / common directories we can get this work moving soon afterwards.

jbkempf:
> A few things to keep in mind:
>     * we probably need to define a prefix for the functions coming from libaom.

They should not be exposed in the API, and should be hidden from the SO.

hassount:
> > A few things to keep in mind:
> > * we probably need to define a prefix for the functions coming from libaom.
> 
> They should not be exposed in the API, and should be hidden from the SO.

Agreed, that said, when we build a static build and link it to ffmpeg then we get the symbol conflict. do you know another way of solving this apart from re-naming (assuming we can not declare them all as static) ?

jbkempf:
> Agreed, that said, when we build a static build and link it to ffmpeg then we get the symbol conflict.

Conflict with what?

1480c1:
There are currently conflicts with libaom and one other library iirc https://github.com/OpenVisualCloud/SVT-AV1/issues/567

elenril:
Quoting Hassene Tmar (2020-01-13 00:21:59)
> @elenril 
> Thank you for the PR and the detailed explanation. 
> I agree that changing the prefixes and standardizing them would
> definitely be beneficial. and doing so now is definitely the best
> time. 
> A few things to keep in mind:
> - we probably need to define a prefix for the functions coming from libaom.

I imagine those should be treated as any other non-static private
functions and
- not exported from dynamic libraries (as JB already said)
- converted to used some unified private prefix - however that would be a
  separate PR

> - we also need to keep in mind the other svt libraries and work with
> the maintaining teams with this get done to port the same logic over. 

Certainly. I intend to file analogous PRs there as well.

> 
> Now, there will be a major PR moving files to the enc / dec / common
> directories we can get this work moving soon afterwards.

Is there any timeline for that. I would rather not postpone this too
much, especially since it's just a simple rename.

elenril:
Updated PR:
- rebased against current master
- covers both encoder and decoder

-- 
Anton Khirnov

hassount:
@joelsole @kylophone @mafonsoN @andrey-norkin any issues with this PR ?

ibex-are-goats:
Hello, I'm just observing you work, and I hope everyone is staying safe! 

The only thing I will add here is grammatical. In the official documentation and websites that I've seen, whether at Intel.com or Github.com, the name of this project/codebase/tool is SVT-AV1 with a hyphen, not a space. Same with SVT-HEVC. And in programming convention I think the underscore "_" is synonymous with a space " ", not a hyphen.

So I think technically we would be looking at "svt-av1_dec_set_parameter", etc. 

Have a great weekend and keep pushing the state of the art, everyone!

hassount:
> Hello, I'm just observing you work, and I hope everyone is staying safe!
> 
> The only thing I will add here is grammatical. In the official documentation and websites that I've seen, whether at Intel.com or Github.com, the name of this project/codebase/tool is SVT-AV1 with a hyphen, not a space. Same with SVT-HEVC. And in programming convention I think the underscore "_" is synonymous with a space " ", not a hyphen.
> 
> So I think technically we would be looking at "svt-av1_dec_set_parameter", etc.
> 
> Have a great weekend and keep pushing the state of the art, everyone!

@ibex-are-goats 
thank you for your comment. while it is agreeable that such would be more grammatical, in C / Cpp and presumable other languages, the dash is considered a minus sign so we can not use it in variable / function names

hassount:
> @joelsole @kylophone @mafonsoN @andrey-norkin any issues with this PR ?

@joelsole @kylophone @mafonsoN @andrey-norkin any issues with this PR ?

jbkempf:
Yay \o/

EwoutH:
Seems that the CI broke after merging this PR

hassount:
> Seems that the CI broke after merging this PR

it's appveyor that hit the limit, this PR can no break ci as it is just renaming the api calls.

PR #954, opened 2020-01-14
Fix Unit Tests

tszumski:
-eb_av1_highbd_jnt_convolve*_avx2 kernels. Call C implementation for w==2
-eb_av1_highbd_wiener_convolve_add_src_avx2 kernel. Store correct size of register, when w is not divisible by 16
-disable Speed test by default
-cleanup unused UT "HBDVarianceTest/HBDVarianceTest"


hassount:
@mafonsoN @andrey-norkin @joelsole any issues with this ?

PR #955, opened 2020-01-15
Enc common separation

andrey-norkin:
Separation of encoder and common (one more PR is expected)


Source/Lib/Common/ASM_SSE2/CMakeLists.txt
```diff
@@ -30,6 +30,6 @@ file(GLOB asm_includes
     "x64inc.asm"
     "x86inc.asm")
 
-list(REMOVE_ITEM all_files ${asm_includes})
+#list(REMOVE_ITEM all_files ${asm_includes})
```
1480c1: Can I ask why this change was made? The line is present to prevent nasm from failing on windows since it returns 3 if a the resulting object has no symbols

andrey-norkin: IIRC, this line was failing on linux

1480c1: I see, I reverted this line and it worked for me at least, interesting.

Issue #945, opened 2020-01-09
Copy encoding progress style from x264/x265


AtakSnajpera:
This issue is sort of appendix to this one https://github.com/OpenVisualCloud/SVT-AV1/issues/779
started by stax76 

Hi! I'm the author of RipBot264 and I would like to suggest changes to how encoder informs user about his work.

Here is an example style used in x265.
y4m  [info]: 640x272 fps 24000/1001 i420p8 sar 1:1 unknown frame count
raw  [info]: output file: \\XEON-PC\RipBot264temp\job1\Chunks\1.265
x265 [info]: HEVC encoder version 3.2+22-a8a2c4c37267
x265 [info]: build info [Windows][GCC 9.2.0][64 bit] 10bit
x265 [info]: using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2 AVX
x265 [info]: Main 10 profile, Level-2.1 (Main tier)
x265 [info]: Thread pool created using 16 threads
x265 [info]: Slices                              : 1
x265 [info]: frame threads / pool features       : 4 / wpp(17 rows)
x265 [warning]: Source height < 720p; disabling lookahead-slices
x265 [info]: Coding QT: max CU size, min CU size : 16 / 8
x265 [info]: Residual QT: max TU size, max depth : 16 / 1 inter / 1 intra
x265 [info]: ME / range / subpel / merge         : hex / 9 / 2 / 3
x265 [info]: Keyframe min / max / scenecut / bias  : 24 / 240 / 40 / 5.00 
x265 [info]: Lookahead / bframes / badapt        : 20 / 4 / 2
x265 [info]: b-pyramid / weightp / weightb       : 1 / 1 / 0
x265 [info]: References / ref-limit  cu / depth  : 3 / off / on
x265 [info]: AQ: mode / str / qg-size / cu-tree  : 2 / 1.0 / 16 / 1
x265 [info]: Rate Control / qCompress            : CRF-20.0 / 0.60
x265 [info]: tools: rd=3 psy-rd=2.00 early-skip rskip signhide tmvp b-intra
x265 [info]: tools: strong-intra-smoothing deblock sao
                                                                                
x265 [info]: frame I:     23, Avg QP:17.34  kb/s: 655.18  
x265 [info]: frame P:    401, Avg QP:21.92  kb/s: 730.71  
x265 [info]: frame B:   1014, Avg QP:26.79  kb/s: 122.20  
[...]

Issue #779, opened 2019-11-08
poor command line interface


stax76:
There are certain CLI traditions and similarities between x264, x265 and nvencc, following traditions and similarities makes the lives of CLI users and GUI authors easier, knowledge and code could be shared.

1. Executable name should be reasonable short, easy to remember and type, example: git, cmd or wt (new windows terminal)

2. it's a tradition that short switches start with a single dash - and long switches start with double dash -- and there are libraries for parsing, it's OK doing it differently like powershell but in doubt I suggest to follow the tradition.

https://en.wikipedia.org/wiki/Getopt

3. switches should use identical names and parameters as other encoders if possible
https://x265.readthedocs.io/en/latest/cli.html#cmdoption-preset

4. there should be good documentation for both CLI --help, -h and online with anchors, there should be switch type information like int, float etc. and there should be categories like Analysis, Slice Decision, IO, Performance etc.

5. config file and cli parameters should be identical

https://mpv.io/manual/master/#configuration-files

https://github.com/ytdl-org/youtube-dl/blob/master/README.md#configuration

6. switches should have a long version --long and a short version -l and not only one of both.

7. I don't know if it's already supported, y4m input support is important

I'm creator of the oldest and most feature rich encoder GUI staxrip that started 2002 with divx, we've built countless encoder gui's, for av1 so far two, one for aomenc and one for rav1e, both encoders are too slow for end users so it was probably a waste of time, many people request to build a third gui for svt-av1 but I have doubts, a poor cli and poor documentation is an indication that an encoder will fail, maybe nvidia will soon have a better av1 encoder that everybody will use.

Sorry for being so direct and not using more kind language, it's not my strength, due to intuition I'm not very optimistic about svt-av1, I hope that I'm wrong and it will be a success.

1480c1:
> 7\. I don't know if it's already supported, y4m input support is important

y4m is already supported

1480c1:
Afaik, the executable is supposed to be more of an example implementation and this was designed to have the end users use the library directly or through ffmpeg/other wrappers, so probably the design quality of the application was not as high of a priority as much as the quality of the internal encoder itself.

moisespr123:
I honestly don't mind about the command line. I also wrote a GUI for SVT-AV1, and it just work fine the way it is.

luigi311:
[...]

PR #957, opened 2020-01-16
Improve encoder interface

OmranAbdelkafi:
## Description
Format and improve help menu
Change encoding progress from number of frame to percentage of progress
## Performance 
No impact
## Issue related
#779
#945

### DIFF
--- a/Source/App/EncApp/EbAppProcessCmd.c
+++ b/Source/App/EncApp/EbAppProcessCmd.c
@@ -1288,16 +1288,16 @@ AppExitConditionType process_output_stream_buffer(EbConfig *config, EbAppContext
                 progress = ((double)frame_count / config->processed_frame_count) * 100;
                 // if statement used for formatting
                 if (progress < 10) 
-				{ 
-					fprintf(stderr, "\b\b\b\b\b%.2f%%", progress);
-				}
-				else
-				{
+                { 
+                    fprintf(stderr, "\b\b\b\b\b%.2f%%", progress);
+                }
+                else
+                {
                     fprintf(stderr, "\b\b\b\b\b\b%.2f%%", progress);
-				}
+                }
 #endif                
-			}
-				
+            }
+            
 #endif
 
             //++frame_count;
--- a/Source/App/EncApp/EbAppProcessCmd.c
+++ b/Source/App/EncApp/EbAppProcessCmd.c
@@ -25,7 +25,7 @@
     ((((width) * (height)) + 2 * (((width) * (height)) >> (3 - csp))) << is_16bit)
 #define YUV4MPEG2_IND_SIZE 9
 //encoder output
-#define FRAME_OUTPUT 0
+#define FRAME_OUTPUT_ONLY 0
 extern volatile int32_t keep_running;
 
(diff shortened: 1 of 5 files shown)


EwoutH:
Percentage looks nice! I would still like to know the framecount however.

Maybe we can display it as `Encoding   42.00% (420/1000)`?

Source/App/EncApp/EbAppProcessCmd.c
```diff
@@ -1277,7 +1279,24 @@ AppExitConditionType process_output_stream_buffer(EbConfig *config, EbAppContext
 #else
             //++frame_count;
             if (!(header_ptr->flags & EB_BUFFERFLAG_IS_ALT_REF))
+            {
+#if FRAME_OUTPUT
                 fprintf(stderr, "\b\b\b\b\b\b\b\b\b%9d", ++frame_count);
+#else
+                double progress = 0;
+                ++frame_count;
+                progress = ((double)frame_count / config->processed_frame_count) * 100;
```
hassount: processed_frame_count should be frames_to_be_encoded

luigi311:
Should we start changing names to match up with other encoders, such as enc-mode over to preset

silverbacknet:
> 
> 
> Should we start changing names to match up with other encoders, such as enc-mode over to preset that way we dont have to change around the double dash commands any more and break any existing implementations again when/if we decided to switch over to more standard names in the future after this patch has been merged.

aomenc uses "cpu-used" in the same way as svt's enc-mode, while the x26x family uses "preset", so the need for a cpu-used synonym is probably more pressing. It be fairly simple to map a set of superfast-to-placebo presets to enc-modes, the commandline parsing just isn't set up for that at all yet.

kylophone:
My only concern is that command line parsing is still not being done with `getopt`. Is there a reason for this?

luigi311:
We can make it backwards compatible by using the method I specified on #779

kylophone:
> We can make it backwards compatible by using the method I specified on #779

Yeah. We'll have to. Don't publish the old options in the readme or help/usage message, but if someone uses an old option, accept it but also print a loud depreciation warning.

luigi311:
Looks like theres also not a --version option

luigi311:
Heres a quick example of how we can support all 3 commands so we will not break existing scripts but move over to something new. This supports
```bash
SvtAv1EncApp --help   *new long method*
SvtAv1EncApp -h       *new short method*
SvtAv1EncApp -help    *legacy support*
```
https://github.com/luigi311/SVT-AV1/blob/32af03d4638858795e4de2096094356174aabf5c/Source/App/EncApp/EbAppConfig.c#L1633

silverbacknet:
I think going with getopt_long_only() with long & short options duplicated to some single short option would be the way to go. getopt does take care of some of the overhead and would really simplify that file. SVT already doesn't support getopt-style short-args (either combined args or no space between arg and extra data), so making everything a long arg would work fine.

hassount:
@OmranAbdelkafi Closing this for now, we'll need to open a new one for the new interface to be ported from cs2

PR #958, opened 2020-01-16
Add Unit tests nightly runs

ePirat:
This adds unit tests to Github Actions to run each night at midnight (UTC).

Currently some tests fail and the tests take waaaay too long (over 6 hours).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,10 @@
-name: svt-av1
-on: [push, pull_request]
+name: CI
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  push:
+  pull_request:
+
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


ePirat:
I've renamed the CI yml file, this was needed to get the schedule cron to actually work, apparently there is some quirk in Github Actions…

ePirat:
(Marked as draft due to the issues mentioned in the entry post about tests taking too long, as this causes them to time out after 6 hours, which is the maximum limit a job can run on Github actions)

EwoutH:
From the [GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources) documentation:

> Each virtual machine has the same hardware resources available.
> - 2-core CPU
> - 7 GB of RAM memory
> - 14 GB of SSD disk space

If the Unit tests are scaling well over multiple tests, we should consider [hosting a runner ourselves](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/hosting-your-own-runners). That would be useful for gathering BDR metrics with the CI as well.

@kylophone or @hassount could you check if you have a machine available at this time and could hook it up? For now something with 8+ cores and 16+ GB RAM would suffice I think.

Part of #873.

1480c1:
Perhaps it might be better to separate out the cron based jobs into another file like `unit-test.yml`? That way anyone currently modifying the same file wont have a lot of conflicts

ePirat:
@1480c1 of course possible but means we would need to copy paste the build instructions and have them duplicated in several files, which might be a bit annoying to maintain, but I have no strong preference. So whatever works best for you is fine for me.

luigi311:
This is now easier to do and maintain with the ci script once that gets merged.

ePirat:
This is already done in a different PR

PR #962, opened 2020-01-16
Appveyor: Fix logic error in pushing an artifact

1480c1:
Used to using `-ne` to test if a variable present, not backwards.


hassount:
@1480c1 ready ?

1480c1:
Yeah

PR #921, opened 2019-12-23
support hierarchical prediction structure level 5.

jerrymiao-tencent:

PR #968, opened 2020-01-19
Support manual prediction structure

elintencent:
This PR implements manual prediction structure, user can specify the prediction structure via the configuration file with command line "-pred-struct-file xx/PredStruct_level0.cfg".

Some example prediction structure configuration files are placed at SVT-AV1\Config\PredStruct_level0.cfg~PredStruct_level5.cfg, which match the fixed prediction structure in current code. 

When prediction structure configration file is used, hierarchical_levels value in command line is ignored, and will be set the actual size automatically accoring to the prediction structure configuration file.

There are still some restrictions on this manual prediction structure due to the current SVT_AV1 framework:
1. Only hierarchical prediction structure is supported
2. The maximum levels supported is 5 (minigop 32)
3. Some other parameters with specific layers not related to the prediction structure (e.g. lambda/qp...) are not added yet
4. In current SVT_AV1, only hierarchical_levels 3 and 4 are allowed to be used in verify_settings(), don't know why levels 0/1/2 can't be used with the internal code being ready(#973 ), and level 5 PR(#921 ) hasn't been merged, so manual prediction structure is also affected by this restriction for now until the above issues are fixed.

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -34,6 +34,7 @@
 #define INPUT_STAT_FILE_TOKEN "-input-stat-file"
 #define OUTPUT_STAT_FILE_TOKEN "-output-stat-file"
 #define STAT_FILE_TOKEN "-stat-file"
+#define INPUT_PREDSTRUCT_FILE_TOKEN "-pred-struct-file"
 #define WIDTH_TOKEN "-w"
 #define HEIGHT_TOKEN "-h"
 #define NUMBER_OF_PICTURES_TOKEN "-n"
@@ -195,6 +196,12 @@ static void set_cfg_input_file(const char *filename, EbConfig *cfg) {
 #endif
 
     cfg->y4m_input = check_if_y4m(cfg);
+};
+
+static void set_pred_struct_file(const char *value, EbConfig *cfg) {
+    if (cfg->input_pred_struct_file) { fclose(cfg->input_pred_struct_file); }
+    FOPEN(cfg->input_pred_struct_file,value, "rb");
+    cfg->enable_manual_pred_struct = EB_TRUE;
 };
 
 static void set_cfg_stream_file(const char *value, EbConfig *cfg) {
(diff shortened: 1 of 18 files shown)


elintencent:
The failed binary-identical-test can be successful on my machine with the same command line.
image

elintencent:
Don't know why encode videos (master) in binary-identical-test failed which uses the original svt_av1 version without my commit，but encode videos (current) in binary-identical-test succeeded which uses my commit.

hassount:
> Don't know why encode videos (master) in binary-identical-test failed which uses the original svt_av1 version without my commit，but encode videos (current) in binary-identical-test succeeded which uses my commit.

I believe that this is a known issue

anaghdin:
Does using the same prediction structure through manual configuration produces the same streams?Have you tried other manual configurations? How do you determine the hidden frames and show existing frames?

elintencent:
> Does using the same prediction structure through manual configuration produces the same streams?Have you tried other manual configurations? How do you determine the hidden frames and show existing frames?

@anaghdin ,thanks for your questions, please find the answers as follows:

1. Using the same prediction structure through manual configuration doesn't produce the same streams with that are produced by previous fixed structures, because the manual scheme finds the releasable position dynamically for each frame, so the dpb index written into stream may be different, but the yuv decoded from them should be bit identical.

2. Other manual configurations can also be used, but only level3/4 can be used for now as I mentioned in PR description.

3. Manual scheme records the last displayed picture number and find whether there is a previous frame can be displayed in dpb list after encoding each frame, if does, it will be set as show_existing_frame. only hierarchical prediction structure is allowed to be used for now, which assures that there is only one show_existing_frame each time.

4. If one frame is ALT and overlay is enabled (pcs_ptr->is_alt_ref==1), the frame is not displayed, which is same as the fixed prediction structure scheme.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -1407,6 +1415,154 @@ int32_t compute_frames_to_be_encoded(EbConfig *config) {
...
+    char varName[CONFIG_FILE_MAX_VAR_LEN];
+    char varValue[CONFIG_FILE_MAX_ARG_COUNT][CONFIG_FILE_MAX_VAR_LEN];
+
+    uint32_t valueIndex;
+
+    uint32_t commentSectionFlag = 0;
+    uint32_t newLineFlag = 0;
+    int32_t  entry_num = 0;
+    int32_t  display_order = 0, num_ref_list0 = 0, num_ref_list1 = 0;
+    int32_t  idx_ref_list0 = 0, idx_ref_list1 = 0;
```
hassount: snake_case

elintencent: > snake_case throughout the added code for variables

@hassount ,thanks for your reminder, but I found all the other code in svt_av1 have been converted from camel_case to snake_case format, so snake_case should be used now, right?

elintencent: > > snake_case throughout the added code for variables
> 
> @hassount ,thanks for your reminder, I found all the other code in svt_av1 have been converted from camel_case to snake_case format, so snake_case should be used now, right?

hassount: correct

elintencent:
Updated svt-av1_encoder_user_guide.md for manual prediction structure configuration.

hassount:
@mafonsoN @andrey-norkin can we proceed with this ?

PR #970, opened 2020-01-20
Fix the tmvp issue

lijing0010:
## Description

this function is meant to save the current mv set and will be used as tmvp for following frames.
However, it overwritten the data, and it will cause some mistakes in some rare cases.
For example, if sb size is 64x64. if we process (0,256) first, and (576, 0) later. the mv of block starting at (0,256) may be overwritten by the mv from (576, 0).
It will affect the mode decision a bit.
For multi-tile case, it is more easy to trigger, and may also affect the entropy part(when building the tmvp stack), and cause corrupt bitstreams


mafonsoN:
@lijing0010 can you give a small description of the issue and fix? Thanks

lijing0010:
@mafonsoN , this function is meant to save the current mv set and will be used as tmvp for following frames. 
However, it overwritten the data, and it will cause some mistakes in some rare cases.
For example, if sb size is 64x64. if we process (0,256) first, and (576, 0) later. the mv of block starting at (0,256) may be overwritten by the mv from (576, 0).
It will affect the mode decision a bit.
For multi-tile case, it is more easy to trigger, and may also affect the entropy part(when building the tmvp stack), and cause corrupt bitstreams

mafonsoN:
Thanks @lijing0010. Please go ahead @hassount

PR #971, opened 2020-01-20
Enable tile parallelism

lijing0010:
## Description

-Enabled tile group parallelism at encDec kernel, now set tile group the same as tile row
-Enabled tile parallelism at enctropy kernel
-Fixed a bug during saving the mvs
-Enabled palette for tile parallelism at entropy kernel
-Improved the tile_size_bytes_minus_1 field

### DIFF
--- a/Source/Lib/Common/Codec/EbPictureBufferDesc.c
+++ b/Source/Lib/Common/Codec/EbPictureBufferDesc.c
@@ -229,8 +229,6 @@ void link_eb_to_aom_buffer_desc_8bit(EbPictureBufferDesc *picBuffDsc,
     }
 }
 
-//Jing: TODO
-//Change here later
 void link_eb_to_aom_buffer_desc(EbPictureBufferDesc *picBuffDsc, Yv12BufferConfig *aomBuffDsc) {
     //NOTe:  Not all fileds are connected. add more connections as needed.
     if (picBuffDsc->bit_depth == EB_8BIT) {
--- a/Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
+++ b/Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
@@ -29,7 +29,23 @@ typedef struct PictureManagerContext {
     EbFifo *picture_control_set_fifo_ptr;
 } PictureManagerContext;
 
-void                    set_tile_info(PictureParentControlSet *pcs_ptr);
+#if TILES_PARALLEL
+// Token buffer is only used for palette tokens.
+static INLINE unsigned int get_token_alloc(int mb_rows, int mb_cols, int sb_size_log2,
+                                           const int num_planes) {
+    // Calculate the maximum number of max superblocks in the image.
+    const int shift          = sb_size_log2 - 4;
+    const int sb_size        = 1 << sb_size_log2;
+    const int sb_size_square = sb_size * sb_size;
+    const int sb_rows        = ALIGN_POWER_OF_TWO(mb_rows, shift) >> shift;
+    const int sb_cols        = ALIGN_POWER_OF_TWO(mb_cols, shift) >> shift;
+
+    // One palette token for each pixel. There can be palettes on two planes.
+    const int sb_palette_toks = AOMMIN(2, num_planes) * sb_size_square;
+
+    return sb_rows * sb_cols * sb_palette_toks;
+}
+#endif
 extern MvReferenceFrame svt_get_ref_frame_type(uint8_t list, uint8_t ref_idx);
 /************************************************
  * Defines
@@ -803,6 +819,9 @@ void *picture_manager_kernel(void *input_ptr) {
                         child_pcs_ptr->parent_pcs_ptr->sad_me         = 0;
                         child_pcs_ptr->parent_pcs_ptr->quantized_coeff_num_bits = 0;
                         child_pcs_ptr->enc_mode = entry_pcs_ptr->enc_mode;
+#if TILES_PARALLEL
+                        child_pcs_ptr->enc_dec_coded_sb_count = 0;
+#endif
 
                         //3.make all  init for ChildPCS
                         pic_width_in_sb = (uint8_t)((entry_scs_ptr->seq_header.max_frame_width +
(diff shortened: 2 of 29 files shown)


hassount:
@lijing0010 
This solves the Tile corruption in the master thank you. 
Does this PR add any performance change, or is it just an infrastructure with bug fixes ?

lijing0010:
@hassount , it will not affect the performance(CPU cycles). But can increase the parallelism if using multiple tiles.
For non-tile case, it's the same as master.
For multiple-tile case, I enabled a logic "tile group", if set tile_group_rows/tile_group_columns as 1x1, it will behave similar as original master, tiles in encDec are processed in serial. (tiles will be processed in parallelism at entropy stage, different from master).

Source/Lib/Common/Codec/EbPictureBufferDesc.c
```diff
@@ -229,8 +229,6 @@ void link_eb_to_aom_buffer_desc_8bit(EbPictureBufferDesc *picBuffDsc,
     }
 }
 
-//Jing: TODO
-//Change here later
 void link_eb_to_aom_buffer_desc(EbPictureBufferDesc *picBuffDsc, Yv12BufferConfig *aomBuffDsc) {
```
mafonsoN: Do you want to keep this comment here?

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
```diff
@@ -812,7 +832,117 @@ void *picture_manager_kernel(void *input_ptr) {
...
+                        uint16_t tile_row, tile_col;
+                        uint32_t  x_sb_index,  y_sb_index;
+                        const int tile_cols = ppcs_ptr->av1_cm->tiles_info.tile_cols;
+                        const int tile_rows = ppcs_ptr->av1_cm->tiles_info.tile_rows;
+                        TileInfo tile_info;
+                        uint8_t tile_group_cols = MIN(tile_cols, entry_scs_ptr->tile_group_col_count_array[entry_pcs_ptr->temporal_layer_index]);
+                        uint8_t tile_group_rows = MIN(tile_rows, entry_scs_ptr->tile_group_row_count_array[entry_pcs_ptr->temporal_layer_index]);
+
+                        if (tile_group_cols * tile_group_rows > 1) {
+                            //Jing: TODO: Tuning segments number for each tile later
```
mafonsoN: You have a lot of TODO comments. Were these all addressed or are these tasks for a future PR? Perhaps it would be also best to remove your name from them.

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
```diff
@@ -812,7 +832,117 @@ void *picture_manager_kernel(void *input_ptr) {
...
+                        }
+
+                        for (uint8_t r = 0; r < tile_group_rows; r++) {
+                            for (uint8_t c = 0; c < tile_group_cols; c++) {
+                                uint16_t tile_group_idx = r * tile_group_cols + c;
+                                //uint16_t top_left_tile_idx = tile_group_col_start_tile_idx[c]
+                                //    + tile_group_row_start_tile_idx[r] * tile_cols;
+                                //uint16_t bottom_right_tile_idx = (tile_group_col_start_tile_idx[c + 1] - 1)
+                                //    + (tile_group_row_start_tile_idx[r + 1] - 1) * tile_cols;
+                                uint16_t top_left_tile_col_idx = tile_group_col_start_tile_idx[c];
```
mafonsoN: Perhaps remove the commented lines of code

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
```diff
@@ -812,7 +832,117 @@ void *picture_manager_kernel(void *input_ptr) {
...
+                                //printf("---TileGroupIdx %d,SB TL (%d, %d), BR (%d, %d), ColXRow %dX%d, Tile TL(%d,%d), BR(%d,%d)\n",
+                                //    tile_group_idx, tg_info_ptr->tile_group_sb_start_x, tg_info_ptr->tile_group_sb_start_y,
+                                //    tg_info_ptr->tile_group_sb_end_x, tg_info_ptr->tile_group_sb_end_y,
+                                //    tg_info_ptr->tile_group_width_in_sb, tg_info_ptr->tile_group_height_in_sb,
+                                //    tg_info_ptr->tile_group_tile_start_x,
+                                //    tg_info_ptr->tile_group_tile_start_y,
+                                //    tg_info_ptr->tile_group_tile_end_x,
+                                //    tg_info_ptr->tile_group_tile_end_y);
+
+                                // Init segments within the tile group
```
mafonsoN: Commented lines.

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
```diff
@@ -812,7 +832,117 @@ void *picture_manager_kernel(void *input_ptr) {
...
+
+                        // Get the tile start index for tile group
+                        for (uint8_t c = 0; c <= tile_group_cols; c++) {
+                            tile_group_col_start_tile_idx[c] = c * tile_cols / tile_group_cols;
+                        }
+                        for (uint8_t r = 0; r <= tile_group_rows; r++) {
+                            tile_group_row_start_tile_idx[r] = r * tile_rows / tile_group_rows;
+                        }
+
+                        for (uint8_t r = 0; r < tile_group_rows; r++) {
```
mafonsoN: You use iterators r and c and below tr and tc. Maybe it would be best to have them consistent. Also, these seem like too many looks over the same ranges. Can this be simplified?

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
lijing0010: OK, I'll remove my names.
Those places are somewhere we can tune for speed for certain use cases

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
lijing0010: OK

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
lijing0010: Done

Source/Lib/Encoder/Codec/EbPictureManagerProcess.c
lijing0010: Change tr/tc to r/c.
There are two different tile related things here.
One is a virtual "tile group", another is tiles.
A tile group can be consisted by NXM tiles, and segments are parallelized within a tile_group.
In entropy kernel, it's parallelized within tiles.

For example, if the input tiles are 4x4, total 16 tiles.
I don't want to start the encdec kernel in 16 tiles spontaneously, that will be very inefficient, the latency will not be good. But I can divide the 4x4 tiles into a 1x4 tile group, will help to reduce the latency.
At entropy stage, all recon dependency are already resolved before entropy kernel, so different tiles can be started spontaneously, it will help to reduce the latency, especially when the bitrate is high

PR #975, opened 2020-01-21
Travis CI: Add build jobs on other CPU architectures

EwoutH:
Add builds jobs for 64-bit ARMv8, PowerPC (64-bit, little-endian) and IBM Z. All on Linux with GCC 9. Are allowed to fail (for now).

For general distribution SVT-AV1 has to be able to be compiled on different CPU architectures then x86. It doesn't need to be fast or optimized, it just needs to compile and run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,9 @@ matrix:
     - name: Coveralls osx+clang
     - name: Unit Tests Linux+gcc
     - name: Unit Tests osx+clang
+    - arch: arm64
+    - arch: ppc64le
+    - arch: s390x
     # Exclude these because if the encoder can run with a release build, the commit is probably fine. Also required for fast_finish.
   include:
     # GCC & Clang builds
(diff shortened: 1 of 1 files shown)


1480c1:
I guess first would be a way to enable compiling c code only

EwoutH:
@1480c1 You're fast!

Was just going to ask, there seems to be some SSE2 assembly that get's compiled no matter the value of `C_FLAG_msse2`. What's the best way to fix that?

1480c1:
FIrst would be to do something in the CMakeLists to have an `option` like `ENABLE_<SIMD>` or something that will selectively disable building specific simd code, then add something else that will make sure the code doesn't rely on the simd instruction based code during compile

EwoutH:
In that case we need an `ENABLE_<SIMD>` option for each SIMD instruction set right? So `ENABLE_SSE2`, `ENABLE_AVX2`, etc.

I there an options with CMake to only compile the supported SIMD extensions by the host processor?

1480c1:
Not right now except for making generalization of checking the cpu arch and enabling on x86_64, but I was thinking about setting them to either user inputted or tying them to if the compiler supports `-m<simd>` and the appropriate symbols are there

EwoutH:
Do you have time to give that a try? As you can see I'm not that familiar with CMake. :)

1480c1:
Perhaps when I have some time

hassount:
so, is this PR to be merged now or until we fix the cmake issue ?

1480c1:
It would probably be easiest to go ahead and merge this

EwoutH:
@hassount Rebased!

PR #983, opened 2020-01-27
Actions: Add ffmpeg and gstreamer build

luigi311:
SVT-AV1 Actions:

- Overall
    - Add ccache to all jobs
    - Add native_deps
    - Add ci script that will handle building svt-av1
    - Always apt-get update prior to install
    - Remove fetch-depth: 1 as that is now the default in checkout v2

- Build
    - Separated the env for matrix for easier reading
- Gstreamer
    - Add support for gstreamer building
- FFMPEG
    - Add support for ffmpeg building with patch
- Binary test
    - Pull from official master and not fork master
    - Remove ref: master since that is the default
    - Add continue-on-error on diff, this will mark the test with a checkmark but when clicked it will show the error. It will still upload the artifact so if the artifact exists then it errored. This is due to https://github.com/OpenVisualCloud/SVT-AV1/issues/873#issuecomment-590451066

Referencing ticket #873


.github/workflows/svt-av1.yml
```diff
@@ -106,3 +108,27 @@ jobs:
...
+          sudo apt-get install -y gcc-9 g++-9 nasm yasm cmake libgstreamer-plugins-base1.0-dev libgstreamer1.0-dev
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Compile SVT-AV1
+        run: |
+          mkdir -p Build/linux/${build_type}
+          cd Build/linux/${buildtype}
+          cmake $GITHUB_WORKSPACE -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type}
+          cmake -j${nproc} --build . && cd $GITHUB_WORKSPACE
```
1480c1:
```suggestion
          cmake -j${nproc} --build .
          sudo cmake -j${nproc} --build . --target install && cd $GITHUB_WORKSPACE
```

luigi311: Will this fix the gstreamer plugin from not finding the package?

luigi311: I switched it over and it looks like --target install is throwing an error.
```bash
CMake Error: The source directory "/home/runner/work/SVT-AV1/SVT-AV1/Build/linux/release/install" does not exist.
```

1480c1: Perhaps switch to `sudo cmake --build . --target install`

luigi311: I switched it over to --target install prior to the --build and that fixed it. Meson still can not find it, on my local machine it installs as SvtAv1EncApp which meson looks like it is searching for SvtAv1Enc without the App at the end. I wonder how travis is finding it.

luigi311: looks like the problem was the -j.

luigi311:
@1480c1 
It doesnt look like caching is doing anything it still takes 7 minutes to build ffmpeg. I am caching a few folders to try it out such as libavcodec and it is still compiling all those files again from scratch.

1480c1:
Perhaps try using ccache and caching `$HOME/.ccache`

luigi311:
> 
> 
> Perhaps try using ccache and caching `$HOME/.ccache`

ccache was not installed so i installed it and set it up but it looks like it is not being used at all. If you look at the end of the build ffmpeg stage i have it do another ccache -s to spit out information and it is returning 0 as the cache size. Can you give it a look over and see if i am missing anything?  
https://github.com/OpenVisualCloud/SVT-AV1/runs/413271979?check_suite_focus=true

1480c1:
export CC as `ccache $CC` or `ln -s` ccache to `$CC` and set that symlinked `$CC` as first in `PATH` so it will pick up ccache(using the `$CC` as it's name) first before the actual compiler

luigi311:
> export CC as `ccache $CC` 

I did a export CC="ccache $CC" at the beginning and it is still returning 0 files in cache. I also tried CC="ccache $CC" ./configure --enable-libsvtav1 and still nothing.

1480c1:
Try setting CC to gcc or using "ccache gcc" instead since you haven't defined previously CC

1480c1:
Actually, use `--cc="ccache gcc" --cxx="ccache g++"` since I'm not sure if ffmpeg adheres to CC

luigi311:
Looks like that was it. I added the --cc and the --cxx to the compiler and it is now using ccache. What i dont understand is why the svt-av1 compilation was not using the CC variable. Are we also not using CC in our cmake?

1480c1:
cmake should pull `$CC` as `CMAKE_C_COMPILER`, but I don't think exporting in one step will affect the other, you need to set CC in under env: or [set-env](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#set-an-environment-variable-set-env)

luigi311:
You sir are a scholar. You have so far solved all my issues lol thanks for the help. Im going to mess around with the cache and some formatting before i am ready to get this reviewed. So far in the last few days of messing with github actions for the first time i have learned a good amount.

luigi311:
So setting the env: CC="ccache gcc" breaks the ffmpeg build even though its not even using CC. Svt-av1 builds successfully with it though lol

luigi311:
This is all set

1480c1:
@luigi311, can you rebase on the master branch and squash as much of the commits as possible?

If not, one thing I often do is
```bash
git rebase upstream/master
git push -f

git reset upstream/master
git add -p
git commit
git add -p
git commit
# ... repeat git add -p and git commit for each commit section you want to do
# git reset -p to unstage a section
# Use s, y, and n to split a diff, add a diff, and n to skip a diff when using git add -p
git diff origin/actions # to make sure you aren't introducing or missing something
git push -f
```

luigi311:
Ohh man this is going to be fun

1480c1:
Nice

luigi311:
Aright everything is all set on my end

.github/workflows/svt-av1.yml
```diff
@@ -1,48 +1,70 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y ${{ matrix.packages }} ${native_deps}
     - uses: actions/checkout@v2
       with:
         fetch-depth: 1
+    - name: Set timestamp
+      id: build_timestamp
+      run: |
+        echo "::set-output name=timestamp::$(date "+%d/%m/%y-%H:%M:%S")"
```
ePirat: What does this do?

ePirat:
I believe it might be better to split the ffmpeg and gstreamer stuff into a different github action yml file.

.github/workflows/svt-av1.yml
```diff
@@ -106,3 +128,107 @@ jobs:
...
+          cd $GITHUB_WORKSPACE/Build/linux/${build_type:-Release}
+          cmake $GITHUB_WORKSPACE -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release}
+          cmake -j${nproc} --build .
+          sudo cmake --build . --target install
+          cd $GITHUB_WORKSPACE
+      - name: Build Gstreamer
+        run: |
+          sudo chown -R $USER: $HOME/.ccache
+          export LD_LIBRARY_PATH+=":/usr/local/lib"
+          export PATH=$PATH:~/.local/bin
```
ePirat: This is inconsistent, the line before you do `+=` and here the `$PATH:…` style.
What is in `~/.local/bin` that this is needed?

.github/workflows/svt-av1.yml
```diff
@@ -106,3 +128,107 @@ jobs:
...
+              ${{ runner.os }}-gstreamer-
+      - name: Compile SVT-AV1
+        env:
+          CC: 'ccache gcc'
+          CXX: 'ccache g++'
+        run: |
+          sudo chown -R $USER: $HOME/.ccache
+          mkdir -p $GITHUB_WORKSPACE/Build/linux/${build_type:-Release}
+          cd $GITHUB_WORKSPACE/Build/linux/${build_type:-Release}
+          cmake $GITHUB_WORKSPACE -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release}
```
ePirat: Why do you invoke CMake in such a complex way using env variable replacements that are not needed here? Or maybe they are set somewhere and I've overlooked it?

.github/workflows/svt-av1.yml
```diff
@@ -106,3 +128,107 @@ jobs:
...
+          cmake -j${nproc} --build .
```
ePirat: Shouldn't this be `$(nproc)`?

.github/workflows/svt-av1.yml
```diff
@@ -106,3 +128,107 @@ jobs:
...
+          sudo cmake --build . --target install
+          cd $GITHUB_WORKSPACE
```
ePirat: This seems not needed? The next step will anyway be in a different state.

.github/workflows/svt-av1.yml
```diff
@@ -106,3 +128,107 @@ jobs:
...
+        run: |
+          sudo chown -R $USER: $HOME/.ccache
+          cd $HOME/ffmpeg
+          git apply $GITHUB_WORKSPACE/ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
+          export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:/usr/local/lib/pkgconfig
+          export CFLAGS=""
+          ./configure --enable-libsvtav1 --cc="ccache gcc" --cxx="ccache g++" || less ffbuild/config.log
+          make --quiet -j$(nproc)
+          sudo make install
+          ccache -s
```
ePirat: I believe it would be a bit nicer to split this in more than one step.

.github/workflows/svt-av1.yml
```diff
@@ -106,3 +128,107 @@ jobs:
...
+        run: |
+          cd $HOME
+          git clone https://github.com/FFmpeg/FFmpeg --branch release/4.2 --depth=1 ffmpeg
+      - name: Build ffmpeg
+        run: |
+          sudo chown -R $USER: $HOME/.ccache
+          cd $HOME/ffmpeg
+          git apply $GITHUB_WORKSPACE/ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
+          export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:/usr/local/lib/pkgconfig
+          export CFLAGS=""
```
ePirat: Is this needed?

.github/workflows/svt-av1.yml
luigi311: This is used to create a variable for the current timestamp. I tried to create it once at the top and call it but it failed and instead used the variable name. The reason you need it is so the cache is always updated with the newest build. If the name is the same as previous cache keys it will say it's the same and not update the cache. Some people use a hash for all the files.

ePirat: Thanks, maybe add a comment about this as it is really not easy to understand, imo.
Can you link me to docs about this? Seems interesting and useful.

.github/workflows/svt-av1.yml
luigi311: This is taken from the gstreamer Travis build. Unfortunately I do not have experience with gstreamer so it was most just a port over. I can try and get this to be more consistent. Is there a method you prefer?

ePirat: export `VAR="${VAR}:…"` as it is not bash specific, unlike `+=`

.github/workflows/svt-av1.yml
luigi311: Are you referring to the -G generator? That was also from Travis. I am not sure if anyone is using that at all even in Travis but it doesn't really do any harm and it will be ready to create an anchor base script like we do in Travis once GitHub actions supports it.

.github/workflows/svt-av1.yml
luigi311: You are correct. With all the damn {} I got so used to using it that I misplace it. Thanks for double checking

.github/workflows/svt-av1.yml
ePirat: Ok

.github/workflows/svt-av1.yml
luigi311: https://github.com/actions/cache/blob/master/examples.md#using-multiple-systems-and-npm-config
This uses the same command to dynamically get the directory for npm to use in the cache.
Here is the original article i read that utilized something similar that i simplified
https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/

.github/workflows/svt-av1.yml
luigi311: That is very true i will go ahead and switch everything over to that

.github/workflows/svt-av1.yml
luigi311: Which part? You included the entire ffmpeg-patch job and a couple more lines from the gstreamer patch

.github/workflows/svt-av1.yml
luigi311: The cmake or the cd? The cmake --target install is from the "base_script". The cd might not have a point though. I can go ahead and remove it

.github/workflows/svt-av1.yml
ePirat: The line I comment on is always the one right above the comment, Github makes it so that it shows more lines above for context, or something…
So specifically the `export CFLAGS=""`

luigi311: Gotcha I did not know that. Thats used to clear any CFLAGS that are set in case someone makes a CFLAG globally thinking only SVT-AV1 is going to use it and might cause issues in the ffmpeg build.

.github/workflows/svt-av1.yml
luigi311: That was mostly being used for testing but it does make more sense to put it in a different step. Do you think its worth even keeping around? I can just toss it since we know the ccache is being utilized now.

ePirat: Ok, removing it sounds good. Maybe split the configure part in a different step though, so if configure fails for some reason, it's easier to spot immediately and have a nicer separated  log?

luigi311:
The problem with splitting the ffmpeg and the gstreamer patches to another workflow item is that we can no longer use the needs function. There is not really a point in running the patches if svt-av1 is broken and does not build.

.github/workflows/svt-av1.yml
luigi311: That makes sense to me

ePirat:
Indeed, let's keep it together for now then.

ePirat:
Ok everything resolved from my side, looks good! Thanks!

.github/workflows/svt-av1.yml
luigi311: I see what you mean by this. Each step gets a "fresh start" and are cd back into the $GITHUB_WORKSPACE i did not know that was a thing.

luigi311:
No problem thanks for the feedback it is always appreciated

luigi311:
This should be all set now

hassount:
can we rebase this PR ?

luigi311:
Rebase has been completed.

luigi311:
Looks like svt-av1 was renamed to ci. I will redo this based on the ci.yml file

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+[ -d "$HOME/.ccache" ] && sudo chown -R $USER: $HOME/.ccache
+export parent_dir=$(pwd)
```
1480c1:
```suggestion
export parent_dir=$PWD
```

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
...
+mkdir -p $parent_dir/Build/linux/${build_type:-Release}
```
1480c1:
```suggestion
mkdir -p $parent_dir/Build/linux/${build_type:=Release}
```

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
...
+cd $parent_dir/Build/linux/${build_type:-Release}
+cmake $parent_dir -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release} -DBUILD_SHARED_LIBS=${shared_libs:-ON} -DBUILD_TESTING=${testing:-OFF} ${CMAKE_EFLAGS}
```
1480c1:
```suggestion
cmake -S $parent_dir -B $parent_dir/Build/linux/$build_type -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=$build_type -DBUILD_SHARED_LIBS=${shared_libs:-ON} -DBUILD_TESTING=${testing:-OFF} ${CMAKE_EFLAGS}
```

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+[ -d "$HOME/.ccache" ] && sudo chown -R $USER: $HOME/.ccache
+export parent_dir=$(pwd)
+mkdir -p $parent_dir/Build/linux/${build_type:-Release}
+cd $parent_dir/Build/linux/${build_type:-Release}
```
1480c1:
```suggestion
```

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
...
+cmake $parent_dir -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=${build_type:-Release} -DBUILD_SHARED_LIBS=${shared_libs:-ON} -DBUILD_TESTING=${testing:-OFF} ${CMAKE_EFLAGS}
+cmake -j$(nproc) --build .
```
1480c1:
```suggestion
cmake -j$(nproc) --build $parent_dir/Build/linux/$build_type
```

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
...
+sudo cmake --build . --target install
```
1480c1:
```suggestion
sudo cmake --build $parent_dir/Build/linux/$build_type --target install
```

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
...
+cd $parent_dir
```
1480c1:
```suggestion
```

.ci/ci.sh
```diff
@@ -0,0 +1,11 @@
...
+$parent_dir/Bin/${build_type:-Release}/SvtAv1EncApp -help
```
1480c1:
```suggestion
$parent_dir/Bin/$build_type/SvtAv1EncApp -help
```

luigi311:
${var:='value'} instead of ${var:-'value'} when you are reusing the same variable multiple times is good to know ill be sure to remember that.

luigi311:
@ultrawide 
Do you think there will be an issue with copy instead of mv for the master branch? It looks like the cache action generates the hash every single time it has to use it so when it generates the hash for uploading the cache it errors out since we were moving everything. Does the binary check look ok before i go ahead and break it by pointing it to the build script?

luigi311:
Got it working with the mv command by storing the hash to a variable prior to moving everything. This is all set now. I am going to go ahead and tell it to use the ci script to build the svt-av1 master. It will fail until this PR is merged.

luigi311:
I will instead not include the build script and leave it build with the ./Build/linux/build.sh static script and then make a new PR to move it over to the new ci script.

luigi311:
This is now ready to be merged

.github/workflows/ci.yml
```diff
@@ -1,50 +1,70 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y ${native_deps} ${{ matrix.packages }}
     - uses: actions/checkout@v2
+    - name: Cache ccache files
+      uses: actions/cache@v1.1.0
       with:
-        fetch-depth: 1
+        path: ~/.ccache
+        key: ${{ runner.os }}-svt-av1-${{ matrix.compiler }}-${{ hashFiles('**/*.c') }}
```
ePirat: Doesn't the hashFiles destroys the whole purpose of using ccache here?

.github/workflows/ci.yml
```diff
@@ -108,53 +146,165 @@ jobs:
         diff $HOME/test-pr-10bit-m8.ivf $HOME/test-master-10bit-m8.ivf
         echo Comparing 10-bit M0: Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m
         diff $HOME/test-pr-10bit-m0.ivf $HOME/test-master-10bit-m0.ivf
+        # Create variable that job succeded
+        # if it does not exist then the job failed and artifact will be uploaded
+        echo ::set-output name=status::success
     - name: On fail, compress SVT-AV1 (master)
-      if: failure()
+      if: steps.binary_compare.outputs.status != 'success'
```
ePirat: Instead of this it seems much easier to just combine the compression and upload steps and then use:

```
if: failure()
```

for the compression and upload step.

.ci/ci.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+[ -d "$HOME/.ccache" ] && sudo chown -R $USER: $HOME/.ccache
+export parent_dir=$PWD
+mkdir -p $parent_dir/Build/linux/${build_type:=Release}
+cmake -S $parent_dir -B $parent_dir/Build/linux/$build_type -G"${generator:-Unix Makefiles}" -DCMAKE_BUILD_TYPE=$build_type -DBUILD_SHARED_LIBS=${shared_libs:-ON} -DBUILD_TESTING=${testing:-OFF} ${CMAKE_EFLAGS}
```
ePirat: IMO using env variables like this is not really nice. Flags for the script would be cleaner. But probably okay for this initial version and can be improved later as this is anyway only intended to be used by CI…

.github/workflows/ci.yml
ultrawide: I had the same thoughts, but the reason the steps are separated is because a step cannot have both the uses and run key

.github/workflows/ci.yml
luigi311: hashfiles will give the hash of all the current C files, it will use that as the name for the cache, if the hash does not exist it will instead pull in the last cache that was uploaded and then upload the new cache under the new hash name. The name has to be unique for each new build or else the cache will never get updated because when there is a "cache hit" where it finds the name in the list of caches it will save it will not upload because it thinks its the exact same cache. You need to give it a unique name, before i was doing a timestamp so the name would always be different but then you get a new cache for every single run where this will only upload a new cache when the actual C files have changed.

luigi311:
Looks like you are right. It is cloning it and then doing a merge with the PR. What did you change on your test? I dont see anything different in the ci.yml

ultrawide:
I haven't made changes to ci.yml. It's the behvaiour I observed.

.github/workflows/ci.yml
luigi311:
```
if: failure()
```
Does not work any more because the binary test will never fail now due to the continue-on-error. 
This is what it would do on a error but with the if: failure() check 
https://github.com/luigi311/Testing-Github-Actions/runs/465602449?check_suite_focus=true
This is with the status check that i put in
https://github.com/luigi311/Testing-Github-Actions/runs/465599447?check_suite_focus=true

As you can see on the first one it does not create the artifact or upload it because failure() is never true. On the second link it will create the artifact and upload it.

.ci/ci.sh
luigi311: I have personally never used this method and this was mostly just a copy over from travis. After this gets merged in I am not apposed to using flags.

.github/workflows/ci.yml
ePirat: Uh ok so apparently GitHubs caching is not really great for that usecase… Seems fine then.

.github/workflows/ci.yml
ePirat: Wow how confusing, they really should update the documentation to clarify that. Fine then.

.github/workflows/ci.yml
luigi311: What do you mean? It does the restore even if the hash does not exist. So you are still benefiting from ccache.

.github/workflows/ci.yml
luigi311: Yeah its terrible. There are a few people complaining about how they are handling continue-on-error and they should instead handle it like travis handles allow failures.

luigi311:
Looks like ref: 'master' is not going anywhere. Throwing it back in will fix the issue. It also looks like it was only an issue on the PR check. If you check the log in my branch it did not have the same behavior. Here is the log of that run from my fork and not the PR 
https://github.com/luigi311/SVT-AV1/runs/465374762?check_suite_focus=true
it does not do a merge and the log message shows the actual commit that is in the actual master

hassount:
Hi guys,

is this PR good to go ?

luigi311:
I resolved the conflicts. Once the testing is complete this is ready to be merged.

PR #995, opened 2020-01-30
FFmpeg plugin: Update readme to use clone depth 1

EwoutH:
Reduces required storage and reduces download times.

CI already uses clone depth 1.


hassount:
@kylophone any issues ?

PR #997, opened 2020-01-31
ALT REF Filtering Improvement

anaghdin:
**Description:**
ALT REF filtering improvement:
- Enable TF for layer 1 in 1 pass encoding (already active in 2 pass) 
- Adjust the filter strength

**Authors**
@anaghdin

**Type of change**
Enhancement

**Tests and performance**
Expected Average PSNR-SSIM BD-rate gain in the range of -0.5% (~-0.2% for TF in layer 1 and -0.3% for strength update) for 1 pass encoding over the fast objective list, using 5 QP values: {20, 32, 43, 55 and 63}.
Speed impact in context of M0 ~1%

### DIFF
--- a/Source/Lib/Encoder/Codec/EbTemporalFiltering.c
+++ b/Source/Lib/Encoder/Codec/EbTemporalFiltering.c
@@ -1954,9 +1954,13 @@ static void adjust_filter_strength(PictureParentControlSet *picture_control_set_
     // unsuccessful and therefore keep the strength as it was set
     if (noise_level > 0) {
         int noiselevel_adj;
+#if ALTREF_IMPROVEMENT
+        if (noise_level < 1.2)
+#else
         if (noise_level < 0.75)
             noiselevel_adj = -2;
         else if (noise_level < 1.75)
+#endif
             noiselevel_adj = -1;
         else if (noise_level < 4.0)
             noiselevel_adj = 0;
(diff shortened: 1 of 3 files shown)


Source/Lib/Encoder/Codec/EbTemporalFiltering.c
```diff
@@ -1982,7 +1986,12 @@ static void adjust_filter_strength(PictureParentControlSet *picture_control_set_
         strength = adj_strength;
     else
         strength = 0;
-
+#if ALTREF_IMPROVEMENT
+    // Decrease the filter strength for low QPs
+    if (picture_control_set_ptr_central->scs_ptr->static_config.qp <= ALT_REF_QP_THRESH){
+        strength = strength - 1;
+    }
+#endif
```
mafonsoN: I think it's better to combine this within the adjustment of strength function. Also, we need to make sure this is not overfitted to the test sequences we are currently using.

anaghdin:
> How much gain/speed impact did each of these changes bring: the adjustment of strength and the addition of tf for level 1 pictures?

The strength adjustment brings around 0.3% and tf for layer 1 0.2%

Source/Lib/Encoder/Codec/EbTemporalFiltering.c
anaghdin: It is now part of the same function (adjust_filter_strength()). I agree, we tried different thresholds, and found that in high bitrates, strong strength hurts the bdRate. Regarding the test sequences, we are expanding the test set  to include different types of content.

PR #1002, opened 2020-02-03
Enhanced RDOQ

### LINKED ISSUE #1001: quantize_fp is used for only INTRA blocks and Fast_RDOQ is not supported
Use quantize_fp for both INTRA and INTER blocks and add Fast_RDOQ support.

PhoenixWorthVCD:
## Description

quantize_fp is now added for INTER blocks @ RDOQ (previously added for INTRA blocks only). quantize_b is still called when RDOQ is not used (quantize_b and quantize_fp are both libaom kernels).
Also added the support for Fast_RDOQ (libaom technique). The idea consists of reducing the number of non-zero quantized coefficients before getting to the main/complex RDOQ stage (it performs an early check whether to zero out or not each of the non-zero quantized coefficients, and early exit if all are zeroed-out).

closes #1001 

## Type of Change

Enhancement

## Author

@hguermaz 
@PhoenixWorthVCD 

## Performance

Testing was performed on the objective-1-fast dataset (8 bit) in context of CORE.  A BD-rate gain of -0.27% was found for M0, with no speed deviation.

For 10 bit testing, fourteen 720p true 10bit clips were used.  In context of CORE, a BD-rate gain of -0.25% was found for M0, with no speed deviation.

### DIFF
--- a/Source/Lib/Encoder/Codec/EbFullLoop.c
+++ b/Source/Lib/Encoder/Codec/EbFullLoop.c
@@ -1414,6 +1414,11 @@ static const int plane_rd_mult[REF_TYPES][PLANE_TYPES] = {
 };
 
 #if FASTER_RDOQ
+/*
+ * Reduce the number of non-zero quantized coefficients before getting to the main/complex RDOQ stage
+ * (it performs an early check of whether to zero out each of the non-zero quantized coefficients,
+ * and updates the quantized coeffs if it is determined it can be zeroed out).
+ */
 static INLINE void update_coeff_eob_fast(uint16_t *eob, int shift, const int16_t *dequant_ptr,
                                          const int16_t *scan, const TranLow *coeff_ptr,
                                          TranLow *qcoeff_ptr, TranLow *dqcoeff_ptr) {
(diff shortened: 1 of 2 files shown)


Source/Lib/Encoder/Codec/EbFullLoop.c
```diff
@@ -1413,6 +1413,31 @@ static const int plane_rd_mult[REF_TYPES][PLANE_TYPES] = {
     {16, 10},
 };
 
+#if FASTER_RDOQ
+static INLINE void update_coeff_eob_fast(uint16_t *eob, int shift, const int16_t *dequant_ptr,
```
hassount: Can we please add some comments to this function

hassount:
@PhoenixWorthVCD, can you please rebase and fix the conflicts

Source/Lib/Encoder/Codec/EbFullLoop.c
PhoenixWorthVCD: Comments were added to the code.

PhoenixWorthVCD:
> The description of the changes should be reflected in the code, i.e. add comments in the code.

Comments have been added.  @fouedb please check if any more should be added.

PR #963, opened 2020-01-17
WIP Super-resolution tool - phase 1

mafonsoN:
## Description

First PR for the implementation of the frame super-resolution tool in svt-av1. Super-resolution is implemented for intra-only frames. Should be working for all enc-modes and also 10-bit inputs.

New command line arguments added: 
`-superres-mode` (0: None, 1: Fixed denominator, 2: Random denominator)
`-superres-denom` (denominator, d, used to determine encoding width = (8 * width + d/2 ) / d
`-superres-kf-denom` (denominator for key frames)

Example command line for super-res with x2 downsampling (d=16)
`./SvtAv1EncApp -i input.y4m -b out_bitstream.ivf -bit-depth 8 -enc-mode 6 -stat-report 1 -stat-file stat.txt -superres-mode 1 -superres-kf-denom 16`

Limitations (to be resolved before merging):

- [ ] Small rec/dec mismatch for some SBs
- [ ] Downsampled width needs to be multiple of 8 or the stream is corrupted
- [ ] No unit tests yet

For the next phase:

- Support for inter frame scaling
- 2-pass support
- More superres-modes

## Performance

No expected bitstream impact when super-res is disabled (default).

Performance impact when enabled TBD.

PR #1003, opened 2020-02-03
Resolve library errors for binary identical test in Github Actions

ultrawide:
There was an issue reported by @mafonsoN and @luigi311 where the the libraries were not being correctly linked when running the branch and master executables in the binary identical test.

See:
- https://github.com/OpenVisualCloud/SVT-AV1/pull/963/checks?check_run_id=413615745
- https://github.com/OpenVisualCloud/SVT-AV1/runs/414096956

I reordered the steps, changed the build from dynamically linked to statically linked, and changed where executables are called for clarity. 

Additionally, if the binary identical test fails on mismatch, we upload both files as artifacts so that they can be inspected for debugging.

Problem discussed in #873

### DIFF
--- a/.github/workflows/svt-av1.yml
+++ b/.github/workflows/svt-av1.yml
@@ -43,17 +43,17 @@ jobs:
         fetch-depth: 1
     - name: Compile SVT-AV1
       run: |
-        ./Build/linux/build.sh release
+        ./Build/linux/build.sh static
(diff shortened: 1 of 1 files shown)


EwoutH:
LGTM, nice work!

kylophone:
lgtm

PR #1005, opened 2020-02-04
Default -scm to 0

hassount:
## Description:

The screen content detector is having a lot of misdetections e.g. niklas clip. Also the detection takes place at the picture decision process and some features would need to be turned on / off (such as MRP, the max block size) at init time and switching them would cause major stability issues if done at run-time
This PR sets the detection to be off by default until further work is done there to solve all of these problems.

## Impact:

Screen content clips would have to be encoded using -scm 1

### DIFF
--- a/Config/Sample.cfg
+++ b/Config/Sample.cfg
@@ -96,7 +96,7 @@ NumberHmeSearchRegionInWidth    : 2                         # Set hierarchical m
 NumberHmeSearchRegionInHeight   : 2                         # Set Hme search region in height
 HmeLevel0TotalSearchAreaWidth   : 64                        # Set hierarchical motion estimation level0 total search area in Width
 HmeLevel0TotalSearchAreaHeight  : 25                        # Set Hme level0 total search area in height
-ScreenContentMode               : 2                         # Set screen content detection level
+ScreenContentMode               : 0                         # Set screen content detection level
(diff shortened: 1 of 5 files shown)


hassount:
@mafonsoN 
this is now adopting -scm 0 as default, so we can change the command lines for AWCY and VideoTracker to -scm 2 and that would cover it. Does that work ? if so, can we merge this ?

mafonsoN:
> @mafonsoN
> this is now adopting -scm 0 as default, so we can change the command lines for AWCY and VideoTracker to -scm 2 and that would cover it. Does that work ? if so, can we merge this ?

Did we make check if this can be changed for specific sequences in AWCY?

mafonsoN:
Or launching two separate runs would work for AWCY as well.

hassount:
@kylophone 
is there a default AWCY command line that we can change to include -scm 2?

hassount:
> > @mafonsoN
> > this is now adopting -scm 0 as default, so we can change the command lines for AWCY and VideoTracker to -scm 2 and that would cover it. Does that work ? if so, can we merge this ?
> 
> @hassount Did we make check if this can be changed for specific sequences in AWCY?

this PR keeps the option of the detection, we just need to add -scm 2 to enable it

PR #1011, opened 2020-02-05
nxm_sad_kernel_sub_sampled_helper_avx2 modification to call avx2 code for block width: 128, 48, 24

tszumski:
## Description:
nxm_sad_kernel_sub_sampled_helper_avx2 modification to call avx2 code for block width: 128, 48, 24

## Performance Impact:
(WxH) different block sizes
128x128 45.43x
128x64 81.64x
24x32 43.57x
24x24 42.09x
24x16 38.75x
48x48 53.03x
48x32 51.83x
48x16 42.86x

### DIFF
--- a/Source/Lib/Encoder/ASM_AVX2/EbComputeSAD_AVX2.h
+++ b/Source/Lib/Encoder/ASM_AVX2/EbComputeSAD_AVX2.h
@@ -73,6 +73,14 @@ uint32_t eb_compute64x_m_sad_avx2_intrin(
     uint32_t       height, // input parameter, block height (M)
     uint32_t       width); // input parameter, block width (N)
 
+uint32_t eb_compute128x_m_sad_avx2_intrin(
+    const uint8_t *src, // input parameter, source samples Ptr
+    uint32_t       src_stride, // input parameter, source stride
+    const uint8_t *ref, // input parameter, reference samples Ptr
+    uint32_t       ref_stride, // input parameter, reference stride
+    uint32_t       height, // input parameter, block height (M)
+    uint32_t       width); // input parameter, block width (N)
+
 void sad_loop_kernel_avx2_hme_l0_intrin(
     uint8_t * src, // input parameter, source samples Ptr
     uint32_t  src_stride, // input parameter, source stride
--- a/Source/Lib/Encoder/ASM_AVX2/EbComputeSAD_Intrinsic_AVX2.c
+++ b/Source/Lib/Encoder/ASM_AVX2/EbComputeSAD_Intrinsic_AVX2.c
@@ -3989,6 +3989,18 @@ uint32_t eb_compute64x_m_sad_avx2_intrin(
     return compute64x_m_sad_avx2(src, src_stride, ref, ref_stride, height);
 }
 
+uint32_t eb_compute128x_m_sad_avx2_intrin(
+    const uint8_t *src, // input parameter, source samples Ptr
+    uint32_t       src_stride, // input parameter, source stride
+    const uint8_t *ref, // input parameter, reference samples Ptr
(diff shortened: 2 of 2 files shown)


tszumski:
I can't reproduce "svt-av1 / binary-identical-test" mismatch on my local platform
The same test on travis "Binary Identical?" passed

hassount:
@tszumski 
Can you please update the performance section with the expected speedups as compared to the non-optimized version ?

tszumski:
@hassount 
I modified first comment with performance impact

mafonsoN:
> I can't reproduce "svt-av1 / binary-identical-test" mismatch on my local platform
> The same test on travis "Binary Identical?" passed

Travis seems to be passing. @hassount do you think this patch needs further investigation?

ttrigui:
> > I can't reproduce "svt-av1 / binary-identical-test" mismatch on my local platform
> > The same test on travis "Binary Identical?" passed
> 
> Travis seems to be passing. @hassount do you think this patch needs further investigation?

I tested couple of clips with different enc-modes and I got identical bitstreams. Can this be a problem with the test?

hassount:
> > > I can't reproduce "svt-av1 / binary-identical-test" mismatch on my local platform
> > > The same test on travis "Binary Identical?" passed
> > 
> > 
> > Travis seems to be passing. @hassount do you think this patch needs further investigation?
> 
> I tested couple of clips with different enc-modes and I got identical bitstreams. Can this be a problem with the test?

I believe the github actions test is broken, travis seems to have passed. Though what seems to be missing here is the unit test for this optimization unless we have it covered somehow, can we confirm ?

mafonsoN:
> > > I can't reproduce "svt-av1 / binary-identical-test" mismatch on my local platform
> > > The same test on travis "Binary Identical?" passed
> > 
> > 
> > Travis seems to be passing. @hassount do you think this patch needs further investigation?
> 
> I tested couple of clips with different enc-modes and I got identical bitstreams. Can this be a problem with the test?

It might be, we need to find that out. Also, we need to be to reproduce it.

mafonsoN:
@ttrigui  The tests seem to be running in ubuntu-bionic, compiled using gcc-9, with instructions up to avx2. Do we have such a system we can test on?

ttrigui:
> > > > I can't reproduce "svt-av1 / binary-identical-test" mismatch on my local platform
> > > > The same test on travis "Binary Identical?" passed
> > > 
> > > 
> > > Travis seems to be passing. @hassount do you think this patch needs further investigation?
> > 
> > 
> > I tested couple of clips with different enc-modes and I got identical bitstreams. Can this be a problem with the test?
> 
> I believe the github actions test is broken, travis seems to have passed. Though what seems to be missing here is the unit test for this optimization unless we have it covered somehow, can we confirm ?

Yes the unitTest exists. It is covered by nxm_sad_kernel_sub_sampled_helper_avx2.

ttrigui:
> @ttrigui The tests seem to be running in ubuntu-bionic, compiled using gcc-9, with instructions up to avx2. Do we have such a system we can test on?

Not sure @hassount do we have such environment? the AWS I'm using is ubuntu-bionic, gcc version 7.4.0 and supports up to AVX512

ttrigui:
> > @ttrigui The tests seem to be running in ubuntu-bionic, compiled using gcc-9, with instructions up to avx2. Do we have such a system we can test on?
> 
> Not sure @hassount do we have such environment? the AWS I'm using is ubuntu-bionic, gcc version 7.4.0 and supports up to AVX512

I installed gcc 9 and I run the test and I was not able to reproduce the mismatch. I also run the reference twice and the modified twice and no r2r.

mafonsoN:
@ttrigui This patch should be fine then.

Issue #994, opened 2020-01-29
Cleanup to do list


hassount:
As a follow up to #668 

**P0**
- [x] cleanup MbModeInfo 
- [ ] unify and concentrate mi geometry (xd->mi_row, xd->mi_col...) into one location
- [ ] use one unified structure between candidate_ptr and blk_ptr that has mode_info refactor
- [ ] cleanup sb_size 
- [ ] av1_inter_prediction (27 parameters passed 👎 )
- [ ] Remove useless NSQ signal(s): nsq_level, is_nsq_table_used. 
- [ ] Clean-up pic_obmc_mode derivation.
- [ ] Remove useless MR_MODE checks.
- [x] Link the 3 GM signals into 1: gm_level, compute_global_motion, global_mv_injection.
- [x] unify trellis rdoq
- [x] enc-dec-common file separation (#181, #702, #888, #917, #955, #1024, #1024, #1045, #1049)

**P1**
- [ ] collapse spare code of global injection (code under if(context_ptr->global_mv_injection) check).
- [ ] add the ability to enter actual count per stage instead of deriving them using bypass stage. 
- [ ] implement commandline switch `-sb-size` for setting superblock size and cleanup/unify the code relating to this (we currently have `sb_sz`, `sb_size`, and `super_block_size` presumably doing similar things)
- [ ] warped_motion_prediction, too many parameters
- [ ] picture decision kernel refactor (#870)
- [ ] replace all neigh arrays with mi map
- [ ] do_mc and do_recon 
- [ ] add / unify signal for is_nsq_table_used no enc-mode check 
- [ ] remove TX_SEARCH_INTER_DEPTH support (inter_depth_tx_search will be removed)
- [ ] eb_av1_fwd_txfm2d_4x16   /eb_av1_fwd_txfm2d_4x8/ eb_av1_fwd_txfm2d_8x4/ eb_av1_fwd_txfm2d_16x4/   declared twice : to remove second declaration and   eb_av1_fwd_txfm2d_4x16_sse4_1
- [ ] to add eb_ prefix for all functions
- [ ] remove enc_mode check at pred_me
- [ ] fix the new line "// Picture" comment 
- [ ] unify reference pointer generation
- [ ] move signal_derivation_mode_decision_config_kernel_oq to the Picture decision kernel
- [ ] clean this:
``` 
[...]

PR #1015, opened 2020-02-06
Improve Warp motion

anaghdin:
**Description:**
Improve warp motion by enabling warp motion for all references in list 0 and list 1

**Authors**
@anaghdin

**Type of change**
Enhancement

**Tests and performance**
Expected Average PSNR-SSIM BD-rate gain in the range of -0.2% for 1 pass encoding over the fast objective list, using 5 QP values: {20, 32, 43, 55 and 63}.
Speed impact in context of M0 ~1%

### DIFF
--- a/Source/Lib/Encoder/Codec/EbInterPrediction.c
+++ b/Source/Lib/Encoder/Codec/EbInterPrediction.c
@@ -6155,8 +6155,13 @@ static void chroma_plane_warped_motion_prediction_sub8x8(
     InterpFilterParams filter_params_x, filter_params_y;
 
     MV mv_l0;
+#if WARP_IMPROVEMENT
+    mv_l0.col = mv_unit->mv[mv_unit->pred_direction].x;
+    mv_l0.row = mv_unit->mv[mv_unit->pred_direction].y;
+#else
     mv_l0.col = mv_unit->mv[REF_LIST_0].x;
     mv_l0.row = mv_unit->mv[REF_LIST_0].y;
+#endif
 
     MV mv_q4 = clamp_mv_to_umv_border_sb(
         blk_ptr->av1xd, &mv_l0, blk_geom->bwidth_uv, blk_geom->bheight_uv, 1, 1);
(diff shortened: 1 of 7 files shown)


Source/Lib/Encoder/Codec/EbInterPrediction.c
```diff
@@ -6569,6 +6653,25 @@ EbErrorType warped_motion_prediction(PictureControlSet *picture_control_set_ptr,
                                                          mv_unit);
 
             // Cr
+#if WARP_IMPROVEMENT
+            if (mv_unit->pred_direction == UNI_PRED_LIST_0 || mv_unit->pred_direction == BI_PRED) {
+                src_ptr_l0 = ref_pic_list0->buffer_cr + (is16bit ? 2 : 1)
+                    * ((ref_pic_list0->origin_x + ((pu_origin_x >> 3) << 3)) / 2
```
mafonsoN: The operation of getting the pointer start is done so many times. Maybe this can be cleaned up to a single function.

Source/Lib/Encoder/Codec/EbModeDecision.c
```diff
@@ -2538,7 +2538,370 @@ void inject_new_nearest_new_comb_candidates(const SequenceControlSet *  scs_ptr,
...
+            uint8_t to_inject_ref_type = svt_get_ref_frame_type(REF_LIST_0, list0_ref_index);
+            uint8_t skip_cand = check_ref_beackout(
+                context_ptr,
+                to_inject_ref_type,
+                context_ptr->blk_geom->shape);
+
+            if (!skip_cand) {
+                for (int i = 0; i < 9; i++) {
+
+                    candidateArray[can_idx].type = INTER_MODE;
```
mafonsoN: There is a lot of repeated code here, for L0 and L1. Can this be done in a loop? Also a lot of indexes are hardcoded.

Source/Lib/Encoder/Codec/EbInterPrediction.c
hassount: added to #994

PR #1027, opened 2020-02-07
Fix issues in manual prediction structure scheme

elintencent:
This PR fixes issues in manual prediction structure scheme:

1. Move updating dependent lists for I slice from I slice to the start of the next minigop, because DPB management in manual prediction structure scheme needs processing frames in encoding order in a minigop, and I slice is always the first frame in encoding order but the last frame in display order in a minigop.
  
2. Fix merge issue for level5, YUV decoded from the streams made by the same manual and fixed prediction structures respectively can be bit identical.

### DIFF
--- a/Config/PredStruct_level5.cfg
+++ b/Config/PredStruct_level5.cfg
@@ -35,4 +35,4 @@ PredStructEntry :  27            25           3              1             1    
 PredStructEntry :  28            30           5              1             1             1                  -1
 PredStructEntry :  29            29           4              1             1             2                  -2
 PredStructEntry :  30            31           5              1             1             1                  -1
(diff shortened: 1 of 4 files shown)


elintencent:
Would someone please tell me why CI failed? I can build the branch succefully on GCC 7.

elintencent:
Auto check failure:
1. svt-av1 / binary-identical-test:
I can obtain the identical streams with the same command line and y4m in the auto check script. 
2. Travis CI - Pull Request:
I can't access the build log link, so don't konw what problem it ran into.

hassount:
@anaghdin is this good to go ?

mafonsoN:
Re-running Travis job for gcc-9

PR #1032, opened 2020-02-07
Fix unnecessary more memory allocation

tianjunwork:
Signed-off-by: Jun Tian <jun.tian@intel.com>

### DIFF
--- a/Source/Lib/Encoder/Codec/EbEncDecProcess.c
+++ b/Source/Lib/Encoder/Codec/EbEncDecProcess.c
@@ -48,7 +48,7 @@ static void enc_dec_context_dctor(EbPtr p) {
     EB_DELETE(obj->inverse_quant_buffer);
     EB_DELETE(obj->input_sample16bit_buffer);
     if (obj->is_md_rate_estimation_ptr_owner) EB_FREE(obj->md_rate_estimation_ptr);
-    EB_FREE_ARRAY(obj->transform_inner_array_ptr);
+    EB_FREE(obj->transform_inner_array_ptr);
     EB_FREE_ARRAY(obj);
 }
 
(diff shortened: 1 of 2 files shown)


tianjunwork:
Hi @xuguangxin, could you review this fix? Thank you.

Source/Lib/Encoder/Codec/EbEncDecProcess.c
```diff
@@ -85,8 +85,7 @@ EbErrorType enc_dec_context_ctor(EbThreadContext *  thread_context_ptr,
         enc_handle_ptr->picture_demux_results_resource_ptr, demux_index);
 
     // Trasform Scratch Memory
-    EB_MALLOC_ARRAY(context_ptr->transform_inner_array_ptr,
-                    3152); //refer to EbInvTransform_SSE2.as. case 32x32
+    EB_MALLOC(context_ptr->transform_inner_array_ptr, 3152); //refer to EbInvTransform_SSE2.as. case 32x32
```
xuguangxin: 3152 / (32x32) = 3.078. So where is the magic number came from?
Could we add some comments or define a descriptive macro for this magic number?
thanks

tianjunwork: Not sure about the magic number.
There are two `//refer to EbInvTransform_SSE2.as. case 32x32`. Both of them should use EB_MALLOC. The other is correct.

xuguangxin: I guess this line copied from [SVTHEVC](https://github.com/OpenVisualCloud/SVT-HEVC/blob/master/Source/Lib/Codec/EbModeDecisionProcess.c#L38). The max transform size of HEVC is 32x32. 32 * 32 * sizeof(uint16) = 2048. so 3120 bytes is enough.

> HEVC uses integer DCT and DST transforms with varied block sizes between 4x4 and 32x32.

But for AV1 the max transform size is 64 x 64, 64 * 64 * sizeof(uint16_t) = 8192. it's large than 3152 bytes. 

> To incorporate AV's extended coding block partitions, we support square, 2:1/1:2, and 4:1/1:4 transform sizes from 4×4 to 64×64.

Could you find some who know the magic number? else it may have potential problem here.

thanks

xuguangxin: @tianjunwork any feedback on this?
thanks

anaghdin: Could you check if the transform_inner_array_ptr is really used. It seems it has been passed to many functions, but voided inside.

tianjunwork: Added a Data Breakpoint with address of transform_inner_array_ptr. The address is not being written to.

tianjunwork: Could you double confirm from code logic? If yes, I will close this PR.
Maybe someone can remove the allocation all together.

xuguangxin: yes, it's never used.  I sent a patch to remove them.  Please help review it.
https://github.com/OpenVisualCloud/SVT-AV1/pull/1037

hassount:
#1037  merged

PR #1033, opened 2020-02-08
Change the QP assignment

anaghdin:
**Description:**
Change the QP assignment for CQP
Support QP assignment for different prediction structures

**Authors**
@anaghdin

**Type of change**
Enhancement

**Tests and performance**
Average PSNR-SSIM BD-rate gain in the range of -0.8% for 1 pass encoding and -1% for two pass encoding over the fast objective list, using 5 QP values: {20, 32, 43, 55 and 63}
Speed impact in context of M0 ~0%

### DIFF
--- a/Source/Lib/Encoder/Codec/EbRateControlProcess.c
+++ b/Source/Lib/Encoder/Codec/EbRateControlProcess.c
@@ -4085,7 +4085,409 @@ enum {
             name = NULL;                                         \
         }                                                        \
     } while (0)
+#if QPS_CHANGE
+static int kf_low_motion_minq_cqp_8[QINDEX_RANGE] = {
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
+    0,    0,    0,    0,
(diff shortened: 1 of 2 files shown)


Source/Lib/Encoder/Codec/EbRateControlProcess.c
```diff
@@ -4085,7 +4085,409 @@ enum {
             name = NULL;                                         \
         }                                                        \
     } while (0)
-
+#if QPS_CHANGE
+static int kf_low_motion_minq_cqp_8[QINDEX_RANGE] = {
```
mafonsoN: Where are these used? Could not find it.

anaghdin: use in the following function:
static int get_kf_active_quality_cqp(const RATE_CONTROL *const rc, int q,
    AomBitDepth bit_depth) {
    int *kf_low_motion_minq_cqp;
    int *kf_high_motion_minq_cqp;
    **ASSIGN_MINQ_TABLE(bit_depth, kf_low_motion_minq_cqp);**
    **ASSIGN_MINQ_TABLE(bit_depth, kf_high_motion_minq_cqp);**

mafonsoN: Thanks, missed that.

anaghdin:
> Do you have performance results for other datasets?

Yes, we evaluated over our new test set (65 clips) and got -0.76% for 1 pass coding

PR #1045, opened 2020-02-14
Enc common separation 7

andrey-norkin:
Separation of the encoder library into encoder and common. Removal of dependency on EntropyCoding in the decoder


hassount:
tested bitstream exactness for -ec-mode 0,4 and 8

PR #1051, opened 2020-02-17
16bit pipeline for filter

mtangtencent:
This PR introduces the changes for 16bit pipeline, which controls whether to use 16bit pipeline for filter (dlf, cdef, rest) according to a new cfg parameter `-16bit-pipeline`.

Currently, recon is copied to recon16 before filter, while recon16 is copied back to recon after filter. The code for copy can be removed once the 16bit pipeline is ready the entire encoding process.

### DIFF
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -152,6 +152,13 @@ typedef struct EbSvtAv1EncConfiguration {
      *
      * Default is 8. */
     uint32_t encoder_bit_depth;
+    /* Specifies whether to use 16bit pipeline.
+     *
+     * 0: 8 bit pipeline.
+     * 1: 16 bit pipeline.
+     * Now 16bit pipeline is only enabled in filter
+     * Default is 0. */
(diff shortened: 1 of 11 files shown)


Source/API/EbSvtAv1Enc.h
```diff
@@ -152,6 +152,13 @@ typedef struct EbSvtAv1EncConfiguration {
      *
      * Default is 8. */
     uint32_t encoder_bit_depth;
+    /* Specifies whether to use 16bit pipeline.
+     *
+     * 0 = 8 bit pipeline.
```
fouedb: Change to: 
     * 0:  8 bit pipeline.
     * 1: 16 bit pipeline.

hassount:
@mtangtencent 
Could you please address @fouedb comment so that we can merge this PR?

Source/API/EbSvtAv1Enc.h
mtangtencent: Thx @fouedb. addressed. @hassount

Issue #687, opened 2019-10-09
Expose features/tools at the encoder configuration


joelsole:
Encoder features/tools to be controlled via configuration flags: enable turning on/off AV1 coding tools for finer control of complexity/efficiency tradeoffs and for easier testing & development.

List of tools to include:

- [X] RDOQ `-rdoq`
- [x] allow_intrabc
- [X] allow_warped_motion `-local-warp`
- [x] atb_mode `-atb`
- [ ] cdf_mode `-cdf`
- [X] combine_class12 `-class-12`
- [X] edge_based_skip_angle_intra `-intra-edge-skp`
- [X] enable_interintra_compound `-interintra-comp`
- [X] enable_restoration `-restoration-filtering`
- [x] enable_cdef `-cdef`
- [X] fractional_search64x64 `-frac-search-64`
- [X] frame_end_cdf_update_mode `-framend-cdf-upd-mode`
- [X] global-motion `-global-motion`
- [X] is_nsq_table_used `-nsq-table-use`
- [X] mfmv_enabled `-mfmv`
- [X] new_nearest_near_comb_injection `-new-nrst-near-comb`
- [X] nx4_4xn_parent_mv_injection `-nx4-4xn-mv-inject`
- [X] over_boundary_block_mode `-over-bndry-blk`
- [ ] perform_quantize_fp `-quantize-fp`
- [X] prune_ref_frame_for_rec_partitions `-prune-ref-rec-part`
- [X] prune_unipred_at_me `-prune-unipred-me`
- [X] redundant_blk `-redundant-blk`
- [X] spatial_sse_full_loop `-spatial-sse-fl`
- [X] trellis_quant_coeff_optimization `-trellis`
- [x] update_cdf `-updt-cdf`
- [X] use_subpel_flag `-subpel`
- [ ] adp_level
- [X] bipred3x3_injection `-bipred-3x3`
- [x] cdef_filter_mode
- [X] chroma_level `-chroma`
- [X] compound_mode (comp diff) `-compound`
- [ ] cu8x8_mode
- [ ] down_sampling_method_me_search
- [ ] fractional_search_method
- [ ] fractional_search_model
- [ ] full_loop_escape
- [ ] half_pel_mode
- [ ] HME tables
- [ ] hme_search_method
- [ ] ibc_mode
- [ ] interpolation_filter_search_blk_size
- [ ] interpolation_search_level
- [ ] intra_pred_mode
[...]

PR #1052, opened 2020-02-17
CDEF command line options

adaria:
| **Configuration file parameter** | **Command line** | **Range** | **Default** | **Description** |
| --- | --- | --- | --- | --- |
| **CDEFMode** | -cdef-mode | [0-5, -1 for default] | -1 | CDEF Mode, 0: OFF, 1-5: ON with 2,4,8,16,64 step refinement, -1: DEFAULT|

Issue #687


adaria:
Investigating binary-identical-test failure

Source/Lib/Encoder/Codec/EbResourceCoordinationProcess.c
```diff
@@ -186,6 +186,11 @@ EbErrorType signal_derivation_pre_analysis_oq(SequenceControlSet *     scs_ptr,
         scs_ptr->seq_header.enable_restoration =
             (uint8_t)scs_ptr->static_config.enable_restoration_filtering;
 
+    if (scs_ptr->static_config.enable_cdef == DEFAULT)
+        scs_ptr->seq_header.enable_cdef = 1;
+    else
+        scs_ptr->seq_header.enable_cdef = scs_ptr->static_config.enable_cdef;
+
     scs_ptr->cdf_mode = (pcs_ptr->enc_mode <= ENC_M6) ? 0 : 1;
```
adaria: The initialization added above was missing and causing the binary mismatch.
is `scs_ptr->cdf_mode` misnamed and should be `scs_ptr->cdef_mode` here?

Source/Lib/Encoder/Codec/EbPictureDecisionProcess.c
```diff
@@ -986,17 +986,21 @@ EbErrorType signal_derivation_multi_processes_oq(
     // 3                                            8 step refinement
     // 4                                            16 step refinement
     // 5                                            64 step refinement
+    // TODO adaria: do I need to decouple enable_cdef and allow_intrabc here? at least signal to the user that they're coupled
```
hassount: they need to be coupled because they can not work together within the same stream

hassount:
@adaria: can't we merge these together into 1 flag ? the reason why there are two of them is that one is set at the sequence header so that the encoder enables writing such in the bitstream then the decision is made per picture. so if the user would like it off, all we need to do is to shut the  sequence header flag off and so they could be all merged into 1 flag?

adaria:
> @adaria: can't we merge these together into 1 flag ? the reason why there are two of them is that one is set at the sequence header so that the encoder enables writing such in the bitstream then the decision is made per picture. so if the user would like it off, all we need to do is to shut the sequence header flag off and so they could be all merged into 1 flag?

Yes, makes sense. I'll merge them into one flag with 0 for OFF and 1-5 for mode.

adaria:
Will combine multiple on/off tools into a larger PR rather than adding them individually

Docs/Appendix-CDEF.md
```diff
@@ -584,14 +584,14 @@ The search `in cdef_seg_search` and in `finish_cdef_search` for the
 filter strength is performed by considering a sub-interval of the filter
 strength index \[0,63\]. The subinterval is given by
 \[`cdef_ref_frame_strength` – `gi_step`, `cdef_ref_frame_strength` +
-`gi_step`\], where `cdf_ref_frame_strength` is the filter strength for
```
hassount: cdf is different from cdef in this case

Issue #129, opened 2019-03-01
[ffmpeg_plugin] Invalid video timestamps when decoding via dav1d


jbeich:
66e8c9d8e578 snapshot, built by Clang 9.0. ffmpeg n4.2.1 / dav1d 0.5.1 / mpv v0.30.0.

```
$ curl -sO https://media.xiph.org/video/derf/y4m/elephants_dream_720p24.y4m.xz
$ xz -d elephants_dream_720p24.y4m.xz

$ ffmpeg -hide_banner -i elephants_dream_720p24.y4m -frames:v 120 -c:v libsvt_av1 -y ed.ivf
Input #0, yuv4mpegpipe, from 'elephants_dream_720p24.y4m':
  Duration: 00:10:53.79, start: 0.000000, bitrate: 265421 kb/s
    Stream #0:0: Video: rawvideo (I420 / 0x30323449), yuv420p(progressive), 1280x720, SAR 1:1 DAR 16:9, 24 fps, 24 tbr, 24 tbn, 24 tbc
Stream mapping:
  Stream #0:0 -> #0:0 (rawvideo (native) -> av1 (libsvt_av1))
Press [q] to stop, [?] for help
SVT [version]:  SVT-AV1 Encoder Lib v0.7.0
SVT [build]  :  GCC FreeBSD Clang 9.0.0 (tags/RELEASE_900/final 372316)  64 bit
LIB Build date: Nov 23 2019 00:28:17
-------------------------------------------
Number of logical cores available: 8
Number of PPCS 72
-------------------------------------------
SVT [config]: Main Profile      Tier (auto)     Level (auto)
SVT [config]: EncoderMode                                                       : 8
SVT [config]: EncoderBitDepth / EncoderColorFormat / CompressedTenBitFormat     : 8 / 1 / 0
SVT [config]: SourceWidth / SourceHeight                                        : 1280 / 720
SVT [config]: Fps_Numerator / Fps_Denominator / Gop Size / IntraRefreshType     : 24 / 1 / 32 / 1
SVT [config]: HierarchicalLevels / BaseLayerSwitchMode / PredStructure          : 4 / 0 / 2
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange                  : CQP / 50 / 33 / 0
-------------------------------------------
SVT Memory Usage:
    total allocated memory:       3.67 GB
        malloced memory:          2.64 GB
        callocated memory:        134.16 MB
[...]

Issue #33, opened 2019-02-08
One frame is written to each IVF packet instead of one temporal unit


tdaede:
Each IVF packet must contain one temporal unit. As a result, svt-av1 playback is currently jittery or broken when using altrefs (the default).

moisesmcardona:
This must be what I've been experimenting...

In this thread I have a sample using SVT-AV1 and the same video encoded with rav1e:
https://www.reddit.com/r/AV1/comments/ankpad/svtav1_encode_test_vs_rav1e_test/

kylophone:
@hassount, just want to make sure you saw this. Looks like the `ivf` muxer needs help.

j-mas:
Please solve this issue! The resulting ivf files are so jerky that they are not usable with the mpv or VLC player, and therefore not usable for me. Many Thanks!

hassount:
@j-mas 
ok, we will work on this with a high priority.

j-mas:
Thank you!

jsunintel:
#33 is closed. Anymore works needed from me?

gnafuthegreat:
To clarify, #42 is closed.  This issue (#33) is not yet resolved.  I'll chime in as another person touched by this issue, and I'd also like to say I've seen some confusion on a few different forums with users having issues either with stuttering playback or issues muxing into other containers because of the malformed IVF container.

@hassount Thanks for working on this.

EwoutH:
@hassount @jsunintel Any progress update?

hassount:
Working on it, found the problem and working on a fix that works with the API design.

hassount:
#107

j-mas:
For me it now works! No longer jerky!
Thank you!

j-mas:
With the newest build in AppVeyor it is jerky and chubby again! The Problem is not solved!

hassount:
@j-mas you mean from the master ? #107 is not merged yet. Please let me know if you're seeing it from the PR build

j-mas:
@hassount yes I mean from the master.

tnti:
@hassount 
 Fast forward will cause problems with the search for keyframes. hassount
[mov,mp4,m4a,3gp,3g2,mj2 @ 0000021240219fc0] seek_frame_generic failed as this stream seems to contain no keyframes after the target timestamp, 1002 non keyframes found
[...]

PR #1053, opened 2020-02-18
packetization_kernel: pack entire temproal unit for a output buffer

xuguangxin:
Currently, we output one frame per get ouptut, and the "show existing frame" shares the same pts with the previous frame.
This will introduce jitter to "show existing frame" and the previous frame.
We should output a temporal unit a time.
A temporal unit includes one TD + zero or more undisplay frame + one display frame.
And we need to stamp a different pts for show existing frame since it is a displayable frame.

after apply this patch, ffmpeg will not report error for following test command:
ffmpeg -i 320x240.y4m -c:v libsvt_av1 -vframes 100 test.mkv  -y &&  ffmpeg -c:v libdav1d -i test.mkv -f null /dev/null

and ffmpeg will not insert td for every frame if we use following command:
ffmpeg -i 320x240.y4m -c:v libsvt_av1 -vframes 100 test.ivf  -y

fixes https://github.com/OpenVisualCloud/SVT-AV1/issues/129, fixes https://github.com/OpenVisualCloud/SVT-AV1/issues/33

### DIFF
--- a/Source/App/EncApp/EbAppProcessCmd.c
+++ b/Source/App/EncApp/EbAppProcessCmd.c
@@ -1026,20 +1026,6 @@ static void write_ivf_stream_header(EbConfig *config) {
 
     return;
 }
-static void update_prev_ivf_header(EbConfig *config) {
-    char header[4]; // only for the number of bytes
-    if (config && config->bitstream_file && config->byte_count_since_ivf != 0) {
-        fseeko(config->bitstream_file,
-               (-(int32_t)(config->byte_count_since_ivf + IVF_FRAME_HEADER_SIZE)),
-               SEEK_CUR);
-        mem_put_le32(&header[0], (int32_t)(config->byte_count_since_ivf));
-        fwrite(header, 1, 4, config->bitstream_file);
-        fseeko(config->bitstream_file,
-               (config->byte_count_since_ivf + IVF_FRAME_HEADER_SIZE - 4),
-               SEEK_CUR);
-        config->byte_count_since_ivf = 0;
-    }
-}
 
 static void write_ivf_frame_header(EbConfig *config, uint32_t byte_count) {
     char    header[IVF_FRAME_HEADER_SIZE];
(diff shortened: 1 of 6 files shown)


xuguangxin:
@Cidana-Developers please help check unit test part.
thanks

Cidana-Developers:
> @Cidana-Developers please help check unit test part.
> thanks

The changes are OK, thank you!

kylophone:
LGTM. Please squash before merging.

hassount:
@xuguangxin thank you very much for this PR as it addressed multiple issues. 
we ran several tests and have noticed that in regular encodes, there's no difference between the bitstreams with your changes and without. 
however, when using tiles, we notice an extra TD and Frame header that gets sent at the very end of the bitstream as compared to the master base of your branch using tiles too. 
the command line parameters is as follows:
-i in.yuv -w 176 -h 144 -b out.ivf -tile-rows 1 -tile-columns 1

Can you explain why we have the extra overhead ?

oddstone:
Hi @hassount ,
Could you share more details? Use your comandline, with or without the patch,  I can get bit matched result.
thanks

[bits.zip](https://github.com/OpenVisualCloud/SVT-AV1/files/4243023/bits.zip)

hassount:
> Hi @hassount ,
> Could you share more details? Use your comandline, with or without the patch, I can get bit matched result.
> thanks
> 
> [bits.zip](https://github.com/OpenVisualCloud/SVT-AV1/files/4243023/bits.zip)

It seems like one of the bitstreams I got was missing 6 bytes at the end of the stream. I was not able to reproduce the mismatch. sorry about the false alarm.

xuguangxin:
It's ok。  :)

hassount:
@xuguangxin 
found a way to reproduce the issue that I was seeing. 

-i blue_sky_360p_60f.yuv -o rec.yuv -intra-period 200 -enc-mode 8 -n 25  -q 40  -lp 1 -lad 0 -b 1.ivf  -w 640 -h 360

the encoder will encode only 24 frames. @xuguangxin could you please take a look at this, seems like an overlay header is missing

xuguangxin:
sure I will check this.
thanks for the comandline

xuguangxin:
@hassount 
It will fix by https://github.com/OpenVisualCloud/SVT-AV1/pull/1092.

PR #1059, opened 2020-02-20
Fix compilation error for old gcc

mtangtencent:
This PR fixes the compilation issue when using old version of gcc, e.g. 4.8.5.
Unnecessary intrinsic data type usage in header files are removed.


hassount:
Tested this on a few clips and got the same bitstreams.

PR #1064, opened 2020-02-21
Travis CI: Fix dependencies not installing on macOS

EwoutH:
An error in Homebrew caused yasm to fail to install on macOS. This PR migrates that by updating Homebrew before install.

See https://github.com/Homebrew/homebrew-bundle/issues/646 and https://travis-ci.community/t/macos-build-fails-because-of-homebrew-bundle-unknown-command/7296/7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
       - yasm
       - ccache
       - python
+    update: true # Remove when Travis macOS images get updated
 notifications:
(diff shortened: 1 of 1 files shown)


EwoutH:
Note to the whole team: If you see CI failing for a reason you did not cause, please create an issue and tag me or some else familiar with the CI. This way we can fix it before accidentally more gets broken!

PR #1065, opened 2020-02-21
CI: Speed up AppVeyor by dropping VS debug job

EwoutH:
As of this moment, AppVeyor is now 40 jobs behind, with an average of 25 minutes per job that's over 16 hours. We should speed it up significantly, this PR starts that by dropping the Visual Studio debug job.

@1480c1 What do you think about further options to speedup AppVeyor?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,5 @@
 image: Visual Studio 2019
-configuration:
-  - Debug
-  - Release
+configuration: Release
 
 environment:
   APPVEYOR_SAVE_CACHE_ON_ERROR: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,11 +3,11 @@ configuration: Release
 skip_branch_with_pr: true
 skip_commits:
   files:
-    - .github/*
-    - Docs/*
-    - .travis.yml
-    - .gitignore
-    - *.md
+    - '.github/*'
+    - 'Docs/*'
+    - '.travis.yml'
+    - '.gitignore'
+    - '*.md'
 environment:
   APPVEYOR_SAVE_CACHE_ON_ERROR: true
   MSYSTEM: MINGW64
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,13 @@
 image: Visual Studio 2019
 configuration: Release
 skip_branch_with_pr: true
-
+skip_commits:
+  files:
+    - .github/*
+    - Docs/*
+    - .travis.yml
+    - .gitignore
+    - *.md
 environment:
   APPVEYOR_SAVE_CACHE_ON_ERROR: true
   MSYSTEM: MINGW64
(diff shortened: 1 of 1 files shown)


luigi311:
We might run into similar issues with CI due to the e2e and unit tests taking up to 35 minutes to run.

appveyor.yml
```diff
@@ -1,7 +1,6 @@
 image: Visual Studio 2019
-configuration:
-  - Debug
-  - Release
+configuration: Release
+skip_branch_with_pr: true
```
luigi311:
```suggestion
skip_commits:
  files:
    - .github/*
    - Docs/*

```

luigi311:
We can also skip commits that only have changes to documentation or github actions.
https://www.appveyor.com/docs/how-to/filtering-commits/#commit-files-github-and-bitbucket-only

appveyor.yml
```diff
@@ -1,8 +1,13 @@
...
-
+configuration: Release
+skip_branch_with_pr: true
+skip_commits:
+  files:
+    - .github/*
+    - Docs/*
+    - .travis.yml
+    - .gitignore
+    - *.md
```
luigi311:
```suggestion
    - '*.md'
```

luigi311: Looks like you need to single quote any entries that start with the asterisk

EwoutH: Indeed

EwoutH:
@luigi311 The commits filtering is a good idea, thanks!

EwoutH:
@1480c1 Interesting option to explore. For now we have taken 3 measures that reduce the AppVeyor usage, shall we check if this helps for a week or so? If it's not enough we can move the MSVC build to Actions.

luigi311:
Looking through the appveyor job that is running right now it seems like almost all the time is taken up by the unix makefile release job. Going through the log of it it seems like we are backing up the .ccache files and restoring them but looks like it still takes 6 minutes to build svt-av1 where as the github actions only takes 20 seconds with ccache, are we certain that we are utilizing ccache in the appveyor build process? If it is not utilizing the ccache then ffmpeg is being built from scratch every single run and that would explain the slow build time on it.

EwoutH:
I'm not really familiar with the caching system or the MinGW64 job, so please look into it!

I would suggest to do that with a separate PR though and merge this ASAP.

luigi311:
That is fine with me

luigi311:
Looks like ffmpeg is atleast using ccache based on the configure command and what it is outputting. I see a restore cache section where it is downloading the ccache directory but i dont see anything referencing an upload ccache, @1480c1 do you know if that happens afterwards and is not in the log? I am not to familiar with appveyor.

kylophone:
lgtm

hassount:
@EwoutH @luigi311 good to go ?

luigi311:
@hassount Looks good to me

Issue #1047, opened 2020-02-14
Performance PRs to be ported to the master


hassount:
@hguermaz
- [x] Spatial Domain Distortion @ only the last stage (PD2/MD_STAGE_3) (#1057)
- [x] Remove layer check(s) @ PD1 settings derivation (#1057)
- [x] Tx Size Search Depth_2 INTRA/INTER 8BIT/10BIT (#1060)
- [x] TXS/RDOQ for block_copy (#1060)
- [x] Rate Estimation Bug Fixes (#1060)
- [x] Improve Nx4 & 4xN INTER candidates injection, and 1/8 Pel ME MVs refinement (#1063)
- [x] Signal(s) clean-up. (#1080)
- [x] IFS at last MDS. (#1113)
- [x] QPS (#1117 )
- [x] plane-wise Altref,, tf ME swicth , packing. (#1120 )

@ttrigui 
- [x] 16bit-pipeline (#1138)

@phoenixworthVCD
- [x] sq weight (#1006 )
- [x] mds 3 (#1009 )
- [x] rdoq opt / quant fp enhancement (#1002 )
- [x] tx size code refactoring ... (#1010 )

@okhlif
- [x] ME Search_Area = f(temporal_distance, frame_rate), HME for 360p.(#1067)
- [x] SB128x128 for 360p. (#1110 )
- [x] Lambda for 10BIT. (#1119)
- [x] NIC Levels (enhancements and refactor) (#1069)
- [x] ME-based ref frames pruning. (#1070 )
- [x] HP/QP Switch. (#1070 )

@nadermahdi
- [x] Optimized Chroma Search. (#1071)
- [x] Skip Depth @ PD2. (#1111)
- [x] NSQ Opt @ Full & Sub Pel. (#1112)

@chkngit 
- [x] ME-based ref frames pruning. (#1040)
- [x] compound similar candidates (#1058)
- [x] intra similar candidates (#1058)
- [x] nsq weight (#1121)

@lijing0010 
- [x] 31B MRP (#1072)

@anaghdin 
- [x] alt-ref enhancement (#997 )
- [x] warped motion enhancement (#1015)
- [x] md rate estimation (#979 )
- [x] qps for incomplete GOP (#1033)

@OmranAbdelkafi 
- [x] structures cleanup (commit c80b6f0 from the cs2 branch) (#1140 )
- [x] cli enhancements (commit 77a40c9 from the cs2 branch) (#1141 )

[...]

PR #1071, opened 2020-02-22
Chroma search optimization 2

NaderMahdi:
## Description
Perform the Chroma search algorithm just before the last md stage and only when at least one intra mode make it to this stage.
This is a replacement of PR#1039

## Related issue
#1047 

## Performance (from PR1039)
* Bdrate 0.02% Bdrate loss
* Speed 3.5% speed up
* Memory 1% increase

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -34,6 +34,13 @@
 extern "C" {
 #endif
 
+#define CHROMA_SEARCH_OPT        1 // Move chroma search to be done on the best intra candidate survived from MD stage 2
+#if CHROMA_SEARCH_OPT
+#define INFR_OPT                 1 // Lossless: Infrastructure work to allow the protability of the chroma search
+#define MOVE_OPT                 1 // Semi-lossless: Move the intra search to just before last md_stage
+#define COMP_OPT                 1 // Speed optimisation by restricting the search only when at least 1 intra candidate survives to the last md_stage
+#endif
+
 #define NICS_CLEANUP        1
 #define COMP_SIMILAR        1 //use previously coded similar blocks to prune compound modes
 #define INTRA_SIMILAR       1 //If previous similar block is intra, do not inject any inter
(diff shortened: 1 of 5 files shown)


Source/Lib/Common/Codec/EbDefinitions.h
```diff
@@ -34,6 +34,12 @@
 extern "C" {
 #endif
 
+#define CHROMA_SEARCH_OPT        1 // Move chroma search to be done on the best intra candidate survived from MD stage 2
```
fouedb: // Chroma search to be performed only for best intra candidate that survives from MD stage 2.

Source/Lib/Common/Codec/EbDefinitions.h
```diff
@@ -34,6 +34,12 @@
...
+#if CHROMA_SEARCH_OPT
+#define INFR_OPT                 1 // Lossless: Infrastructure work to allow the protability of the chroma search
+#define MOVE_OPT                 1 // Semi-lossless: Move the intra search to just before last md_stage
```
fouedb: Please clarify.

Source/Lib/Common/Codec/EbDefinitions.h
```diff
@@ -34,6 +34,12 @@
...
+#define COMP_OPT                 1 // Speed optimisation by restricting the search only when at least 1 intra candidate survives to the last md_stage
```
fouedb: Speed optimization  - Restrict the search only to the case when at least 1 intra candidate survives to the last md_stage. 

Clarify what "Restrict the search" means.

mafonsoN:
Can you put the performance impact of these changes here again?
Did you apply the clang format to these changes?

hassount:
> Did you apply the clang format to these changes?

The clang format seems to be creating some funky lines, there's someone working on fixing this, then we can apply it to all files again to fix everything.

Source/Lib/Common/Codec/EbDefinitions.h
NaderMahdi: These are Macros that will be removed eventually.

Source/Lib/Common/Codec/EbDefinitions.h
NaderMahdi: These are Macros that will be removed eventually.

Source/Lib/Common/Codec/EbDefinitions.h
NaderMahdi: These are Macros that will be removed eventually.

PR #1072, opened 2020-02-24
Enable MRP for 31B

lijing0010:
under MRP_31B_SUPPORT

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -63,6 +63,7 @@ extern "C" {
 #define NON_AVX512_SUPPORT
 #endif
 
+#define MRP_31B_SUPPORT 1
 #define TILES_PARALLEL 1
 
 #if TILES_PARALLEL
--- a/Source/Lib/Encoder/Codec/EbPictureDecisionProcess.c
+++ b/Source/Lib/Encoder/Codec/EbPictureDecisionProcess.c
@@ -2486,6 +2486,803 @@ static void  av1_generate_rps_info(
             //Layer1 toggle 3->4
             context_ptr->lay1_toggle = 1 - context_ptr->lay1_toggle;
         }
+#if MRP_31B_SUPPORT
+    } else if (pcs_ptr->hierarchical_levels == 5) {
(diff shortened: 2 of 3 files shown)


Source/Lib/Common/Codec/EbDefinitions.h
```diff
@@ -63,6 +63,7 @@ extern "C" {
 #define NON_AVX512_SUPPORT
 #endif
 
+#define MRP_31B_SUPPORT 1
```
bzhutencent: Can we add some comment to explain what it is for?

lijing0010: just as the macro name suggested, multi-refs support for 31B(hierarchical-level 5)
Previous only have 1 forward and 1 backward

bzhutencent: Thanks for the reply,  have a few more comments/questions:
1.  The macro name is clear to us, but might not be so straightforward to others, especially the people who are new to this.
2. Do we have the similar macro for other gop structures?
3. Looks this one is a super set of current implementation, can we just directly replace the current implementation with the new one? Generally too many #ifdef just impact the readability of the code.

lijing0010: All other minigop size(level 0,1,2,3,4) support multi-ref except level5, that's why I'm enabling it. When we enabled the multi-ref for other minigops, we also have the similar macros, and we removed them later when it's stable. 
I know it's boring and makes the code hard to read.
Given we'll integrate a lot of PRs from different persons, sometimes we may have bugs/crash and need to locate/resolve it quickly, that's why for each new feature we'll have it under a macro. And I think we'll remove those #ifdefs once proven stable. 
@hassount maybe you can comment more on this?

bzhutencent: The change looks fine with me if that is how it was done before.

hassount:
@bzhutencent is this good to go ?

OmranAbdelkafi:
We have ran some tests using "-hierarchical-levels 5" and we have noticed two things : 
- Significant increase in memory.
- Deadlock after encoding the first frame of any clip

PR #1076, opened 2020-02-24
fix memory leaks for tile encoder

xuguangxin:
Bin/Debug/SvtAv1EncApp -i in.yuv -w 176 -h 144 -b out.ivf -tile-rows 1 -tile-columns 1
you will get the leak print out


hassount:
For all reviewers, this is a simple fix.
Merging assuming no objections.

PR #853, opened 2019-12-03
EbPictureBufferDesc: fix crash for none 16 aligned 10bits input

xuguangxin:
example commandline

> SvtAv1EncApp -i 424x240.yuv -w 424  -h 240 -bit-depth 10 -fps 24.0  -n 2877 -rc 2 -tbr 200000  -irefresh-type 2 -enc-mode 5 -b test.ivf

you will get following crash stack.

> 
> [Switching to Thread 0x7ffdbe66b700 (LWP 170530)]
> 0x00007fffed3d0966 in dc_store_8xh (dst=0x7fff74424c68, stride=372, height=8, dc=0x7ffdbe6687f0) at /home/thomas/codec/SVT-AV1/Source/Lib/Common/ASM_SSE2/highbd_intrapred_sse2.c:410
> 410             _mm_store_si128((__m128i *)dst, dc_dup);
> (gdb) bt
> 0  0x00007fffed3d0966 in dc_store_8xh (dst=0x7fff74424c68, stride=372, height=8, dc=0x7ffdbe6687f0) at /home/thomas/codec/SVT-AV1/Source/Lib/Common/ASM_SSE2/highbd_intrapred_sse2.c:410
> 1  0x00007fffed3d0a63 in eb_aom_highbd_dc_left_predictor_8x8_sse2 (dst=0x7fff74424980, stride=372, above=0x7ffdbe668a10, left=0x7ffdbe668b50, bd=10) at /home/thomas/codec/SVT-AV1/Source/Lib/Common/ASM_SSE2/highbd_intrapred_sse2.c:479
> 2  0x00007fffed1d469a in build_intra_predictors_high (xd=0x7ffdbe668f88, topNeighArray=0x7ffdbe669ab2, leftNeighArray=0x7ffdbe6699a2, dst=0x7fff74424980, dst_stride=372, mode=DC_PRED, angle_delta=0, filter_intra_mode=FILTER_INTRA_MODES, tx_size=TX_8X8, disable_edge_filter=0, n_top_px=0, n_topright_px=0,
>     n_left_px=8, n_bottomleft_px=8, plane=1, bd=10) at /home/thomas/codec/SVT-AV1/Source/Lib/Common/Codec/EbIntraPrediction.c:3892
[...]

PR #1078, opened 2020-02-24
asm: change aligned read/store to unaligned read/store

xuguangxin:
this is a follow up for https://github.com/OpenVisualCloud/SVT-AV1/pull/853

following assembly code changed:
_mm_load_ps,
_mm_store_ps,
_mm256_load_si256,
_mm256_store_si256,
_mm256_load_ps,
_mm256_store_ps,

enc-mode 8 and 4 are tested, no obvious fps drop.

### DIFF
--- a/Source/Lib/Common/ASM_AVX2/EbHighbdIntraPrediction_AVX2.c
+++ b/Source/Lib/Common/ASM_AVX2/EbHighbdIntraPrediction_AVX2.c
@@ -1365,9 +1365,9 @@ static INLINE void load_right_weights_8(const uint16_t *const above, __m256i *co
     *r = _mm256_set1_epi16((uint16_t)above[7]);
 
     // 0 1 2 3  0 1 2 3
-    weights[0] = _mm256_load_si256((const __m256i *)(sm_weights_d_8 + 0x00));
+    weights[0] = _mm256_loadu_si256((const __m256i *)(sm_weights_d_8 + 0x00));
     // 4 5 6 7  4 5 6 7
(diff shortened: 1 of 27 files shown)


xuguangxin:
I have compare the bitstream before and after the patch.
enc8 can get md5 match. 
enc4 will not, because master have r2r issues.

MinhaoTang:
There is one usage of "_mm_store_si128" in ComputeSAD AVX512.
The others looks good to me.

mtangtencent:
There is one usage of "_mm_store_si128" in ComputeSAD AVX512.
The others looks good to me.

xuguangxin:
@mtangtencent , thanks for reminder. fixed.

we still have two aligned store and load at 
https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Source/Lib/Common/ASM_AVX512/synonyms_avx512.h#L32

But I don't have AVX512 machine to test it.
@hassount , @mtangtencent  do you sugget me change it too?
thanks

mtangtencent:
I think we should include the change of replacing avx512 aligned load/store.
In theory, it is always safe to replace load/store by loadu/storeu, but we still better test it.

xuguangxin:
the concern is the performance, I did not see any performance description at https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_load_si512&expand=3345
We need find a CLX to test it.

hassount:
> the concern is the performance, I did not see any performance description at https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_load_si512&expand=3345
> We need find a CLX to test it.

this should not cause a problem

PR #1083, opened 2020-02-25
Bitstream: refact input params and hide implement details to make code more readable.

xuguangxin:
this is a pr only improve code readability. 
It's no impact to speed or quality.


joelsole:
Please, add a description to the PR; mention explicitly the performance and speed impact (that's good practice, even if it is obvious for the implementer that there is none).

xuguangxin:
@joelsole , thanks for suggestion, done

EricLamVCD:
2 random clips from each resolution with at least one SC clip for 8 bit and 10 bit. 


Speed   Deviation | BD-Rate   Deviation
-- | --
0.32% | 0.00%

PR #1092, opened 2020-02-26
packetization_kernel: fix missed last frame

xuguangxin:
we will miss last frame if it is a show existing frame

test command line:
-i blue_sky_360p_60f.yuv -o rec.yuv -intra-period 200 -enc-mode 8 -n 25 -q 40 -lp 1 -lad 0 -b 1.ivf -w 640 -h 360

### DIFF
--- a/Source/Lib/Encoder/Codec/EbPacketizationProcess.c
+++ b/Source/Lib/Encoder/Codec/EbPacketizationProcess.c
@@ -623,6 +623,14 @@ static void release_frames(EncodeContext *encode_context_ptr, int frames) {
     encode_context_ptr->packetization_reorder_queue_head_index = get_reorder_queue_pos(encode_context_ptr, frames);
 }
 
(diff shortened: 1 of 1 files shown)


xuguangxin:
@hassount I tried 3 times to make this pass all build.
seems we have random failure for unit test.
see https://github.com/OpenVisualCloud/SVT-AV1/pull/1093/checks?check_run_id=469724506 for details.

EricLamVCD:
A total of 12 clips at 5 qp's each were tested. They were tested with -n 25 frames and -n 49 frames for 2 random clips for each resolution at 8 bit and 10 bit. 

The encoder worked fine for -n 25 and -n 49.

hassount:
> lgtm - maybe the two comments are not necessary and then, the braces of the two if statements with them can be removed as per the style guidelines.

@xuguangxin could you please address the above comments

xuguangxin:
thanks joelsole and hassount for help on this

PR #1095, opened 2020-02-26
Travis: disable coveralls for the moment

kylophone:
Since our coveralls is not reporting an accurate coverage number, we will disable the CI hook for now. When this is fixed, this commit can be reverted.


hassount:
@kylophone 
can we also disable the unit tests from travis as we have them in actions ? that would save us quite a bit of time each PR / push

PR #1094, opened 2020-02-26
Remove unnecessary memset when allocating YUV buffer

tianjunwork:
Reduce init time the more the higher resolution.

Signed-off-by: Jun Tian <jun.tian@intel.com>

closes #1088

PR #1096, opened 2020-02-27
Modify NoiseExtractTest

Cidana-Developers:
1. add memory zero to fix NoiseExtractTest failure after optimized eb_picture_buffer_desc_dctor ([PR1094](https://github.com/OpenVisualCloud/SVT-AV1/pull/1094))

### DIFF
--- a/test/NoiseExtractTest.cc
+++ b/test/NoiseExtractTest.cc
@@ -70,10 +70,15 @@ class ExtractFilterTest : public ::testing::Test {
         subsampling_y_ = (pbd_init_data.color_format >= EB_YUV422 ? 1 : 2) - 1;
 
         eb_picture_buffer_desc_ctor(&in_pic_, &pbd_init_data);
+        zero_pic(&in_pic_);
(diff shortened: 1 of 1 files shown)


silverbacknet:
As mentioned in the other commit, this just papers over the test failure. Whether the failure is because the encoding functions read out-of-bounds or the tests read out-of-bounds, the test should not be modified to ensure that out-of-bounds reads pass, the actual bug should be fixed.

Cidana-Developers:
> As mentioned in the other commit, this just papers over the test failure. Whether the failure is because the encoding functions read out-of-bounds or the tests read out-of-bounds, the test should not be modified to ensure that out-of-bounds reads pass, the actual bug should be fixed.

Thank you for kindly remind! We are thinking about the same thing. 

The test code is designed to verify the whole allocated buffer, all parts of data (include frame data, stride area and other part actually not in use) should be identical. Because the overflow might be happening inside the implementation, that’s the aim of unit test. That’s why we change the test code to zero the whole buffer instead of only check the available frame data.

PR #1099, opened 2020-02-28
fixed gm_down bitstream corruption issue

kelvinhu325:
global motion gm_down case gets bitstream corruption for global motion MV may be out of range after upscale by 2 in mode decision part, adding global motion MV range check to fix this issue.


hassount:
@magwyz Could you please review this fix?

magwyz:
@kelvinhu325 Why not clamping in `set_global_motion_field` the translation parameters as it is done here ?

https://github.com/OpenVisualCloud/SVT-AV1/blob/a56b273b864b504d571a4808697e83a5b7ec7b27/Source/Lib/Encoder/Codec/EbModeDecisionConfigurationProcess.c#L184

kelvinhu325:
> I agree with the fix. However, it makes more sense to move the fix inside as suggested by @magwyz

Hi @chkngit @magwyz ,  actually I considered adding it in set_global_motion_field before, if clamping it after upscaling in mode decision stage, this global MV is not accurate and may mislead winner MV generating, so moving it to gm searching stage may be better because it could go on 
 other gm MVs searching after gave up this out of range MV, how do you think about it?

PR #1101, opened 2020-02-28
16b pipeline support in decoder

nijilk:
16b pipeline support is added in decoder.

Encoder is verified to produce identical bit-streams for limited tests for enc-mode 0 and 8.

Decoder library changes are under DEC_16BIT_PIPELINE macro.
Common changes are under COMMON_16BIT macro.


nijilk:
@andrey-norkin @hassount 
Please review.
The check failures are present in previous commits as well.

PR #1110, opened 2020-03-02
Adopt sb 128

okhlif:
Address #1047


mafonsoN:
@okhlif So, this is enabling 128x128 SBs for resolutions lower than 720p, right? Any known compression efficiency improvements?

okhlif:
> @okhlif So, this is enabling 128x128 SBs for resolutions lower than 720p, right? Any known compression efficiency improvements?

No significant improvement for 360p  ( fast1-obj) , just to align all resolutions. ( not needed difference)

PR #1112, opened 2020-03-03
Me nsq opt

NaderMahdi:
## Description
Reduce the complexity of the motion estimation by avoiding the full_pel and sub_pel searches for NSQ blocks and inherit their motion vectors from SQ blocks. This feature is enabled only for screen content.

## Results
* BDrate 0.17% loss
* Speed 24.5% gain
* Memory 0%

InputSequence | Resolution | Bdrate PSNR-YUV | Bdrate SSIM-YUV | Encoding time   deviation
-- | -- | -- | -- | --
MINECRAFT_60f_420 | 1920x1080 | 0.09% | 0.10% | 22.02%
SlideEditing_1280x720_30_60f | 1280x720 | 0.08% | 0.13% | 18.59%
SlideShow_1280x720_20_60f | 1280x720 | 0.05% | 0.04% | 22.60%
wikipedia_420 | 1920x1080 | 0.45% | 0.42% | 34.66%
  | Average | 0.17% | 0.17% | 24.47%

PR #1118, opened 2020-03-04
Shutdown threads in a graceful way

tianjunwork:
## Description
Avoid using TerminateThread/pthread_cancel which causes issue with multiple continues init/deinit of encoder lib. A customer reported this issue before to SVT-HEVC. This is the fix of it by sending quit signal from main thread to all the kernels. Kernel will break out from the loop once quit signal is set.
Ported from SVT-HEVC. Author: jing.b.li@intel.com

## Results
* BDrate: same
* Speed: same
* Memory: Around 1200(number of fifo object) * EbBool(uint8_t) = ~1200 byte mem increase

Signed-off-by: Jun Tian <jun.tian@intel.com>

### DIFF
--- a/Source/Lib/Common/Codec/EbSystemResourceManager.c
+++ b/Source/Lib/Common/Codec/EbSystemResourceManager.c
@@ -377,6 +377,7 @@ static EbErrorType eb_object_wrapper_ctor(EbObjectWrapper *wrapper, EbSystemReso
     ret            = object_creator(&wrapper->object_ptr, object_init_data_ptr);
     if (ret != EB_ErrorNone) return ret;
     wrapper->release_enable      = EB_TRUE;
+    wrapper->quit_signal         = EB_FALSE;
     wrapper->system_resource_ptr = resource;
     wrapper->object_destroyer    = object_destroyer;
     return EB_ErrorNone;
--- a/Source/Lib/Common/Codec/EbSystemResourceManager.c
+++ b/Source/Lib/Common/Codec/EbSystemResourceManager.c
@@ -76,6 +76,21 @@ static EbErrorType eb_fifo_pop_front(EbFifo *fifoPtr, EbObjectWrapper **wrapper_
     return return_error;
 }
 
+static EbErrorType eb_fifo_shutdown(EbFifo *fifo_ptr) {
+
+    EbErrorType return_error = EB_ErrorNone;
+
+    // Acquire lockout Mutex
+    eb_block_on_mutex(fifo_ptr->lockout_mutex);
+    fifo_ptr->quit_signal = EB_TRUE;
+    // Release Mutex
+    eb_release_mutex(fifo_ptr->lockout_mutex);
+    //Wake up the waiting process if any
+    eb_post_semaphore(fifo_ptr->counting_semaphore);
+
+    return return_error;
+}
+
 static void eb_circular_buffer_dctor(EbPtr p) {
     EbCircularBuffer *obj = (EbCircularBuffer *)p;
     EB_FREE(obj->array_ptr);
@@ -377,7 +392,6 @@ static EbErrorType eb_object_wrapper_ctor(EbObjectWrapper *wrapper, EbSystemReso
     ret            = object_creator(&wrapper->object_ptr, object_init_data_ptr);
     if (ret != EB_ErrorNone) return ret;
     wrapper->release_enable      = EB_TRUE;
-    wrapper->quit_signal         = EB_FALSE;
     wrapper->system_resource_ptr = resource;
     wrapper->object_destroyer    = object_destroyer;
     return EB_ErrorNone;
@@ -468,6 +482,15 @@ EbFifo *eb_system_resource_get_producer_fifo(const EbSystemResource *resource_pt
 
 EbFifo *eb_system_resource_get_consumer_fifo(const EbSystemResource *resource_ptr, uint32_t index) {
     return eb_muxing_queue_get_fifo(resource_ptr->full_queue, index);
+}
+
+EbErrorType eb_shutdown_process(const EbSystemResource *resource_ptr) {
+    //notify all consumers we are shutting down
+    for (unsigned int i = 0; i < resource_ptr->full_queue->process_total_count; i++) {
+        EbFifo *fifo_ptr = eb_system_resource_get_consumer_fifo(resource_ptr, i);
+        eb_fifo_shutdown(fifo_ptr);
+    }
+    return EB_ErrorNone;
 }
 
 /*********************************************************************
@@ -613,7 +636,12 @@ EbErrorType eb_get_full_object(EbFifo *full_fifo_ptr, EbObjectWrapper **wrapper_
     // Acquire lockout Mutex
     eb_block_on_mutex(full_fifo_ptr->lockout_mutex);
 
-    eb_fifo_pop_front(full_fifo_ptr, wrapper_dbl_ptr);
+    if (!full_fifo_ptr->quit_signal) {
+        eb_fifo_pop_front(full_fifo_ptr, wrapper_dbl_ptr);
+    } else {
+        *wrapper_dbl_ptr = NULL;
+        return_error = EB_NoErrorFifoShutdown;
+    }
 
     // Release Mutex
     eb_release_mutex(full_fifo_ptr->lockout_mutex);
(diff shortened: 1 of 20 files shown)


tianjunwork:
Hi @lijing0010 @xuguangxin please kindly review this PR. Thank you.

Source/Lib/Common/Codec/EbSystemResourceManager.c
```diff
@@ -377,6 +377,7 @@ static EbErrorType eb_object_wrapper_ctor(EbObjectWrapper *wrapper, EbSystemReso
     ret            = object_creator(&wrapper->object_ptr, object_init_data_ptr);
     if (ret != EB_ErrorNone) return ret;
     wrapper->release_enable      = EB_TRUE;
+    wrapper->quit_signal         = EB_FALSE;
```
xuguangxin: Instead of putting this in the wrapper, could we put this in the EbSystemResource ?

Source/Lib/Common/Codec/EbSystemResourceManager.h
```diff
@@ -297,6 +300,18 @@ extern EbErrorType eb_get_full_object_non_blocking(EbFifo *          full_fifo_p
      *      pointer to EbObjectWrapper to be released.
      *********************************************************************/
 extern EbErrorType eb_release_object(EbObjectWrapper *object_ptr);
+
+#define EB_SEND_END_OBJ(fifo_ptr, count)               \
```
xuguangxin: If some one get one object like sps, and never release it, will this get a dead lock?

Source/Lib/Encoder/Codec/EbCdefProcess.c
```diff
@@ -513,6 +513,7 @@ void *cdef_kernel(void *input_ptr) {
     for (;;) {
         // Get DLF Results
         eb_get_full_object(context_ptr->cdef_input_fifo_ptr, &dlf_results_wrapper_ptr);
```
xuguangxin: How about 
1. return some status in eb_get_full_object. and we check the status.
2. use EB_GET_FULL_OBJECT macro instead of get and check.

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -1732,6 +1732,26 @@ __attribute__((visibility("default")))
 EB_API EbErrorType eb_deinit_encoder(EbComponentType *svt_enc_component){
     if(svt_enc_component == NULL)
         return EB_ErrorBadParameter;
+
+    EbEncHandle *handle = (EbEncHandle*)svt_enc_component->p_component_private;
+    if (handle) {
+        EB_SEND_END_OBJ(eb_system_resource_get_producer_fifo(handle->input_buffer_resource_ptr, 0),                    EB_ResourceCoordinationProcessInitCount);
```
xuguangxin: how about add one function named eb_system_resource_shut_down. Then we can hide all eb_system_resource_get_producer_fifo details in the function

Source/Lib/Common/Codec/EbSystemResourceManager.c
tianjunwork: what is the benefit of doing that?

Source/Lib/Common/Codec/EbSystemResourceManager.h
tianjunwork: if so, we need to debug and fix dead lock, right?

Source/Lib/Encoder/Codec/EbCdefProcess.c
tianjunwork: You can't actually debug them (in the sense of stepping into a 'multiline' macro function), you can only step past them comparing the before and after state. This is one of the major reasons why complicated macro functions can be a bad idea.
Not only step into, the 'data breakpoint' which traces the content change of buffer pointer also doesn't work with Macro code.
I don't prefer to use Macro with many lines of code or functions inside.

Source/Lib/Encoder/Globals/EbEncHandle.c
tianjunwork: do you mean eb_deinit_encoder calls eb_system_resource_shut_down? Move all these EB_SEND_END_OBJ into eb_system_resource_shut_down?
eb_system_resource_get_producer_fifo still show up in this file. how should I hide them?

hassount:
@tianjunwork, thank you for the contribution. In order to help streamline the testing process, could you please elaborate in the description section with more details about what the commit is about, including a performance section to state whether the bitstream is expected to change or not. Something similar to this #1112

Source/Lib/Common/Codec/EbSystemResourceManager.h
xuguangxin: It's normal usage, not a bug, suppose someone needs to compare current sps with previous sps. So she/he needs to keep previous sps.

Source/Lib/Common/Codec/EbSystemResourceManager.c
xuguangxin: If we put this to EbSystemResource , everybody can get the quit signal immediately. we do not need to wait empty object.
Wait for empty objects may introduce delay. Some objects acquired at every beginning of the pipeline, released at the end of the pipeline.  get new empty may need to wait a very long time.

Source/Lib/Encoder/Codec/EbCdefProcess.c
xuguangxin: agree, but we already introduced a macro EB_CHECK_END_OBJ. and EB_CHECK_END_OBJ always comes with one single line eb_get_full_object. :)

Source/Lib/Encoder/Globals/EbEncHandle.c
xuguangxin: EB_SEND_END_OBJ  is implementation details for EbSystemResource, external file EbEecHandle.c does not need to know these details.
This will give you flexibility. If you want to change the EB_SEND_END_OBJ   mechanism, you do not need change EbEncHandle.c.

Source/Lib/Common/Codec/EbSystemResourceManager.c
tianjunwork: `If we put this to EbSystemResource , everybody can get the quit signal immediately` ==> not sure how to do that. If a kernel is using a system res right now, how to quit the the loop? There must be a check point, right?

Source/Lib/Encoder/Codec/EbCdefProcess.c
tianjunwork: pls suggest pseudo code to help communicate.

Source/Lib/Encoder/Globals/EbEncHandle.c
tianjunwork: Do you mean add `eb_kernel_shut_down` in EbSystemResourceManager.h/c to hide EB_SEND_END_OBJ?
Actually, all ` the system res ptr` and their size are showed up in this file, it is easier to read and trace the code.

Source/Lib/Common/Codec/EbSystemResourceManager.h
tianjunwork: fifo used in EB_SEND_END_OBJ are not accessible to lib user through api.

tianjunwork:
> @tianjunwork, thank you for the contribution. In order to help streamline the testing process, could you please elaborate in the description section with more details about what the commit is about, including a performance section to state whether the bitstream is expected to change or not. Something similar to this #1112

done

EricLamVCD:
Random 3 clips per each resolution for 8 bit and 10 bit
Speed   Deviation | BD-Rate   Deviation
-- | --
-3.08% | 0.00%

Source/Lib/Encoder/Codec/EbCdefProcess.c
xuguangxin: how about the last 3 commits form https://github.com/xuguangxin/SVT-AV1/tree/shutdown?
hide the impelment details can make your easily rerfact your code

hassount:
@tianjunwork @xuguangxin is this ready to go?

tianjunwork:
> @tianjunwork @xuguangxin is this ready to go?

Not yet. I am addressing the review comment.

tianjunwork:
Hi @xuguangxin , thank you for the commits on your branch. I added some comments.
I agree with `refact: hide EB_SEND_END_OBJ in eb_system_resource_shutdown` is a cleaner.
Have some concern with the rest.

Source/Lib/Encoder/Codec/EbCdefProcess.c
xuguangxin: @tianjunwork , no worry, just take the first commit.
With it, we can easily refact the code after we faced the real problem.

tianjunwork:
Hi @xuguangxin , I added your suggested changes. Thank you again.
I measured the time between commit 2 and 1, the difference is very minimum. But from logic vise, commit 2 is better. Besides, memory for quit_signal is smaller.

xuguangxin:
thanks @tianjunwork . we are ready to go.

PR #1119, opened 2020-03-04
Fix lambda for HBD

okhlif:
### Description
_________________________________________________________________________________

- Fix lambda for HBD (hbd mode 0, 1 and 2)
- Fix the mismatch of using 10bit vs 8bit lambda depending on the hbd-mode and/or the kernel.
### Author(s)
_________________________________________________________________________________
@okhlif 

### Type of change
_________________________________________________________________________________
Enhancement.
### Tests and performance
_________________________________________________________________________________
Data generated on the 14 720p true 10bit in context in enc-mode 0.

Speed Deviation | BD-Rate Deviation
-- | --
no speed diff | -0.45%

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -34,6 +34,18 @@
 extern "C" {
 #endif
 
+#define OMARK_LAMBDA                1 // 2. fix lambda calculation for HBD0
+#define OMARK_HBD1_CDEF             1 // 3. fix CDEF lambda for hbd1/2
+#define OMARK_HBD0_ED               1 // 4. fix ED lambda for hbd0
+#define OMARK_HBD0_IFS              1 // 5. fix IFS lambda for hbd0
+#define OMARK_HBD0_MD               1 // 5. fix MD lambda for hbd0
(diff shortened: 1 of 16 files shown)


joelsole:
Changes look good. Could you address the following?
- add a description of what is actually being fixed
- consolidate the macros into three, two or less
- name the macros after the purpose of the new code

okhlif:
> Changes look good. Could you address the following?
> 
> * add a description of what is actually being fixed
> * consolidate the macros into three, two or less
> * name the macros after the purpose of the new code

Thank you Joel. I added in the description what was fixed in general. 
For the macros, I can add a macro that will be enable/disable all. I am trying to keep it as close as possible to the cs2 branch so it is easier to debug. Would that work?

joelsole:
By 'debug', do you mean closing potential misalignments between cs2 and master? I don't think adding yet another macro helps.

BTW - when were the lambda values derived? Are they coming directly from libaom?

okhlif:
> By 'debug', do you mean closing potential misalignments between cs2 and master? I don't think adding yet another macro helps.
> 
> BTW - when were the lambda values derived? Are they coming directly from libaom?

Yes correct potential misalignment, especially if I will have to rebase. The lambda tables come from libaom.

PR #1123, opened 2020-03-05
fixed gm_down bitstream corruption issue

kelvinhu325:
global motion gm_down case gets bitstream corruption for global motion MV may be out of range after upscale by 2 in mode decision part, adding global motion MV range clamping change to fix this issue.


hassount:
@kylophone as you've approved #1099 being to the cs2 branch I take it that this one is good to go (the same to master) ?

PR #1125, opened 2020-03-06
fixed run2run issue caused by uint32_t recon_offset

kelvinhu325:
The MV offset recon_offset is taken as unsigned value in function spatial_full_distortion_kernel_avx2(), then minus value MV like (-36, -10) generate very big recon_offset for forcing to be unsigned int type, it causes the recon buffer points to far away invalid memory, it should change to  signed value int32_t type.
Quality and speed are impacted only when functions spatial_full_distortion_kernel and full_distortion_kernel16bit are called, very slight impact.

Speed Deviation | BD-Rate Deviation
-- | --
-- | 0.00%


hassount:
@kelvinhu325 
can you please rebase and fix the build so that we can merge this ?

kelvinhu325:
> @kelvinhu325
> can you please rebase and fix the build so that we can merge this ?

Hi @hassount , I added the SpatialFullDistortionTest.cc change, it looks like a ULT test and the recon_offset type should be also fixed, my new commit is submitted, but don't know why the system does not re-build it.

kelvinhu325:
hi @hassount , after my more changes on unit tests and full_distortion_kernelbit16, it shows all tests passed now.

PR #1126, opened 2020-03-06
Bug fix in common inv_transform sse3 functions and decoder LF-MT

nijilk:
Bug fixes in inv_transform sse3 functions related to identity transform case
    Bug fix in decoder LF MT path


nijilk:
@andrey-norkin @joelsole @hassount 
Please review

joelsole:
With this patch, the decoder is now conformant for all my test streams (~800), fixing the issue previously existing in 19 cases.

PR #1132, opened 2020-03-12
Fix the r2r

lijing0010:
Contains 2 cases
    - When skiping stage1/2, the full cost is used directly without been
    initialized
    - At the boundary when it's not a complete super block. We may only
    use part of the blocks in MD and didn't set its parent blocks
    properly. it will generate the wrong CDF context for the next SB

### DIFF
--- a/Source/Lib/Encoder/Codec/EbProductCodingLoop.c
+++ b/Source/Lib/Encoder/Codec/EbProductCodingLoop.c
@@ -1318,6 +1318,12 @@ void fast_loop_core(ModeDecisionCandidateBuffer *candidate_buffer, PictureContro
         1,
         context_ptr->intra_luma_left_mode,
         context_ptr->intra_luma_top_mode);
+
+#if R2R_FIX
+    // Init full cost in case we by pass stage1/stage2
+    if (context_ptr->md_staging_mode == MD_STAGING_MODE_0)
+        *(candidate_buffer->full_cost_ptr) = *(candidate_buffer->fast_cost_ptr);
+#endif
 }
 #if NICS_CLEANUP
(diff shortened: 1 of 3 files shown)


Source/Lib/Encoder/Codec/EbProductCodingLoop.c
```diff
@@ -1318,6 +1318,13 @@ void fast_loop_core(ModeDecisionCandidateBuffer *candidate_buffer, PictureContro
         1,
         context_ptr->intra_luma_left_mode,
         context_ptr->intra_luma_top_mode);
+
+#if R2R_FIX
+    // Init full cost in case we by pass stage1/stage2
+    if (context_ptr->md_staging_mode == MD_STAGING_MODE_0) {
+        *(candidate_buffer->full_cost_ptr) = *(candidate_buffer->fast_cost_ptr);
+    }
```
joelsole: no braces

mafonsoN:
@lijing0010 Can you clarify what r2r cases this patch will fix (enc-modes, etc)? Are these are the sources of r2rs we that have been found in the current master?

Source/Lib/Encoder/Codec/EbProductCodingLoop.c
lijing0010: done

lijing0010:
> @lijing0010 Can you clarify what r2r cases this patch will fix (enc-modes, etc)? Are these are the sources of r2rs we that have been found in the current master?

This will resolve 2 r2r in current master, not all of them.
One is running in M6 or M7, has a small chance.(caused by skipping the staging1/2)
Another is in M3(where SB is 128 and height is not 128 aligned), could also happen at other non sb_size aligned resolution, the chance of this one is low(the logic is: blk_idx not right->wrong blk size->wrong entropy ctx for the next super block->wrong value when evaluating bits for "split"->wrong MD results)

PR #1143, opened 2020-03-16
Fix the r2r caused by wrong mv in warp case

lijing0010:


hassount:
@lijing0010 
can you please address the unit test failing cases ?

lijing0010:
> Could you explain how this solves the r2r?
> Should function chroma_plane_warped_motion_prediction_sub8x8 be changed accordingly?

It will have a small chance to cause enc/dec mismatch too.
The pred_direction can be 0/1/2 while mv_unit->mv is an array of 2. 
So when the pred_direction is BI_PRED,  codes like "mv_l0.col = mv_unit->mv[mv_unit->pred_direction].x" will be visiting wrong memory

PR #1145, opened 2020-03-17
Fix r2r for -enable-overlays cases

### LINKED ISSUE #929: r2r mismatch when overlay is enabled
I was testing the option `enable-overlays` and found r2r mismatch (both release and debug have this issue) using the following command line.
./SvtAv1EncApp.exe -i BasketballDrill_832x480_50.yuv -w 832 -h 480 -b test.ivf -enc-mode 8 -n 200 -rc 0 -q 24 -intra-period 31 -fps-num 50 -fps-denom 1 -enable-overlays 1

The version of SVT-AV1 I used was v0.8.0 (af16f709)

kelvinhu325:
## Description

In includes 4 part changes below:
1, to set int32_t recon_offset in half pixel spatial_full_distortion_kernel()
Found it gets crash in a special case  for invalid memory access in function spatial_full_distortion_kernel() function, and root cause the input big recon_offset value does not reasonable, actually it’s a minus value and converted to be unsigned int, which makes it taken as very big value, this error is not fatal for it does not always access invalid memory;
2, to initialized context_ptr->p_sb_best_mv[] in beginning of function motion_estimate_sb()
Found the ME output MVs always get r2r and something the MV value could be very big, then checked the ME module and found this overlay test is very special, it generates search_window size 0x0 of one block in full pixels searching part, then the context_ptr->p_sb_best_mv[] array is not updated at all, which is used to output MVs to MD, then I memset it to 0 and found output MVs do not that much “r2r” as before.
3, copying input_padded_picture_ptr in perform_simple_picture_analysis_for_overlay()
to avoid motion_estimate_sb use dirty input_padded_picture_ptr data
Found distortion value calculated in function nxm_sad_kernel() function is always gets r2r, double checked the input pixels data, it is from input_padded_picture_ptr, which is also gets r2r, then root caused the function perform_simple_picture_analysis_for_overlay() only updates pixels data in input_picture_ptr and forget updates input_padded_picture_ptr’s. 
4, increase interpolate height by 1 in interpolate_search_region_avc() to
avoid r2r in last line pixels data in pos_h_buffer and pos_j_buffer
After above 3 changes, the full pixels test(setting -subpel 0 in cmdline) does not get r2r now, then focused on sub_pixel searching code in ME for ME output MVs still get r2r, sa_width and sa_height are 0 for overlay picture, it results in only 63x63 pixels are interpolated in interpolate_search_region_avc(), while distortion calculating r2r needs 64x64 interpolated pixels, fixing it by limit interpolate win to be min 64x64 for fixing overlay r2r, this change does not impact non-overlay pictures for sa_width and sa_height is 0 for only overlay picture.

## Issue:
Fixes #929

### DIFF
--- a/Source/Lib/Encoder/Codec/EbMotionEstimation.c
+++ b/Source/Lib/Encoder/Codec/EbMotionEstimation.c
@@ -4733,7 +4733,7 @@ void interpolate_search_region_avc(
             context_ptr->pos_b_buffer[list_index][ref_pic_index],
             context_ptr->interpolated_stride,
             search_area_width_for_asm,
-            search_area_height + ME_FILTER_TAP,
+            search_area_height + ME_FILTER_TAP + 1,
             context_ptr->avctemp_buffer,
             EB_FALSE,
             2,
@@ -4748,7 +4748,7 @@ void interpolate_search_region_avc(
             context_ptr->pos_h_buffer[list_index][ref_pic_index],
             context_ptr->interpolated_stride,
             search_area_width_for_asm,
-            search_area_height + 1,
+            search_area_height + 2,
             context_ptr->avctemp_buffer,
             EB_FALSE,
             2,
@@ -4763,7 +4763,7 @@ void interpolate_search_region_avc(
             context_ptr->pos_j_buffer[list_index][ref_pic_index],
             context_ptr->interpolated_stride,
             search_area_width_for_asm,
-            search_area_height + 1,
+            search_area_height + 2,
             context_ptr->avctemp_buffer,
             EB_FALSE,
             2,
@@ -4851,8 +4851,8 @@ static void pu_half_pel_refinement(
 
     int16_t x_mv           = _MVXT(*p_best_MV);
     int16_t y_mv           = _MVYT(*p_best_MV);
-    int16_t x_search_index = (x_mv >> 2) - x_search_area_origin;
-    int16_t y_search_index = (y_mv >> 2) - y_search_area_origin;
+    int32_t x_search_index = (x_mv >> 2) - x_search_area_origin;
+    int32_t y_search_index = (y_mv >> 2) - y_search_area_origin;
 
     (void)scs_ptr;
     (void)encode_context_ptr;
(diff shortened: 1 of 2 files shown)


Source/Lib/Encoder/Codec/EbMotionEstimation.c
```diff
@@ -4763,7 +4763,7 @@ void interpolate_search_region_avc(
             context_ptr->pos_j_buffer[list_index][ref_pic_index],
             context_ptr->interpolated_stride,
             search_area_width_for_asm,
-            search_area_height + 1,
+            search_area_height + 2,
```
mafonsoN: Can you explain why you have to increase the interpolated area if your search area is known?

kelvinhu325: > Can you explain why you have to increase the interpolated area if your search area is known?

Hi @mafonsoN , sure, the pos_b_buffer, pos_h_buffer and pos_j_buffer are generated here and are used in function pu_half_pel_refinement() to get best distortion and generate winner MV inside neighbor 8 half-pixels points, and the MV (search_region_index) makes it point out of range 1 line in special case, which causes dirty data is used to calculate SAD and r2r appears, one example is that input MV is (mvx=0, mvy=2), the bottom line data of 64x64 block gets r2r data, back tracing it and found the pos_h_buffer interpolating window range is 80x68, half pixel filter is based on 6 TAP, it needs top 2 and bottom 3 additional pixels, total 2+64+3=69 pixels, similar requirement to pos_h_buffer and pos_j_buffer, so I inreased search height by 1 for them, then the r2r gets disappeared.

kelvinhu325: > @kelvinhu325 is the issue related to overlay pictures?

This issue is only for overlay picture, double checked it, the sa_width and sa_height are 0(interpolate 63x63 pixels) for overlay picture, which is very special, it means search window is 0x0, while it still calculate distortion for sub pixel, looks not reasonable, I tried to let the sa_width and sa_width to have min value 1, this change also can fix this r2r issue, looks it's better.

kelvinhu325:
@anaghdin and @mafonsoN @hassount , add fixing for interpolate part for removing non-overlay picture impact, pls review it.
sa_width and sa_height are 0 for overlay picture, it results in only 63x63 pixels are interpolated in interpolate_search_region_avc(), while distortion calculating r2r needs 64x64 interpolated pixels, fixing it by limit interpolate win to be min 64x64 for fixing overlay r2r, this change does not impact non-overlay pictures for sa_width and sa_height is 0 for only overlay picture.

hassount:
@kelvinhu325 can you please commit this change to the beyond-cs2 branch ?

kelvinhu325:
> @kelvinhu325 can you please commit this change to the beyond-cs2 branch ?

Done:)

PR #1138, opened 2020-03-13
16bit pipeline support

ttrigui:
16bit pipeline support for EncDec
Configuration parameter: -16bit-pipeline ( 1: enabled, 0: disabled )

PR #1146, opened 2020-03-18
fix 2 16bits memory leaks 

xuguangxin:
**Test command:**
../Bin/Debug/SvtAv1EncApp -i /bits/480x264.y4m  -n 1 -b test.ivf

**No fps or quality impact.**

Hi @hassount ,
is it possible to add some memory leak check for ci? else we always can get some memory leak.

thanks

### DIFF
--- a/Source/Lib/Encoder/Codec/EbPictureControlSet.c
+++ b/Source/Lib/Encoder/Codec/EbPictureControlSet.c
@@ -453,9 +453,6 @@ EbErrorType picture_control_set_ctor(PictureControlSet *object_ptr, EbPtr object
                eb_recon_picture_buffer_desc_ctor,
                (EbPtr)&input_pic_buf_desc_init_data);
     }
-    EB_NEW(object_ptr->recon_picture16bit_ptr,
-           eb_recon_picture_buffer_desc_ctor,
-           (EbPtr)&coeff_buffer_desc_init_data);
     // Film Grain Picture Buffer
     if (init_data_ptr->film_grain_noise_level) {
         if (is_16bit) {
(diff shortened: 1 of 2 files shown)


hassount:
@xuguangxin thank you for the PR. I believe @ePirat is working on adding memory check tools to the ci

hassount:
@ttrigui are these the temp buffers needed for the 16bit pipeline ?

ttrigui:
> @ttrigui are these the temp buffers needed for the 16bit pipeline ?

Yes these buffers are needed for 16bit pipeline.
@xuguangxin Can we leave the allocation and make sure they are freed properly to avoid the leak?

xuguangxin:
@ttrigui , this is a double allocation problem. 
the 16 bits buffer already allocated by 
https://github.com/OpenVisualCloud/SVT-AV1/pull/1146/files#diff-b250b2886176e4b9f205df3e3acde8fdL448
and
https://github.com/OpenVisualCloud/SVT-AV1/pull/1146/files#diff-da85d23c9d6b6642157875844db1f83aL151

The question is, do we need reference_picture16bit when picture_buffer_desc_init_data_16bit_ptr.bit_depth == EB_8BIT?
thanks

ttrigui:
> @ttrigui , this is a double allocation problem.
> the 16 bits buffer already allocated by
> https://github.com/OpenVisualCloud/SVT-AV1/pull/1146/files#diff-b250b2886176e4b9f205df3e3acde8fdL448
> and
> https://github.com/OpenVisualCloud/SVT-AV1/pull/1146/files#diff-da85d23c9d6b6642157875844db1f83aL151
> 
> The question is, do we need reference_picture16bit when picture_buffer_desc_init_data_16bit_ptr.bit_depth == EB_8BIT?
> thanks

we have new a feature that will be introduced soon to support 16bit pipeline regardless of the input bit-depth. So for an 8bit input we will have two choices: use a 16bit pipeline (i.e need to use 16bit buffers) or use the 8bit path (default path). These buffers are needed if is16bit-pipeline is enabled.
The immediate solution can be a check on "encoder_16bit_pipeline" only allocate when enabled.

hassount:
This is addressed as part of #1138

PR #1155, opened 2020-03-26
support c only build for non x86 platform 

ttrigui:
add the ability to detect the platform support: if x86 is not detected c only code will be used


Source/Lib/Encoder/Codec/EbSourceBasedOperationsProcess.c
```diff
@@ -128,14 +130,15 @@ void *source_based_operations_kernel(void *input_ptr) {
...
             _mm_prefetch((const char *)y_mean_ptr, _MM_HINT_T0);
+#endif
             uint8_t *cr_mean_ptr = pcs_ptr->cr_mean[sb_index];
             uint8_t *cb_mean_ptr = pcs_ptr->cb_mean[sb_index];
-
+#ifdef ARCH_X86
             _mm_prefetch((const char *)cr_mean_ptr, _MM_HINT_T0);
             _mm_prefetch((const char *)cb_mean_ptr, _MM_HINT_T0);
-
+#endif
```
tszumski: I looked into _mm_prefetch() description and i think that we can remove usage of it (as it is only suggestion to load data into cache and can be skipped), then the code will be simplified
```
        for (sb_index = 0; sb_index < sb_total_count; ++sb_index) {
            context_ptr->y_mean_ptr  = pcs_ptr->y_mean[sb_index];
            context_ptr->cr_mean_ptr = pcs_ptr->cr_mean[sb_index];
            context_ptr->cb_mean_ptr = pcs_ptr->cb_mean[sb_index];
            if (pcs_ptr->sb_params_array[sb_index].is_complete_sb) context_ptr->complete_sb_count++;
        }
```

Source/Lib/Encoder/Codec/EbTemporalFiltering.c
```diff
@@ -565,17 +567,20 @@ static void create_me_context_and_picture_control(
                          ->buffer_y[buffer_index + sb_row * input_picture_ptr_central->stride_y])),
                   BLOCK_SIZE_64 * sizeof(uint8_t));
     }
-
+#ifdef ARCH_X86
```
tszumski: Same as above

Source/Lib/Encoder/Codec/EbSourceBasedOperationsProcess.c
ttrigui: I remember seeing some speed improvement when using prefetch in HEVC. I'm not sure about the performance now if no speed is recorded we will remove it otherwise let's keep it.

hassount:
@EwoutH @1480c1 FYI

EwoutH:
@ttrigui Thanks a lot for this! Any specific reason why it's needed to roll back to GCC 7 in Travis?

hassount:
> @ttrigui Thanks a lot for this! Any specific reason why it's needed to roll back to GCC 7 in Travis?

@EwoutH, seems that the ARM vms are having issues finding the gcc9 package we can go there once it's fixed. do you think we can port these tests to actions ?

EwoutH:
Will give it a try

Source/Lib/Common/Codec/EbDefinitions.h
```diff
@@ -30,6 +30,17 @@
 #define inline
 #endif
 
+#if defined(_MSC_VER) //&& defined(_M_IX86)
```
ePirat, line 33: Why is the `//&& defined(_M_IX86)` here?

Source/Lib/Common/Codec/EbIntraPrediction.c
```diff
@@ -22,12 +22,36 @@
 #include "EbIntraPrediction.h"
 #include "EbModeDecisionProcess.h"
 #include "common_dsp_rtcd.h"
-#include "EbIntraPrediction_AVX2.h"
+//#include "EbIntraPrediction_AVX2.h"
```
ePirat, line 25: If this is not needed, remove it instead of commenting it out. Even if it might be relevant at some point, there is always the git history one can check.

andrey-norkin:
Why is the Unit test failing?

ttrigui:
> Why is the Unit test failing?

re-basing should fix the UT issue. A commit was lately merged to fix a UT problem.

Source/Lib/Common/Codec/EbDefinitions.h
ttrigui: not needed. removed.

Source/Lib/Common/Codec/EbIntraPrediction.c
ttrigui: done.

hassount:
@andrey-norkin is this good to go on your side ?

hassount:
@ttrigui can you please rebase and solve the conflict?
@andrey-norkin this seems to be a blocker to the xcode project work, is this good to go ?

hassount:
@ttrigui 
based on a discussion we just had in the open dev meeting, could you please remove the allowed to fail flag on the ARM builds so that we don't break them moving forward ?

1480c1:
Nice

EwoutH:
Nice indeed!

PR #1162, opened 2020-03-29
EbFileUtils: fix handling of a 0 frame unit size

ePirat:
When reading Annex-B OBUs in obudec_read_temporal_unit, when reading
a frame unit size of 0, it would call realloc with 0, which has not
well defined behavior.


ePirat:
The failed CI builds so far are due to failure to connect to `azure.archive.ubuntu.com` when fetching packages.

PR #1166, opened 2020-03-29
EbMotionEstimation: add assert in default case

ePirat:
This case should never happen, add an assert to ensure it would be
easier detected if it happens in debug builds.

Additionally this might help the static analyzer to figure out that
this does not normally happen.

### DIFF
--- a/Source/Lib/Encoder/Codec/EbMotionEstimation.c
+++ b/Source/Lib/Encoder/Codec/EbMotionEstimation.c
@@ -8,6 +8,7 @@
  */
 
 #include <stdio.h>
+#include <inttypes.h>
 
 #include "aom_dsp_rtcd.h"
 #include "EbPictureControlSet.h"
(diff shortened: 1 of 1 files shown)


Source/Lib/Encoder/Codec/EbMotionEstimation.c
```diff
@@ -6535,7 +6535,9 @@ static void set_quarter_pel_refinement_inputs_on_the_fly(uint8_t * pos_Full, //[
 
         break;
 
-    default: break;
+    default:
+        assert(0);
```
hassount: Could we add an Error message explaining why are we asserting ?

ePirat: Added one to the assert. Or do you want one using the logging function? If so, whats the correct logging call to use here?

hassount: the logging function would be better I think. I believe this would be an SVT_ERROR.

ePirat: Done.

PR #1168, opened 2020-03-30
Fix windows debug build and fix typo

tszumski:
Windows debug build shows: 
Run-Time Check Failure #2 - Stack around the variable 'message' was corrupted.


tszumski:
Should I change this PR as in #1171, or remove fix(then #1171 will be also merged) ?

hassount:
I think either remove it and let #1171 do it, or port the change from #1171 and add mark as a coauthor then we can close #1171

tszumski:
I replaced "magic numbers" as in #1171, so this changes can be merged and PR #1171 closed

PR #1171, opened 2020-03-30
message buffer too small

intelmark:
Increasing size of message buffer to hold messages that can be
WARNING_LENGTH in size (expanding from 50 to 100 bytes)
e.g. enc_mode will be deprecated soon, please use --preset

Signed-off-by: Mark Feldman <mark.feldman@intel.com>


ePirat:
Once #1168 is merged this will conflict with it, but this PR seems like the better fix for it, at least at first glance, as it does not sue some arbitrary hardcoded magic number.

hassount:
@tszumski seems that this PR is still valid ?

tszumski:
Thats the same changes as in #1168, this PR can be closed without merge

PR #1173, opened 2020-03-31
Port cs2 settings master latest

okhlif:
add missing feature/settings from cs2


chkngit:
Debugging CI:
fr MSVC builds shard3, the failed command is   SvtAv1EncApp -i kirland_640_480_30.yuv -w 640 -h 480 -bit-depth 8 -colour-space 420 -enc-mode 5 -b output.ivf -o recon.yuv

==> tried this in standalone app, the encoder is rejecting the param -colour-space
by removing that param, the encode is completely normal
@hassount  should we just remove that param from the E2E test?

PR #1154, opened 2020-03-25
valigrnd fix for mode_decision_update_neighbor_arrays

tszumski:
Issue #733 
This PR fixes:

```
Conditional jump or move depends on uninitialised value(s)
   at 0x504AE27: mode_decision_update_neighbor_arrays (EbProductCodingLoop.c:107)
   by 0x504BD5F: md_update_all_neighbour_arrays (EbProductCodingLoop.c:743)
   by 0x504BDEF: md_update_all_neighbour_arrays_multiple (EbProductCodingLoop.c:762)
   by 0x50590B2: mode_decision_sb (EbProductCodingLoop.c:8433)
   by 0x4FD3D3F: enc_dec_kernel (EbEncDecProcess.c:3104)
   by 0x5EA5DD4: start_thread (in /usr/lib64/libpthread-2.17.so)
   by 0x64B9EAC: clone (in /usr/lib64/libc-2.17.so)

Use of uninitialised value of size 8
   at 0x504AFA5: mode_decision_update_neighbor_arrays (EbProductCodingLoop.c:164)
   by 0x504BD5F: md_update_all_neighbour_arrays (EbProductCodingLoop.c:743)
   by 0x504BDEF: md_update_all_neighbour_arrays_multiple (EbProductCodingLoop.c:762)
   by 0x50590B2: mode_decision_sb (EbProductCodingLoop.c:8433)
   by 0x4FD3D3F: enc_dec_kernel (EbEncDecProcess.c:3104)
   by 0x5EA5DD4: start_thread (in /usr/lib64/libpthread-2.17.so)
   by 0x64B9EAC: clone (in /usr/lib64/libc-2.17.so)

Use of uninitialised value of size 8
   at 0x504B026: mode_decision_update_neighbor_arrays (EbProductCodingLoop.c:182)
   by 0x504BD5F: md_update_all_neighbour_arrays (EbProductCodingLoop.c:743)
   by 0x504BDEF: md_update_all_neighbour_arrays_multiple (EbProductCodingLoop.c:762)
   by 0x50590B2: mode_decision_sb (EbProductCodingLoop.c:8433)
   by 0x4FD3D3F: enc_dec_kernel (EbEncDecProcess.c:3104)
   by 0x5EA5DD4: start_thread (in /usr/lib64/libpthread-2.17.so)
   by 0x64B9EAC: clone (in /usr/lib64/libc-2.17.so)
```

PR #1183, opened 2020-04-07
Fix UT, read outside of allocated memory

tszumski:
UT is failing for example here: https://github.com/OpenVisualCloud/SVT-AV1/pull/1154/checks?check_run_id=566794496

I can reproduce it on windows, visual studio 2019 Debug build with error:
```
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
```
Root cause is that test allocate too small input buffer


hassount:
@andrey-norkin @kylophone @mafonsoN @joelsole any issues with this ?

PR #1188, opened 2020-04-09
CI: Add simple cppcheck job

### LINKED ISSUE #1161: Github CI: Static analyzer build
Make a build with a static analyzer and upload the results somewhere. Possible choices would be the clang static analyzer or cppcheck (never used that one before though).

ePirat:
This adds static analysis using [cppcheck](http://cppcheck.sourceforge.net) and creates annotations that show up in the job information. Additionally if a PR introduces new errors, those should show up inline in the diff.

Fix #1161

### DIFF
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -1,0 +1,26 @@
+name: Cppcheck
+on: [push, pull_request]
+
+jobs:
+  check:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y nasm cppcheck
+      - name: Run CMake
+        run: |
+          cd Build
+          cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
+      - name: Run Cppcheck
+        run: |
+          cd Build
+          cppcheck \
+            --project=compile_commands.json \
+            --error-exitcode=0 \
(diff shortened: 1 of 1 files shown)


1480c1:
Nice.

It might be better to enable failure for now until all of the issues from cppcheck are addressed

ePirat:
@intelmark Hi, thanks for the feedback. Yes this is an option I checked too but it uses among other things the clang static analyzer which can do very complex analysis, which is great but has the side-effect that sometimes its not really easy to understand how it came to a specific conclusion. Additionally it produces a lot of reports, and to manage those properly and actually be able to see the detailed report with the full trace that explains how it concludes specific issues, it would probably be better to use a tool like codechecker, as I mentioned in #1161 already.

That said, we could add it and see how it goes, although maybe probably fixing the cppcheck warnings first before adding even more is better?

intelmark:
@ePirat - The cppcheck is solid starting point and we can upgrade it later once it stops reports anything useful.  Curious to see the other reviewer's comments.

.github/workflows/cppcheck.yml
```diff
@@ -0,0 +1,25 @@
...
+          sudo apt-get install -y nasm cppcheck
+      - name: Run CMake
+        run: |
+          cd Build
+          cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
+      - name: Run Cppcheck
+        run: |
+          cd Build
+          cppcheck \
+            --project=compile_commands.json \
```
1480c1, line 22: I recommend using `--enable=all` to print some other helpful messages that's also helpful for style too

PR #1191, opened 2020-04-09
fix encoding for non multiple-of-8 pictures

kylophone:
Cherry-picked and rebased from the `cs2` branch. 

Co-authored-by: Jing Li <jing.b.li@intel.com>
Co-authored-by: chkngit <42250551+chkngit@users.noreply.github.com>

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -2787,10 +2787,8 @@ EbErrorType read_command_line(int32_t argc, char *const argv[], EbConfig **confi
 
                 // Assuming no errors, add padding to width and height
                 if (return_errors[index] == EB_ErrorNone) {
-                    configs[index]->input_padded_width =
-                        configs[index]->source_width + configs[index]->source_width % 8;
-                    configs[index]->input_padded_height =
-                        configs[index]->source_height + configs[index]->source_width % 8;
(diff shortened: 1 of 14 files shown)


joelsole:
I did a couple of fast encodes and compared the SVT-AV1 reconstructed, SVT-AV1 decoded, and libaom decoded yuvs:
 - For 1 frame, the three outputs match
 - For 2 frames, SVT-AV1 recon & decoded match, but not libaom decoded
 - For 12 frames: the three outputs are different. 
This happened with the first two sequences picked, so I guess it is not a corner case.

Do you observe the same at your end?

hassount:
> I did a couple of fast encodes and compared the SVT-AV1 reconstructed, SVT-AV1 decoded, and libaom decoded yuvs:
> 
> * For 1 frame, the three outputs match
> * For 2 frames, SVT-AV1 recon & decoded match, but not libaom decoded
> * For 12 frames: the three outputs are different.
>   This happened with the first two sequences picked, so I guess it is not a corner case.
> 
> Do you observe the same at your end?

I did an encode and the reconstructed vs the decoded (through ffmpeg / libaom decoder) have matched fully for 60 frames

joelsole:
We can note that this PR is fixing the 8-bit case, merge it, and make an issue for fixing the 10-bit case.

hassount:
Added #1225

PR #1193, opened 2020-04-10
CMake: Do not use GLOB for source files

ePirat:
Using a GLOB to collect lists of source files is not recommended,
instead explicitly list source files.

Especially in a project with lots of contributors, using a GLOB can cause huge headache for other contributors, as someone might add a file or change a existing files name without having to change any `CMakeLists.txt`, while this seems convenient at first, when now other contributors pull this change, there is no `CMakeLists.txt` change, so CMake will not evaluate the GLOB again, the build will be done with the new file missing or the renamed file not found and will cause confusing linker or compiler errors.

So while it is now a little bit more effort for the individual contributor (they need to change or add the filename in `CMakeLists.txt`), it's less work for everyone else as there is no need to always re-run CMake now.


Source/Lib/Encoder/ASM_SSE4_1/CMakeLists.txt
```diff
@@ -23,9 +23,15 @@ endif()
...
+set(all_files
+    av1_txfm1d_sse4.h
+    av1_txfm_sse4.h
+    EbComputeSAD_Intrinsic_SSE4_1.c
+    EbTemporalFiltering_constants.h
+    EbTemporalFiltering_highbd_sse4.c
+    EbTemporalFiltering_sse4.c
+    highbd_fwd_txfm_sse4.c
+    obmc_variance_sse4.c
+)
```
1480c1, line 35: Please make sure that your changes stick to one style, it would be easier to fix style issues at the beginning instead of creating style commits later one

ePirat: Whoops, missed that one! Thanks, fixed.

hassount:
@ePirat thank you for the PR. 
at this point and as we are still adding features / cleaning up and moving code around. I believe it would be best to keep it as is. Once items are stable we can consider this part later.

ePirat:
Well especially at this point when changing frequently the reason why you should not use GLOB applies even more...

ePirat:
What are the plans for sources btw? Will they be renamed to a more consistent naming? Currently its a mix of CamelCase with the files having an `Eb` prefix and non-camel-cased ones…

ePirat:
@hassount If there are plans to rename, move or add lots of files though I can see why we might want to wait with this PR and just rebase later. It's not like this is anything time-sensitive to merge.

hassount:
@ePirat yes, file nomenclature might change and be unified and files names might change, so let's wait on this until we stabilize on that aspect.

ePirat:
@hassount Ok, thanks for the update.

One issue I am facing without this is that doing conditional compilation (compile one file for Windows, another for other OSes) is currently tricky to do.

andrey-norkin:
In general, I agree with @ePirat here. Having the list of files updated is useful during the development. We should merge this PR after we do a round of remaining the files.

hassount:
I invite others to chime in here, that said, I believe we agree that we should push this down further. can we close this until then ?

ePirat:
Let's leave it open so we do not forget? I can mark it was WIP.

PR #1195, opened 2020-04-10
build.sh: fix an issue in arg splitting

1480c1:
Arguments weren't shifted properly for single dash options that used an argument

### DIFF
--- a/Build/linux/build.sh
+++ b/Build/linux/build.sh
@@ -307,16 +307,52 @@ else
             while [ $i -ne $((${#match} + 1)) ]; do
                 case "$(echo "$match" | cut -c$i | tr '[:upper:]' '[:lower:]')" in
                 h) parse_options help ;;
-                a) parse_options all && i=$((i + 1)) ;;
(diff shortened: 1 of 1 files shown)


Build/linux/build.sh
```diff
@@ -307,16 +307,53 @@ else
...
-                j) parse_options jobs="$1" && i=$((i + 1)) ;;
-                p) parse_options prefix="$1" && i=$((i + 1)) ;;
-                s) parse_options target_system="$1" && i=$((i + 1)) ;;
-                t) parse_options toolchain="$1" && i=$((i + 1)) ;;
-                x) parse_options build_static && i=$((i + 1)) ;;
-                v) parse_options verbose && i=$((i + 1)) ;;
+                a)
+                    parse_options all
+                    i=$((i + 1))
+                    shift
```
joelsole, line 313: is this shift intended?

1480c1: Nope

PR #1200, opened 2020-04-14
Write build output by default to build dir

ePirat:
Writing the build output to the source directory by default
is very confusing when one has multiple build directories configured
in different ways (like one with sanitizers enabled, one without) and
really defeats the benefits of building outside the source directory.

Note: I've change the build.sh to retain the old build results path, so people that use the script will get the old behavior. I did **not** update the build.bat because this one is really complicated to understand and I am a bit clueless what there even the old behavior was, as it used the `CMAKE_BUILD_TYPE` but on Windows it seems one can has two build types that are only actually defined when building? But the `CMAKE_ARCHIVE_OUTPUT_DIRECTORY` is evaluated on initial configure as far as I can see, so I wonder what it would set there then…


ePirat:
So it turns out when run on Windows with the bat file, the output depends on the generator used apparently, which is a bit weird…

1480c1:
When you say `build output`, are you talking about the generated files etc (Makefile etc), or are you talking about what get's created through `make` or `ninja`?

ePirat:
@1480c1 About what is created through make / ninja, specifically in this case the final libraries and binaries, which currently end up in `SVT-AV1/Build/linux/{debug,release}`. (Or similar location for Windows, although for Windows it's a bit weird, as the build type is not know to CMake at configure time)

ePirat:
So basically my goal is that the binaries and libraries end up at the expected location, in the build directory, not the source directory.
Having them in the source directory leads to extremely annoying situation that it's not possible to have two build dir of the same build type with different other settings (different compilers for example, or different sanitizer flags) as they would always write the binaries and libraries to the exact same location, overwriting each other.

CMakeLists.txt
```diff
@@ -34,9 +34,10 @@ include(GNUInstallDirs)
 include(CheckCCompilerFlag)
 include(CheckCXXCompilerFlag)
 
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/Bin/${CMAKE_BUILD_TYPE}/)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/Bin/${CMAKE_BUILD_TYPE}/)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/Bin/${CMAKE_BUILD_TYPE}/)
+set(BUILD_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/Bin/" CACHE PATH "Build results output directory")
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${BUILD_OUTPUT_DIRECTORY})
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${BUILD_OUTPUT_DIRECTORY})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${BUILD_OUTPUT_DIRECTORY})
```
1480c1, line 40: One thing you can do instead then is to use `CMAKE_BINARY_DIR` to make something like
```suggestion
set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/Bin")
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/Bin")
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/Bin")
```
Since I'm not sure if adding another variable just for setting where you want the build output files to go to

ePirat: The reason I did this was to be able to keep the old paths for the scripts. Alternatively I could just adapt the CI and documentation to the new path, which is probably a better long-term solution.

1480c1: > adapt the CI and documentation to the new path, which is probably a better long-term solution.

I personally think this would be much better overall since writing to the "source" dir negates and usefulness of out of source builds

ePirat: I absolutely agree, I will adapt the patchset.

ePirat: Is there any difference between `PROJECT_BINARY_DIR` and `CMAKE_BINARY_DIR` btw?

1480c1: I guess the difference would come from if you make SVT-AV1 as a subproject. So `PROJECT_BINARY_DIR` will be set by the most recent `project(name)` cmakelist command while `CMAKE_BINARY_DIR` will by set by the most recent `cmake` command

ePirat: Ok. I do not have much experience with CMake, should we prefer `PROJECT_BINARY_DIR` in order to not put files in the parent project build dir when it is used as subproject or just use `CMAKE_BINARY_DIR`?

1480c1: I guess it would be better to go ahead and just use `PROJECT_BINARY_DIR`, there isn't any harm in choosing it

ePirat:
This is ready now, I believe. I've replaced all paths with the new ones.

The binary identical tests fail currently, as this change has not landed in master yet and it tests against the master version.

CMakeLists.txt
ePirat: Done the requested change.

ePirat:
Ok, appveyor was failing, should hopefully be fixed now…

1480c1:
@hassount, since we have a msvc build job in github actions, is it still necessary to keep the msvc job in appveyor?

ePirat:
@1480c1 It seems appveyor currently creates the release artifacts.

I can move those over to Github Actions soon too and then I guess it's redundant to have appveyor.

1480c1:
> @1480c1 It seems appveyor currently creates the release artifacts.
> 
> I can move those over to Github Actions soon too and then I guess it's redundant to have appveyor.

Totally forgot that we are using it for the artifacts.

I personally wish we could move them to github actions, but it seems that github themselves do not provide an upload release action or similar

ePirat:
@1480c1 I believe they do provide one, I saw one a while ago when I looked into automating releases for a different project.

I first would like to do some cleanup of the GH Actions stuff though as things are getting a bit messy and harder to maintain.

1480c1:
> @1480c1 I believe they do provide one, I saw one a while ago when I looked into automating releases for a different project.
> 
> I first would like to do some cleanup of the GH Actions stuff though as things are getting a bit messy and harder to maintain.
> 
> EDIT: For future reference: https://github.com/actions/upload-release-asset

Nice, I did not find this previously when I was looking for it. I guess I looked too early

ePirat:
As per discussion with @hassount on the Meeting today, I'll change this back to how it was before to not change how the scripts behave but only have the "new" behavior when running CMake manually without specifying output path.

1480c1:
@ePirat will you be working on cleaning up github actions then adding upload-release-asset?

ePirat:
@1480c1 Yes, I will probably do that tomorrow.

ePirat:
> @1480c1 Yes, I will probably do that tomorrow.

Just follow up on this, the approaches to cleaning up did not work due to limitations of Github Actions (lack of YAML anchors, or includes, extends, etc. directives).

CMakeLists.txt
```diff
@@ -34,9 +34,9 @@ include(GNUInstallDirs)
 include(CheckCCompilerFlag)
 include(CheckCXXCompilerFlag)
 
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/Bin/${CMAKE_BUILD_TYPE}/)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/Bin/${CMAKE_BUILD_TYPE}/)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR}/Bin/${CMAKE_BUILD_TYPE}/)
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/Bin)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/Bin)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/Bin)
```
1480c1, line 39: Should we set these as cache variables for now instead? That way we can manually set the output directory when we configure a build.

ePirat: Indeed that might at least allow it to easily override locally? I dont fully understand how CMake cached variables work…

1480c1: It will allow the user to set `-DCMAKE_ARCHIVE_OUTPUT_DIRECTORY=$PWD` when running cmake and not override it

ePirat: That sounds like a good idea

PR #1202, opened 2020-04-15
Remove EB_NULL

ePirat:
There is no good reason to have an own definition that is essentially
the same as `NULL`, it just makes the code more confusing to read.

### DIFF
--- a/Source/App/EncApp/EbAppConfig.h
+++ b/Source/App/EncApp/EbAppConfig.h
@@ -35,10 +35,6 @@ API.  This is a 32 bit pointer and is aligned on a 32 bit word boundary.
 */
 typedef void *EbPtr;
 
-/** The EB_NULL type is used to define the C style NULL pointer.
-*/
-#define EB_NULL ((void *)0)
-
 #define WARNING_LENGTH 100
 
 // memory map to be removed and replaced by malloc / free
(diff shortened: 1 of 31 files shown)


1480c1:
`The remote server returned an error: (503) Server Unavailable. Service Unavailable`
It seems `choco install nasm` doesn't properly return an error. Maybe it might be a good idea to add `nasm --version` around the same time we call `cmake --version` to make sure it will fail in the dependency step if nasm failed to install

ePirat:
Well it ends up with the same result anyway of the test failing… Maybe choco does return an error but it does not show because fo the cmake --version afterwards which returns successfully? Just a guess, no idea how powershell interprets these things…

1480c1:
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

`Fail-fast behavior when possible. For pwsh and powershell built-in shell, we will prepend $ErrorActionPreference = 'stop' to script contents.`

Theoretically, this should stop the execution once any command that exits with a non-zero code, although tbh, I haven't had a good experience with it since I mainly dealt with *nix tools that also somehow tripped it with just printing to stderr, but idk if that still applies anymore

joelsole:
@hassount , any reason on your side to keep EB_NULL?

hassount:
> @hassount , any reason on your side to keep EB_NULL?

I believe this is part of a legacy code, @chkngit any issues with EB_NULL gone ?

EricLamVCD:
Results are as expected: 
  | Speed Deviation | BD-Rate   Deviation
-- | -- | --
Reported | - | 0.00%
Tested | -0.11% | 0.00%

PR #1203, opened 2020-04-15
Documentation fix: keyint -> intra-period

sandersaares:
The documentation says --keyint but in reality it seems that the name of the parameter is --intra-period (at least in the latest available release)


hassount:
@sandersaares  thank you for the contribution
it is actually --keyint which has changed recently from -intra-period, the encoder does still accept the -intra-period option as a transition but it will be deprecated soon and it shows a warning for such. the --intra-period option does not work

sandersaares:
Oh okay, so the documentation is just ahead of the latest release? Got it - I will close this PR in that case.

silverbacknet:
FWIW, the manual for any particular release can be browsed by tag, as in https://github.com/sandersaares/SVT-AV1/blob/v0.8.1/Docs/svt-av1_encoder_user_guide.md

PR #1209, opened 2020-04-15
EncApp/EbAppContext.c: Remove call for allocate_output_buffers

1480c1:
Since nothing was using the `stream_buffer_pool`, it would be better to remove it for now, plus anyone needing it can grep it from the git log.

There seem to be a lot of variables that are generally unused it seems.

### DIFF
--- a/Source/App/EncApp/EbAppContext.c
+++ b/Source/App/EncApp/EbAppContext.c
@@ -382,32 +382,6 @@ EbErrorType allocate_output_recon_buffers(EbConfig *config, EbAppContext *callba
     return EB_ErrorNone;
 }
 
-EbErrorType allocate_output_buffers(EbConfig *config, EbAppContext *callback_data) {
-    uint32_t output_stream_buffer_size = (uint32_t)(
-        EB_OUTPUTSTREAMBUFFERSIZE_MACRO(config->input_padded_height * config->input_padded_width));
-
-    EB_APP_MALLOC(EbBufferHeaderType *,
-                  callback_data->stream_buffer_pool,
-                  sizeof(EbBufferHeaderType),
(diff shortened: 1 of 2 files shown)


1480c1:
```none
Maximum allowed artifact storage size of 100000 Mb will be exceeded.
APPVEYOR_SAVE_CACHE_ON_ERROR is set to 'true' - cache will be updated
```
I need to find a way to make it not upload on pull requests

hassount:
> ```
> Maximum allowed artifact storage size of 100000 Mb will be exceeded.
> APPVEYOR_SAVE_CACHE_ON_ERROR is set to 'true' - cache will be updated
> ```
> 
> 
> I need to find a way to make it not upload on pull requests

does github actions upload binaries ? the only good thing right now about appveyor is that we get the ffmpeg build ready to get

1480c1:
> > ```
> > Maximum allowed artifact storage size of 100000 Mb will be exceeded.
> > APPVEYOR_SAVE_CACHE_ON_ERROR is set to 'true' - cache will be updated
> > ```
> > 
> > 
> > I need to find a way to make it not upload on pull requests
> 
> does github actions upload binaries ? the only good thing right now about appveyor is that we get the ffmpeg build ready to get

https://github.com/OpenVisualCloud/SVT-AV1/pull/1200#issuecomment-613810093 It seems so

mazharaIntel:
Tested with 2 clips from each resolutions. 6 - 8bit and 6 10bit clips:

  | Speed Deviation | BD-Rate Deviation
-- | -- | --
M0 (8bit) | 0.07% | 0.00% | 
M0 (10bit) | -1.09% | 0.00% |

PR #1211, opened 2020-04-16
Fix frames_to_be_encoded for non-file input

### LINKED ISSUE #1205: stdin for input only works if the input is a file, not when streaming from FFmpeg
On Windows 10, given this input file: 
[demo.zip](https://github.com/OpenVisualCloud/SVT-AV1/files/4479702/demo.zip)

Using ffmpeg version git-2020-03-15-c467328 and SVT-AV1 [0.8.1] - 2020-01-28

This sequence of commands works fine:

```
ffmpeg -i demo.mp4 -f yuv4mpegpipe demo.y4m
SvtAv1EncApp.exe -i stdin -w 1920 -h 800 -fps 24 < demo.y4m
```

This does not work:

```
ffmpeg -i demo.mp4 -f yuv4mpegpipe - | SvtAv1EncApp.exe -i stdin -w 1920 -h 800 -fps 24
```

In case 1 I write from FFmpeg to file and then pipe that file to SVT-AV1.

In case 2 I pipe directly from FFmpeg to SVT-AV1, without a file in between.

The latter appears to trigger some fault condition where SVT-AV1 thinks the input is bad. Maybe it tries to read input before FFmpeg has produced any? Just speculating.

The output printed in the second case contains the following relevant lines:

```
Error instance 1: Input yuv does not contain enough frames
Error in configuration, could not begin encoding! ...
Run SvtAv1EncApp.exe -help for a list of options
Encoder finished
```

FFmpeg just sees that pipe was closed:

```
av_interleaved_write_frame(): Broken pipe
Error writing trailer of pipe:: Broken pipe
```

sandersaares:
Fixes #1205

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -2782,7 +2782,6 @@ EbErrorType read_command_line(int32_t argc, char *const argv[], EbConfig **confi
     /***************************************************************************************************/
     /**************************************   Verify configuration parameters   ************************/
     /***************************************************************************************************/
-
     // Verify the config values
     if (return_error == 0) {
         return_error = EB_ErrorBadParameter;
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -2230,6 +2230,10 @@ int32_t compute_frames_to_be_encoded(EbConfig *config) {
     uint32_t frame_size;
     uint64_t curr_loc;
 
+    // Pipes contain data streams whose end we cannot know before we reach it.
+    // For pipes, we leave it up to the eof logic to detect how many frames to eventually encode.
+    if (config->input_file == stdin || config->input_file_is_fifo) return -1;
+
     if (config->input_file) {
         curr_loc = ftello(config->input_file); // get current fp location
         fseeko(config->input_file, 0L, SEEK_END);
@@ -2778,6 +2782,7 @@ EbErrorType read_command_line(int32_t argc, char *const argv[], EbConfig **confi
     /***************************************************************************************************/
     /**************************************   Verify configuration parameters   ************************/
     /***************************************************************************************************/
+
     // Verify the config values
     if (return_error == 0) {
         return_error = EB_ErrorBadParameter;
@@ -2799,7 +2804,10 @@ EbErrorType read_command_line(int32_t argc, char *const argv[], EbConfig **confi
                     configs[index]->frames_to_be_encoded =
                         compute_frames_to_be_encoded(configs[index]);
 
-                if (configs[index]->frames_to_be_encoded == -1) {
+                // For pipe input it is fine if we have -1 here (we will update on end of stream)
+                if (configs[index]->frames_to_be_encoded == -1
+                    && configs[index]->input_file != stdin
+                    && !configs[index]->input_file_is_fifo) {
                     fprintf(configs[index]->error_log_file,
                             "Error instance %u: Input yuv does not contain enough frames \n",
                             index + 1);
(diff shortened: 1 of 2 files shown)


silverbacknet:
I feel like it would be significantly better to have a separate sentinel value to mean "indefinite" (-2?) vs "error" (-1), and that there's no reason to modify compute_frames_to_be_encoded() when you could just skip it entirely by testing early.

sandersaares:
> I feel like it would be significantly better to have a separate sentinel value to mean "indefinite" (-2?) vs "error" (-1),

There is no more "-1 means error" interpretation necessary, -1 can be taken as "unknown" everywhere. There remains a check along the lines of "must not be unknown if it is a file", of course, where unknown leads to an error but that seems to follow logically enough. So there is only a single interpretation necessary.

>  there's no reason to modify compute_frames_to_be_encoded() when you could just skip it entirely by testing early.

I think then it would no longer be -1 for unknown but have the default value of 0. This is the function that sets it to the -1 value. Computing the number of frames returns the value "don't know" - code seems to match semantics quite well.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -2799,7 +2804,10 @@ EbErrorType read_command_line(int32_t argc, char *const argv[], EbConfig **confi
                     configs[index]->frames_to_be_encoded =
                         compute_frames_to_be_encoded(configs[index]);
```
tianjunwork, line 2805: On linux `compute_frames_to_be_encoded` `frame_count` is a huge minus value, while on Windows it is 0. 
So it won't run into if at line 2808 on linux.
I agree with silverbacknet, just skip compute_frames_to_be_encoded at line 2805 for stdin or fifo. And keep it as 0.

tianjunwork: Ping @sandersaares

sandersaares: > I agree with silverbacknet, just skip compute_frames_to_be_encoded at line 2805 for stdin or fifo. And keep frames_to_be_encoded as 0.

0 is also a valid value for the case "File is too small to even fit one frame of data", so this would lead to re-using the same value for two purposes (a real count of zero and also "unknown"). I am not comfortable with this type of reuse of a valid value to designate "unknown".

Furthermore `process_input_buffer()` has some logic that checks for frame count and expects `< 0` to mean unknown. I am not confident that I understand the logic of that function enough to determine the appropriate adjustment for line 955, though.

If you insist on this change, to treat 0 as "unknown", I ask that someone other than me do it so that the logic could be adjusted without introducing errors.

tianjunwork: @sandersaares Thank you for the reply. I see your point. 
And Thanks adding the comments in the code. Basically `-1` is treated as `unknown` which is ok for non-file input. I agree with the fix.

tianjunwork:
With this PR, 'ffmpeg -i demo.mp4 -f yuv4mpegpipe - | SvtAv1EncApp.exe -i stdin -w 1920 -h 800 -fps 24' works on Windows now.
@sandersaares btw, for y4m stream, `-w -h -fps` are not needed, will be over written by `read_y4m_header`.

tianjunwork:
Hi @sandersaares, can I use title `Fix frames_to_be_encoded for non-file input`?

sandersaares:
Sounds good to me

mazharaIntel:
Tested with 2 clips from each resolutions. 6 - 8bit and 6 10bit clips:

  | Speed Deviation | BD-Rate Deviation 
-- | -- | -- 
M0   (8bit) | 0.78% |  0.00% |  
M0   (10bit) | -0.25% | 0.00% |

PR #1212, opened 2020-04-16
Support stdout for output

### LINKED ISSUE #701: Support output to stdout for improved pipe support {feature request}

sandersaares:
Closes #701

### DIFF
--- a/Docs/svt-av1_encoder_user_guide.md
+++ b/Docs/svt-av1_encoder_user_guide.md
@@ -122,8 +122,8 @@ The encoder parameters present in the `Sample.cfg` file are listed in this table
 #### Options
 | **Configuration file parameter** | **Command line** | **Range** | **Default** | **Description** |
 | --- | --- | --- | --- | --- |
-| **InputFile** | -i | any string | None | Input file path |
-| **StreamFile** | -b | any string | null | output bitstream file path |
+| **InputFile** | -i | any string | None | Input file path; "stdin" to take input from standard input stream |
+| **StreamFile** | -b | any string | null | Output bitstream file path; "stdout" to emit output to standard output stream |
 | **ConfigFile** | -c | any string | null | Configuration file path |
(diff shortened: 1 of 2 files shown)


silverbacknet:
The usual convention for stdout/stdin is a single dash "-", not "stdout" though. Linux has /dev/stdout to explicitly name it, although Windows doesn't natively have an equivalent.

1480c1:
I think this is to match the `-i stdin` we are currently using

sandersaares:
Yeah, exactly. Oh, that reminds me I should also update docs to mention this.

tianjunwork:
Tested on both Linux and Windows with:
 `SvtAv1EncApp -i ../../yuv/akiyo_cif.y4m -b stdout | ffmpeg -i - -c:v copy out.mkv`
It works on both with newer ffmpeg version. But it has nothing to do with the implementation of this PR. Just to share this info in case anyone runs into issue.

PR #1220, opened 2020-04-17
Print verbose version info in encoder and decoder

### LINKED ISSUE #1208: Add git hash to the build version
Right now any build would have the same version that is defined statically every tag. to fit better with snap as in #38 we would need to add the git hash to the revision number similar to other encoders such as libaom e.g. g66603e7ee

ePirat:
Fix #1208

This is an initial step, it adds the build version without affecting much else. If not git is found or the  source is not a git checkout, it falls back to the project version.


hassount:
@ePirat 
thank you very much for the PR. 
when I run the encoder on the branch, I get the following output:
```
SVT-AV1 Encoder
SVT [version]:  SVT-AV1 Encoder Lib v0.7.5-297-gb4cc1c86
SVT [build]  :  Visual Studio 2017 64 bit
LIB Build date: Apr 17 2020 13:57:06
```

Is there a reason why it doesn't use 0.8.1 as an initial tag version ?

ePirat:
@hassount I can't reproduce that, for me it has the expected value there. Are you sure you actually have fetched all the tags?

Can you run:

```
git tag -l
```

and

```
git describe --tags
```

and post the output here?

hassount:
```
$ git tag -l
v0.5.0
v0.6.0
v0.7.0
v0.7.5

$ git describe --tags
v0.7.5-297-gb4cc1c86

```

seems like your master is not up-to-date ?

another thing that's weird is that I don't see the printout in the ci:

https://github.com/OpenVisualCloud/SVT-AV1/pull/1220/checks?check_run_id=595839640

I don't see the printout:
```
SVT-AV1 Encoder
SVT [version]:	SVT-AV1 Encoder Lib v0.8.1
SVT [build]  :	GCC 9.2.1 20191102	 64 bit
LIB Build date: Apr 17 2020 15:41:02
```

this might be more of a ci problem :/

ePirat:
The CI does a limited depth clone I believe, so it lacks the necessary info. Which is why it falls back to the version info from the project.

ePirat:
@hassount So the output is expected for you given you lack more recent tags. You should fetch them with `git pull --tags`

ePirat:
I've just checked and it stated in the [README](https://github.com/actions/checkout#whats-new) for the checkout action:

> Fetches only a single commit by default

So if we want CI to have the `git describe` output we need to adjust that.

hassount:
> @hassount So the output is expected for you given you lack more recent tags. You should fetch them with `git pull --tags`

I did that and it says everything is uptodate. The main repo master shows all tags, the fork doesn't when I go to the fork, I don't see tags after 0.7.5 
https://github.com/ePirat/SVT-AV1/releases

anyways it should work once it's merged

ePirat:
Oh indeed, I've pushed the tags now, sorry!

CMakeLists.txt
```diff
@@ -19,7 +19,8 @@ if(YASM)
     endif()
 endif()
 
-project(svt-av1 C CXX ASM_NASM)
+project(svt-av1 VERSION 0.8.1
+    LANGUAGES C CXX ASM_NASM)
```
tianjunwork, line 23: Is this change needed since version only depends on git tag?

1480c1: This is needed in the case a person builds from a tarball and doesn't have `.git`

Source/Lib/Common/Codec/CMakeLists.txt
```diff
@@ -9,6 +9,17 @@ else ()
     set (X86 FALSE)
 endif ()
 
+add_custom_target(EbVersionHeaderGen
+    ${CMAKE_COMMAND}
+        -D INPUT_FILE=${CMAKE_CURRENT_SOURCE_DIR}/EbVersion.h.in
+        -D OUTPUT_FILE=EbVersion.h
```
tianjunwork, line 15: Minor feedback:
maybe put EbVersion.h in the same folder of EbVersion.h.in. then include_directories doesn't need to change.
EbVersion.h can be added to .gitignore.

ePirat: This would put it in the source dir, which I think is not a good idea, as all build artifacts should be in the build folder(s).

CMakeLists.txt
ePirat: Yeah or in the case that the clone has no history `--depth 1`, or similar reasons where `git describe` would fail. Then it will fall back to this version.

The `LANGUAGES` change is needed because CMake requires it, when a `VERSION` is given.

hassount:
@ePirat 

are these definitions still needed ? 

https://github.com/ePirat/SVT-AV1/blob/b4cc1c868af87c8b5ff5d86d34047fe0f8c8317a/Source/API/EbSvtAv1.h#L17

also can we link the new rev to snap/snapcraft.yaml ?

ePirat:
@hassount Yes those are still needed, it's in the installed headers and used for public API to check at compile time which version of SVT-AV1 is used. I can submit subsequent PRs that consolidate this a bit more and provide an API to get the runtime library version too (as that does not necessarily matches the version it was originally compiled against).

> also can we link the new rev to snap/snapcraft.yaml ?

I'll check that, I am not sure as I am really not familiar at all with it.

hassount:
> PRs that consolidate this a bit more and provide an API to get the runtime library version too (as that does not necessarily matches the version it was originally compiled against).

sounds good, we can improve this, I just wanted to have this as part of the tag, we can add more items later.

ePirat:
@hassount I've done the snap version change in #1226

tianjunwork:
Hi @ePirat, change in Hevc. https://github.com/OpenVisualCloud/SVT-HEVC/pull/310/files. Only need to update CMakeLists.txt for a new tag. Consolidating of the version is a good idea.

ePirat:
@tianjunwork Thanks for the link! I'll do something similar for SVT-AV1 and submit a PR with it soon.

hassount:
@ePirat 
building with ./build.sh causes a build error after this PR:
```
Source/Lib/Encoder/Globals/EbEncHandle.c:16:10: fatal error: EbVersion.h: No such file or directory
 #include "EbVersion.h"
```

silverbacknet:
Related to #1193? I had to clear CMake cache to build.

ePirat:
@hassount make sure you cleared the build folder before. CMakes caching is weird.
Given that CI succeeds and it uses build.sh it definitely should work

hassount:
> @hassount make sure you cleared the build folder before. CMakes caching is weird.
> Given that CI succeeds and it uses build.sh it definitely should work

downloaded a fresh copy and I still see the issue. 
the windows build is now failing for the same reason: https://ci.appveyor.com/project/tianjunwork/svt-av1/builds/32266674/job/o3vn7cjiw377orh2

could you check ? 

I am trying to tag by tomorrow, so if you could please let me know if we should revert this for now and put it back later.

ePirat:
@hassount Sorry, when reworking the file location I got the dependency wrong. This should be fixed with #1229.

hassount:
> @hassount Sorry, when reworking the file location I got the dependency wrong. This should be fixed with #1229.

thnx @ePirat, tested locally and it went good, let's check that appveyor would pass so that we can merge

tianjunwork:
> @tianjunwork Thanks for the link! I'll do something similar for SVT-AV1 and submit a PR with it soon.

Hi @ePirat , just checking, could we have it before 0.8.4 release?

ePirat:
@tianjunwork Consolidating the versions, you mean?

tianjunwork:
@ePirat yes:)

ePirat:
@1480c1 Do you have any work on that? Else I guess I can look into that this week probably.

1480c1:
None really atm

PR #1224, opened 2020-04-17
Fix calling NULL pointer

### LINKED ISSUE #1216: Encoder segmentation fault after PR1155
On mac. Stack trace below. It does NOT happen without option "--pred-struct-file ../../Config/PredStruct_level4.cfg". 

run  -i DayStreet_12fr_256x256_60fps_10bits_420.yuv -w 256 -h 256 -q 50 --fps-num 60 --fps-denom 1 -b DayStreet_12fr_256x256_60fps_10bits_420_eo0_pr3_qp50.ivf --keyint 31 --input-depth 10 --pred-struct-file ../../Config/PredStruct_level4.cfg -o recon_DayStreet_12fr_256x256_60fps_10bits_420_eo0_pr3_qp50.yuv --preset 3 --enable-stat-report 1 --lp 1 -n 12 --scd 0
Process 80115 launched: '/Users/jsole/Documents/code/SVTAV1/ovc_SVT-AV1/Bin/Debug/SvtAv1EncApp' (x86_64)

SVT-AV1 Encoder
SVT [version]:  SVT-AV1 Encoder Lib v0.8.1
SVT [build]  :  GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.16)      64 bit
LIB Build date: Apr 16 2020 15:01:57
Process 80115 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000000000000
error: memory read failed for 0x0
Target 0: (SvtAv1EncApp) stopped.

* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x000000010055cb01 libSvtAv1Enc.0.dylib`**prediction_structure_group_ctor**(pred_struct_group_ptr=0x000000010df69910, enc_mode='\x03', config=0x000000010d008010) at EbPredictionStructure.c:2153:9
    frame #2: 0x000000010042d3ff libSvtAv1Enc.0.dylib`svt_av1_enc_set_parameter(svt_enc_component=0x000000010b600090, config_struct=0x000000010c000a00) at EbEncHandle.c:3181:5
    frame #3: 0x000000010000ac54 SvtAv1EncApp`init_encoder(config=0x000000010c000000, callback_data=0x000000010c000a00, instance_idx=0) at EbAppContext.c:485:20
    frame #4: 0x000000010000c52c SvtAv1EncApp`main(argc=33, argv=0x00007ffeefbff180) at EbAppMain.c:162:25
    frame #5: 0x00007fff707173d5 libdyld.dylib`start + 1

ePirat:
The eb_memcpy is a function pointer to NULL which is only initialized
after a call to setup_common_rtcd_internal, which is not done yet
when this is called.

Fix #1216

This issues might be present in other cases, this code is a bit messy and the way those pointer are initialized being done in `setup_common_rtcd_internal` is really weird…
Even more confusing that there is `EB_MEMCPY` which does not do the same as `eb_memcpy` with regard to which implementation is called…
(It was introduced in #1155 where some `EB_MEMCPY` were replaced with `eb_memcpy`)

### DIFF
--- a/Source/Lib/Encoder/Codec/EbPredictionStructure.c
+++ b/Source/Lib/Encoder/Codec/EbPredictionStructure.c
@@ -2150,11 +2150,11 @@ EbErrorType prediction_structure_group_ctor(PredictionStructureGroup *pred_struc
     // Insert manual prediction structure into array
     if (config->enable_manual_pred_struct) {
         prediction_structure_config_array[config->hierarchical_levels].entry_count = config->manual_pred_struct_entry_num;
-        eb_memcpy(prediction_structure_config_array[config->hierarchical_levels].entry_array,
+        EB_MEMCPY(prediction_structure_config_array[config->hierarchical_levels].entry_array,
           &config->pred_struct[config->manual_pred_struct_entry_num - 1],
           sizeof(PredictionStructureConfigEntry));
         if (config->manual_pred_struct_entry_num > 1) {
-            eb_memcpy(prediction_structure_config_array[config->hierarchical_levels].entry_array + 1,
+            EB_MEMCPY(prediction_structure_config_array[config->hierarchical_levels].entry_array + 1,
               &config->pred_struct[0],
               (config->manual_pred_struct_entry_num - 1) * sizeof(PredictionStructureConfigEntry));
         }


ttrigui:
> The eb_memcpy is a function pointer to NULL which is only initialized
> after a call to setup_common_rtcd_internal, which is not done yet
> when this is called. So use a normal memcpy here.
> 
> Fix #1216
> 
> This issues might be present in other cases, this code is a bit messy and the way those pointer are initialized being done in `setup_common_rtcd_internal` is really weird…
> Even more confusing that there is `EB_MEMCPY` which does not do the same as `eb_memcpy` with regard to which implementation is called…
> (It was introduced in #1155 where some `EB_MEMCPY` were replaced with `eb_memcpy`)

eb_memcpy was defined in ebDefinitions.h so regardless of the optimization level (ie. asm 0,1, 2,..) we always call the optimized version. In order to assign either c or optimized code the function assignment was moved to setup_common_rtcd_internal where all the optimized functions are being assigned.
The problem here is that this is being used in the application and the manual prediction case ( which was missed) before calling setup_common_rtcd_internal so we should be using EB_MEMCPY which points to eb_memcpy_app ( equivalent to eb_memcpy) and it points to the c version memcpy if x86 architecture is not detected.
To summarize, I prefer that we call EB_MEMCPY for this fix rather than memcpy to keep calling the optimized code by default until we figure out a way to switch between c and optimized code for the app and manual prediction case when asm is being set by user.
I understand that this might be confusing but I hope this clarifies things a bit.

ePirat:
Yeah this is kind of what I understood already but I do not really see the benefit at all in these cases here to call an optimized memcpy, as this is not a hot code path that copies image data or so?

ePirat:
Oh and what was the reason for changing it in #1155 in the first place?

ttrigui:
> Yeah this is kind of what I understood already but I do not really see the benefit at all in these cases here to call an optimized memcpy, as this is not a hot code path that copies image data or so?

For the app it matters as it copies pictures. Here it is only the prediction structure so the reason for using EB_MEMCPY is mainly consistency.

ttrigui:
> Oh and what was the reason for changing it in #1155 in the first place?

We got some feedback to fix -asm 0 to use c code. And since #1155 is mainly about building c only code, we chose to change it there.

ePirat:
Ok, I can change it to `EB_MEMCPY` in a moment.

ttrigui:
> Ok, I can change it to `EB_MEMCPY` in a moment.

thank you.

ePirat:
I've updated the code and commit message.

PR #1226, opened 2020-04-17
Snap: Use 'git' as version

ePirat:
The magic version 'git' will be replaced with the git describe string
as per the snapcraft.yaml documentation:
https://snapcraft.io/docs/snapcraft-yaml-reference


hassount:
@markshuttle @EwoutH @jbkempf FYI
does this address the improvements needed in #38 ?

PR #1232, opened 2020-04-19
Docs: Tell correct SCM default mode in EbAppConfig.C

EwoutH:
Fixes a small command line documentation forgotten when scm was defaulted to 0 in #1005.


EwoutH:
Thanks for merging! Is any work planned to improve screen content detecting? I noticed it made quite a big difference in some clips!

hassount:
Not in the near future as this is considered an application issue rather than an encoder one.

PR #1243, opened 2020-04-22
Fixed CLI string in E2E tests

Cidana-Developers:
1. Replaces hard coding of CLI token with getting from SvtAv1EncApp helper


hassount:
@chkngit FYI

PR #1244, opened 2020-04-22
ffmpeg: fix memory overwrite for complex video content

xuguangxin:
previous code assumes raw frame size > tu size, it's not true.
A tu may include multiple undisplayable frames + one displayable frame.
So for some complex video pattern, a tu size may > multiple raw frame size.
Instead of using raw frame size as ffmpeg buffer size, This commit will
automatically adjust ffmpeg buffer size based on libsvtav1's output

test:
dd if=/dev/urandom bs=115200 count=30 of=test.yuv
ffmpeg -f rawvideo -vcodec rawvideo -s 320x240 -r 30 -pix_fmt yuv420p -i test.yuv -c:v libsvt_av1 -f ivf -y test.ivf

expected result:
no crash.


tianjunwork:
Test pass with with the failing clip. Thanks guangxin.

PR #1249, opened 2020-04-22
CI: Add sanitizer runs

ePirat:
closes #1249

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,76 @@
+name: Sanitizers
+on:
+  push:
+    paths-ignore:
+      - "Docs/*"
+      - .travis.yml
+      - appveyor.yml
+      - .gitignore
+      - "*.md"
+  pull_request:
+    paths-ignore:
+      - "Docs/*"
+      - .travis.yml
+      - appveyor.yml
+      - .gitignore
+      - "*.md"
+env:
+  CC: clang
+  CXX: clang++
+
+jobs:
+  sanitizers:
+    name: Sanitizer
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: ['address', 'memory', 'thread']
+        include:
+          - sanitizer: address
+            allows-failure: false
+            env:
+              ASAN_OPTIONS: detect_leaks=1
+              CFLAGS: -fsanitize=address
+              CXXFLAGS: -fsanitize=address
+              LDFLAGS: -fsanitize=address
+
+          - sanitizer: memory
+            allows-failure: false
+            env:
+              CFLAGS: -fsanitize=memory -fno-omit-frame-pointer
+              CXXFLAGS: -fsanitize=memory -fno-omit-frame-pointer
+              LDFLAGS: -fsanitize=memory
+            svt-args: '--asm 0'
+
+          - sanitizer: thread
+            allows-failure: true
+            env:
+              CFLAGS: -fsanitize=thread
+              CXXFLAGS: -fsanitize=thread
+              LDFLAGS: -fsanitize=thread
+    env:
+        ${{ matrix.env }}
+
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y yasm ninja-build
+
+    - uses: actions/checkout@v2
+
+    - name: Run CMake
+      run: cmake -S . -B Build -G Ninja -DCMAKE_BUILD_TYPE=Debug
+
+    - name: Build
+      run: cmake --build Build
+
+    - name: Fetch test videos
+      run: |
+        wget -nc https://raw.githubusercontent.com/OpenVisualCloud/SVT-AV1-Resources/master/video.tar.gz
+        tar xzf video.tar.gz
+
+    - name: Run Encoder
+      run: ./Bin/Debug/SvtAv1EncApp -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m ${{ matrix.svt-args }} -n 120 --preset 8 -b output.ivf
+      continue-on-error: ${{ matrix.allows-failure }}
--- a/Source/Lib/Common/Codec/EbThreads.c
+++ b/Source/Lib/Common/Codec/EbThreads.c
@@ -9,6 +9,12 @@
 // and mutexs.  The goal is to eliminiate platform #define
 // in the code.
 
+#if defined(__has_feature)
+#  if __has_feature(thread_sanitizer)
+#    define EB_THREAD_SANITIZER_ENABLED
+#  endif
+#endif
+
 /****************************************
  * Universal Includes
  ****************************************/
@@ -70,6 +76,7 @@ EbHandle eb_create_thread(void *thread_function(void *), void *thread_context) {
     if (th == NULL)
         return NULL;
 
+#ifndef EB_THREAD_SANITIZER_ENABLED
     pthread_attr_init(&attr);
     pthread_attr_setschedpolicy(&attr, SCHED_FIFO);
     pthread_attr_setinheritsched(&attr, PTHREAD_EXPLICIT_SCHED);
@@ -85,6 +92,15 @@ EbHandle eb_create_thread(void *thread_function(void *), void *thread_context) {
         // parameters failed, retry creating the thread without them.
         ret = pthread_create(th, NULL, thread_function, thread_context);
     }
+#else
+    // When running with thread sanitizer, we are not running as root
+    // so the above priority change will always fail, which will cause
+    // issues with the thread sanitizer.
+    // See https://github.com/google/sanitizers/issues/1088
+    // Therefore we never try this, with the thread sanitizer
+    // and just create a normal thread here.
+    ret = pthread_create(th, NULL, thread_function, thread_context);
+#endif
 
     if (ret != 0) {
         free(th);


.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,112 @@
...
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y yasm ninja-build
+
+    - uses: actions/checkout@v2
+
+    - name: Run CMake
+      run: cmake -B Build -G Ninja -DCMAKE_BUILD_TYPE=Debug \
```
1480c1, line 26: You can use `-S .` to specify source instead of the trailing `.` you currently have

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,112 @@
...
+             -DCMAKE_C_FLAGS="-fsanitize=address" -DCMAKE_CXX_FLAGS="-fsanitize=address" \
```
1480c1, line 27: The compiler flags can be set by `CFLAGS` and `CXXFLAGS`

https://github.com/Kitware/CMake/blob/master/Modules/CMakeCInformation.cmake

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,112 @@
...
+             -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" -DCMAKE_MODULE_LINKER_FLAGS="-fsanitize=address" .
```
1480c1, line 28: The Linker flags can be set by `LDFLAGS`

https://github.com/Kitware/CMake/blob/07cfb18f9d29cfc0588ede928846a03ec5599c48/Modules/CMakeCommonLanguageInclude.cmake

.github/workflows/sanitizers.yml
ePirat: I had tried that and it made CMake try to interpret following options to belong to the source directory, I have no idea why…

.github/workflows/sanitizers.yml
ePirat: I can't do -DCFLAGS though, can I?

.github/workflows/sanitizers.yml
ePirat: Although it was using an CI env variable so maybe that was the issue, I'll try again with `.`

.github/workflows/sanitizers.yml
1480c1: I mean environment variable, so with
```yaml
env:
  CFLAGS: ...
  CXXFLAGS: ...
```

.github/workflows/sanitizers.yml
ePirat: Works with `.`, changed.

.github/workflows/sanitizers.yml
ePirat: Good point! Thanks. It's more readable and less verbose like that.

.github/workflows/sanitizers.yml
ePirat: Done.

ePirat:
The failures look legitimate due to bugs in SVT-AV1.

ePirat:
@1480c1 Thanks a lot for the review, it should be much cleaner now! 👍

1480c1:
@ePirat would it be possible to combine the jobs using a matrix, or would that make it too ugly?

ePirat:
@1480c1 Oh, nice idea, as it's mostly just different env vars now, this might actually work.

andrey-norkin:
Regarding the two sanitizers failures, do we want to fix it before merging the PR? Or should we merge first and fix later? @chkngit, any opinions on that?

@ePirat, are you planning to update the PR according to your yesterday comment or would you like to merge it as is?

tianjunwork:
Another question, is below sanity test good enough for sanitizer run?
`./Bin/Debug/SvtAv1EncApp -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m -n 120 --preset 8 -b output.ivf`

ePirat:
@tianjunwork it should probably be extended with more samples but this needs insight from people more familiar with the code so they can suggest good samples that actually test different codepaths and not just the same over and over, as thats pointless.

@andrey-norkin Yeah I'll update the PR according to what @1480c1 suggested.

chkngit:
we should try to fix these issues now, they might be straightforward fixes, if not we can just postpone

ePirat:
@1480c1 Simplified!

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,65 @@
...
+        wget -nc https://raw.githubusercontent.com/OpenVisualCloud/SVT-AV1-Resources/master/video.tar.gz
+        tar xzf video.tar.gz
+
+    - name: Run Encoder
+      run: ./Bin/Debug/SvtAv1EncApp -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m \
+             ${{ matrix.svt-args }} -n 120 --preset 8 -b output.ivf
+
+    - name: Cleanup
+      if: always()
+      run: rm -f output.ivf
```
1480c1, line 65: Is the cleanup necessary? We aren't doing anything to the image that we need to preserve

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,65 @@
...
+jobs:
+  sanitizers:
+    name: Sanitizer
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: ['address', 'memory', 'thread']
+        include:
+          - sanitizer: 'address'
```
1480c1, line 18: these don't really need single quotations marks, anything after the `:` is generally taken as a string unless there are some special rules

1480c1:
Also @ePirat maybe this might also be a good time to review the styles for the actions yaml files since we seem to have multiple different styles between files and also between jobs in the same file

.github/workflows/sanitizers.yml
ePirat: Not really necessary, it just felt a bit nicer to clean it up. I can remove it.

ePirat:
@1480c1 Feel free to file an issue and propose something about that or just make a PR changing them, I have not really any strong opinions on any of this as long as we are consistent.

spawlows:
> we should try to fix these issues now, they might be straightforward fixes, if not we can just postpone

Hi, We are working to define how difficult will be to fix all issues. Until we not know root cause of issues, we can not decide that sanitizer should block CI, and we should wait until fixes from cppcheck and static check analyzer will be merged.

ePirat:
Updated to address the remaining comments by @1480c1

spawlows:
Hi, @ePirat 

For memory sanitizer you run cmd with "--asm 0", but output shows that encoder still use AVX2:
[asm level selected : up to avx2]

Commandline: 
SvtAv1EncApp -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m \ --asm 0 -n 120 --preset 8 -b output.ivf

"\ --asm 0" this is parsed in code as: " --asm 0" (param start with space) and it is skipped.
Could you fix this?

spawlows:
Hi, @ePirat

Could you rebase branch to master? (Sanitizer memory should pass).

ePirat:
@spawlows Thanks for the info, rebased.

spawlows:
Hi @ePirat,

About Sanitizer (thread) error:

`ThreadSanitizer:DEADLYSIGNAL`
`==6803==ERROR: ThreadSanitizer: SEGV on unknown address 0x000000001048 (pc 0x559cdf669671 bp 0x7fe4066efe60 sp 0x7fe4066efc00 T6807)`
`==6803==The signal is caused by a READ memory access.`
`ThreadSanitizer:DEADLYSIGNAL`

I have that same error, when I run test without root privileges, even for small videos.
I am worried, that you have to use root privileges in CI, or disable this type of tests.

ePirat:
@spawlows Updated the PR which should fix this.

Source/Lib/Common/Codec/EbThreads.c
```diff
@@ -9,6 +9,12 @@
 // and mutexs.  The goal is to eliminiate platform #define
 // in the code.
 
+#if defined(__has_feature)
+#  if __has_feature(thread_sanitizer)
+#    define EB_TSAN_ENABLED 1
```
tianjunwork, line 14: you just need `#    define EB_TSAN_ENABLED` since you use `#ifndef` below.
I think full name EB_THREAD_SANITIZER_ENABLED is easier to read.

ePirat: Updated.

andrey-norkin:
@ePirat, could you rerun the tests?

ePirat:
@andrey-norkin You mean rebase to master and run again?

spawlows:
Hi,
I analyzed few warnings from Sanitizer(Thread). Fixing all warnings is to complicated to be done in few days. I did not see any warning which can affect output. 
Sanitizer thread pointed out interesting places in the code. Some members in SequenceControlSet set from static_config should be configured during encoder initialization, not repeatable updated in the thread. To fix this it need deeper reorganization. I assume that fixing all warnings could increase performance by ~1-2% for M8

I propose change test Sanitizer(thread) to allow to fail and merge this PR.

tianjunwork:
Hi @ePirat , could you change Sanitizer (thread) to `allow to fail` for now. So that we can move forward with this PR.

tianjunwork:
Reading the document, it seems that `continue-on-error` should add to `jobs.steps` so that the `sanitizer thread` job can show pass? Otherwise, I didn't see any difference of the CI result compared with previous commit.
```
jobs.<job_id>.steps.continue-on-error
Prevents a job from failing when a step fails. Set to true to allow a job to pass when this step fails.
```
Will it involve too much change? If so, I am fine with what we have now.

ePirat:
I copied that from the documentation, so I expected it to work… I'll have another look tomorrow.

tianjunwork:
Thank you. I think you are currently using below setting, which has no effect on job itself.
```
jobs.<job_id>.continue-on-error
Prevents a workflow run from failing when a job fails. Set to true to allow a workflow run to pass when this job fails.
```

ePirat:
Well it says "Prevents a workflow run from failing" which apparently does not work? I am really confused about this… In general the way Github actions handles jobs that are allowed to fail is extremely annoying as it just has two states, succeeded and failed… So there is no middle-ground like an "pass with warnings" or so for jobs that are allowed to fail… Instead it shows it as succeeded and makes it quite hard to know it did not actually succeed…

tianjunwork:
Hi @ePirat , could you try `jobs.<job_id>.steps.continue-on-error`? If there is no change to the result, let's just remove `continue-on-error` all together. How about that?

ePirat:
@tianjunwork Sure, let's see if it works now…

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,64 @@
+name: Sanitizers
+on:
+  - push
+  - pull_request
```
1480c1, line 4: Can you modify this to include the paths-ignore?

```yaml
on:
  push:
    paths-ignore:
      - "Docs/*"
      - .travis.yml
      - appveyor.yml
      - .gitignore
      - "*.md"
  pull_request:
    paths-ignore:
      - "Docs/*"
      - .travis.yml
      - appveyor.yml
      - .gitignore
      - "*.md"
```

1480c1:
Also, would it be possible to use ccache here?

1480c1:
@ePirat 🏓

ePirat:
Sure, I'll update the PR.

Regarding Ccache maybe you can add that in a follow-up PR? I am not really sure whats the "proper" way to do it with the other cleanups going on.

1480c1:
> Sure, I'll update the PR.
> 
> Regarding Ccache maybe you can add that in a follow-up PR? I am not really sure whats the "proper" way to do it with the other cleanups going on.

Sure I can do that, I sort of want this PR to get mered as soon as possible so I can rebase the cleanup PR on this work too

ePirat:
In general, do we know if the CCache stuff works fine, like did we actually see it ever caching things properly causing build speedups? As I would have expected it to have a bit bigger impact on build times.

1480c1:
Although I'm not sure about exact numbers, but looking at three jobs, it seems to at least give a good amount of decreased time

Original: 35s
https://github.com/1480c1/SVT-AV1/runs/687211609?check_suite_focus=true
image


First run after adding cache: 45s
https://github.com/1480c1/SVT-AV1/runs/687212891?check_suite_focus=true
image


Run using prepared cache: 3s
https://github.com/1480c1/SVT-AV1/runs/687218025?check_suite_focus=true
image

ePirat:
Indeed that looks good! Thanks!

tianjunwork:
Hi all reviewers, this PR is ready to review. Thanks @spawlows solving some of the issues.
Now with this PR, sanitizer memory/address can pass successfully. 
Sanitizer thread fails and needs more time to look into. https://github.com/OpenVisualCloud/SVT-AV1/pull/1249#issuecomment-628766207. It is set `allow to fail` for now. I will create an issue to track.
Please review. Thanks.

PR #1255, opened 2020-04-23
Actions: Add macOS ffmpeg build and functionality test

1480c1:


.github/workflows/ci.yml
```diff
@@ -287,15 +287,39 @@ jobs:
...
     needs: [build]
     env:
       CMAKE_EFLAGS: '-DBUILD_SHARED_LIBS=OFF'
+
     steps:
       - name: Install dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install -y ${native_deps}
+          if type apt-get > /dev/null 2>&1; then
```
ePirat, line 304: I would prefer if you do it like I did here at https://github.com/OpenVisualCloud/SVT-AV1/blob/master/.github/workflows/ci.yml#L44

.github/workflows/ci.yml
```diff
@@ -287,15 +287,39 @@ jobs:
...
+            echo "::set-env name=CC::ccache gcc"
+            echo "::set-env name=CXX::ccache g++"
+            gcc --version
+          elif type brew > /dev/null 2>&1; then
+            (
+              set +e
+              brew update
+              brew install yasm ccache cmake pkg-config
+            ) || true
+            echo "::set-env name=CC::ccache clang"
```
ePirat, line 316: Setting the env vars in the matrix will make this a lot more readable.

.github/workflows/ci.yml
```diff
@@ -305,27 +329,29 @@ jobs:
...
-          git apply $GITHUB_WORKSPACE/ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
+          git clone --branch release/4.2 --depth 1 https://github.com/FFmpeg/FFmpeg "$HOME/ffmpeg"
+          git -C "$HOME/ffmpeg" apply $GITHUB_WORKSPACE/ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
       - name: Configure ffmpeg
         run: |
           cd $HOME/ffmpeg
-          export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:/usr/local/lib/pkgconfig
-          export CFLAGS=""
-          ./configure --enable-libsvtav1 --cc="ccache gcc" --cxx="ccache g++" || less ffbuild/config.log
+          export LD_LIBRARY_PATH=/usr/local/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH} PKG_CONFIG_PATH=/usr/local/lib/pkgconfig${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}
```
ePirat, line 343: Why are you doing `${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH} ` here? (Same for PKG_CONFIG_PATH below)

.github/workflows/ci.yml
```diff
@@ -305,27 +329,29 @@ jobs:
...
+          export CFLAGS=''
+          ./configure --enable-libsvtav1 --cc="${CC:-ccache gcc}" --cxx="${CXX:-ccache g++}" || { less ffbuild/config.log; exit 1; }
```
ePirat, line 345: I don't think there is a need to do this, the env vars should work just fine.

1480c1: These are all copy overs from travis-ci.yml FFmpeg patch job
The primary intention was to move that job from travis-ci to github actions and add a line to actually run `ffmpeg` since it was removed when we disabled coveralls

1480c1: If you are talking about the `--cc` and `--cxx`, that's because FFmpeg doesn't actually adhere to the `CC` and `CXX` variables

.github/workflows/ci.yml
1480c1: copy-overs from travis-ci

.github/workflows/ci.yml
1480c1: Done

.github/workflows/ci.yml
1480c1: Done

.github/workflows/ci.yml
```diff
@@ -287,15 +287,48 @@ jobs:
...
+          cmake --version
+          ccache --version
+      - name: Install dependencies (mac)
+        if: matrix.os == 'macos-10.15'
+        run: |
+          (
+            set +e
+            brew update
+            brew install yasm ccache cmake pkg-config
+          ) || true # Since brew seems to exit 1 even though it was successful
```
ePirat, line 327: That's weird, I never had that happen with brew… Are you sure there isn't some error happening?

.github/workflows/ci.yml
ePirat: Ok, I guess we can simplify it later.

.github/workflows/ci.yml
1480c1: I think this was one of the jobs that failed, but I will try without the subshell and see again

.github/workflows/ci.yml
ePirat: Oh right I forgot the ffmpeg configure script is a bit weird and does not really respect usual conventions there…

1480c1: Yeah, it really annoyed me when I kept on trying to build FFmpeg with not the default `CC`, but it never changed anything and I had to actually look through the configure script for any instance of `CC`

.github/workflows/ci.yml
ePirat: Oh and there is no need to install CMake, the runner already has Cmake 3.17.1

1480c1: Seems maybe it was a one time only kind of thing

tianjunwork:
Hi @ePirat @joelsole , good to go?

ePirat:
@tianjunwork Ok from my side

PR #1258, opened 2020-04-24
Fix R2R in non-8 aligned resolution streams

lijing0010:

### DIFF
--- a/Source/Lib/Common/Codec/EbDefinitions.h
+++ b/Source/Lib/Common/Codec/EbDefinitions.h
@@ -101,6 +101,7 @@ extern "C" {
 #define MRP_31B_SUPPORT 1
 #define TILES_PARALLEL 1
 #define R2R_FIX 1
+#define R2R_FIX_PADDING 1
 
 #if TILES_PARALLEL
 #define MAX_TILE_CNTS 128 // Annex A.3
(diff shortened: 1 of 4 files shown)


tianjunwork:
Hi @lijing0010 , if it is to fix a bug, why use Macro around the fix? Is it related to dev process?

lijing0010:
> Hi @lijing0010 , if it is to fix a bug, why use Macro around the fix? Is it related to dev process?

yeah, I put all my changes into a macro in case someone want to debug on regression related issues.
Also helps to cherry-pick between branches. 
But will make the code a bit hard to read. anyway I think we may clean up those macros when achieve some stable version or a release

tianjunwork:
I see, thank you Jing.

EricLamVCD:
Tested 12 random clips with mixed resolutions of 6 - 8bit and 6 10bit clips: 
  | Speed Deviation  | BD-Rate   Deviation
-- | -- | --
Tested | -0.39% | 0.00%

lijing0010:
> can you please explain the problem ?

@chkngit . It's caused by padding. 
For non-8 aligned bitstreams, we need to:
1) first padding to 8-aligned
2) then padding around(128+32 pixels).
However in previous implementations, some part forgot to do padding 1), or did 2) before 1).

tianjunwork:
Hi @chkngit , is this PR good to go?

chkngit:
@lijing0010  could you please review related PR #1263 as it seems related to this, and port this fix to svt-01 and svt_av1_2020_04_25 branch?

PR #1257, opened 2020-04-23
revert log2f changes due to rate loss

ttrigui:
move back log2f/LOG2F to EbDefinitions.h/EbUtility.h due to rate degradation caused by moving it to common_dsp_rtcd.c until further investigation

PR #1266, opened 2020-04-27
Fix log2f, and revert "Revert log2f changes due to rate loss

### LINKED ISSUE #1261: Adding the ability to switch between c and optimized code for Log2f caused a rate loss
#1257

spawlows:
in M1 kirkland (#1257)"

uint Log2 should return 0 for arg 0.

Fixes #1261 
Protecting kernel Log2f_ASM() will return 0 when argument is 0.
In argument always sets first bit, to not get undefined behavior from asm bsr.
Results not changes for values different that 0.

### DIFF
--- a/Source/Lib/Common/ASM_SSE2/EbPictureOperators_SSE2.asm
+++ b/Source/Lib/Common/ASM_SSE2/EbPictureOperators_SSE2.asm
@@ -480,6 +480,8 @@ Label_PictureAverageKernel_SSE2_WIDTH16:
     ret
 
 ; ----------------------------------------------------------------------------------------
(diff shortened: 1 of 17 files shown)


spawlows:
Closes #1261

ttrigui:
@OmranAbdelkafi can you please confirm that with this fix the kirkland issue in M1 is fixed?

tianjunwork:
Hi @mafonsoN , thanks reviewing previous related PR. Could you review this one also? Thank you:)

tianjunwork:
Thank you @mafonsoN 👍

PR #1275, opened 2020-05-01
Multi thread buffer tuning

ttrigui:
Add the ability to pin/unpin to/from cores for any number of cores(used to be restricted to 1 core)
Tune buffer settings to get a better tradeoff point for user/real time going from lp 1 to lp 8

### DIFF
--- a/Docs/svt-av1_encoder_user_guide.md
+++ b/Docs/svt-av1_encoder_user_guide.md
@@ -290,11 +290,11 @@ so -lp 4 with -unpin 0 would restrict the encoder to work on cpu0-3 and reduce t
 
 Example: 72 core machine:
 
-72 cores x --lp 1 --unpin 1 (In order to maximize the CPU utilization 72 jobs are run simultaneously with each job utilitizing 1 core without being pined to a specific core)
+72 jobs x --lp 1 --unpin 1 (In order to maximize the CPU utilization 72 jobs are run simultaneously with each job utilitizing 1 core without being pined to a specific core)
 
-36 cores x --lp 2 --unpin 1
+36 jobs x --lp 2 --unpin 1
 
-18 cores x --lp 4 --unpin 1
+18 jobs x --lp 4 --unpin 1
 
 (`-ss`) and (`-unpin 1`) is not a valid combination.(`-unpin`) is overwritten to 0 when (`-ss`) is used.
 ## Legal Disclaimer
--- a/Docs/svt-av1_encoder_user_guide.md
+++ b/Docs/svt-av1_encoder_user_guide.md
@@ -149,7 +149,7 @@ The encoder parameters present in the `Sample.cfg` file are listed in this table
 | **HighDynamicRangeInput** | --enable-hdr | [0-1, 0 for default] | 0 | Enable high dynamic range(0: OFF[default], ON: 1) |
 | **Asm** | --asm |  [0 - 11] or [c, mmx, sse, sse2, sse3, ssse3, sse4_1, sse4_2, avx, avx2, avx512, max] | 11 or max | Limit assembly instruction set ("0" is equivalent to "c", "1" is "mmx" etc, max value is "11" or "max"), by default select highest assembly instruction that is supported by CPU |
 | **LogicalProcessorNumber** | --lp | [0, total number of logical processor] | 0 | The number of logical processor which encoder threads run on.Refer to Appendix A.1 |
-| **UnpinSingleCoreExecution** | --unpin-lp1 | [0, 1] | 1 | Unpin the execution . If logical_processors is set to 1, this option does not set the execution to be pinned to core #0 when set to 1. this allows the execution of multiple encodes on the CPU without having to pin them to a specific mask  0=OFF, 1= ON |
+| **UnpinExecution** | --unpin | [0, 1] | 0 | Allows the execution to be pined/unpined to/from a specific number of cores.--unpin is overwritten to 0 when --ss is set to 0 or 1. 0=OFF, 1= ON |
 | **TargetSocket** | --ss | [-1,1] | -1 | For dual socket systems, this can specify which socket the encoder runs on.Refer to Appendix A.1 |
 
 #### Rate Control Options
@@ -286,14 +286,17 @@ The combinational use of (`-unpin`)  with (`-lp`) results in memory reduction wh
 
 This is an example on how to use them together.
 
-so -lp 4 with unpin 0 would restrict the encoder to work on cpu0-3 and reduce the resource allocation to only what's needed to using 4 cores. -lp 4 with -unpin 1, would reduce the allocation to what's needed for 4 cores but not restrict the encoder to run on cpu 0-3, in this case the encoder might end up using more than 4 cores due to the multi-threading nature of the encoder, but would at least allow for more multiple -lp4 encodes to run on the same machine without them being all restricted to run on cpu 0-3 or overflow the memory usage.
+so -lp 4 with -unpin 0 would restrict the encoder to work on cpu0-3 and reduce the resource allocation to only what's needed to using 4 cores. -lp 4 with -unpin 1, would reduce the allocation to what's needed for 4 cores but not restrict the encoder to run on cpu 0-3, in this case the encoder might end up using more than 4 cores due to the multi-threading nature of the encoder, but would at least allow for more multiple -lp4 encodes to run on the same machine without them being all restricted to run on cpu 0-3 or overflow the memory usage.
 
 Example: 72 core machine:
-72 cores x -lp 1 -unpin 1
-36 cores x -lp 2 -unpin 1
-18 cores x -lp 4 -unpin 1
 
-(`-ss`) and (`-unpin 1`) is not a valid combination.(`-unpin `) is overwritten to 0 when (`-ss`) is used.
+72 cores x --lp 1 --unpin 1 (In order to maximize the CPU utilization 72 jobs are run simultaneously with each job utilitizing 1 core without being pined to a specific core)
+
+36 cores x --lp 2 --unpin 1
+
+18 cores x --lp 4 --unpin 1
+
+(`-ss`) and (`-unpin 1`) is not a valid combination.(`-unpin`) is overwritten to 0 when (`-ss`) is used.
 ## Legal Disclaimer
 
 ### Optimization Notice
--- a/Docs/svt-av1_encoder_user_guide.md
+++ b/Docs/svt-av1_encoder_user_guide.md
@@ -280,6 +280,20 @@ If only TargetSocket is set, threads run on all the logical processors of socket
 
 If both LogicalProcessorNumber and TargetSocket are set, threads run on 20 logical processors of socket 0. Threads guaranteed to run only on socket 0 if 20 is larger than logical processor number of socket 0.
 
+The (`-unpin`) option allows the user to pin/unpin the execution to/from a specific number of cores.
+
+The combinational use of (`-unpin`)  with (`-lp`) results in memory reduction while allowing the execution to work on any of the cores and not restrict it to specific cores.
+
+This is an example on how to use them together.
+
+so -lp 4 with unpin 0 would restrict the encoder to work on cpu0-3 and reduce the resource allocation to only what's needed to using 4 cores. -lp 4 with -unpin 1, would reduce the allocation to what's needed for 4 cores but not restrict the encoder to run on cpu 0-3, in this case the encoder might end up using more than 4 cores due to the multi-threading nature of the encoder, but would at least allow for more multiple -lp4 encodes to run on the same machine without them being all restricted to run on cpu 0-3 or overflow the memory usage.
+
+Example: 72 core machine:
+72 cores x -lp 1 -unpin 1
+36 cores x -lp 2 -unpin 1
+18 cores x -lp 4 -unpin 1
+
+(`-ss`) and (`-unpin 1`) is not a valid combination.(`-unpin `) is overwritten to 0 when (`-ss`) is used.
 ## Legal Disclaimer
 
 ### Optimization Notice
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -552,14 +552,14 @@ typedef struct EbSvtAv1EncConfiguration {
      * OS thread scheduler. */
     uint32_t logical_processors;
 
-    /* Unpin the execution . If logical_processors is set to 1, this option does not
-    * set the execution to be pinned to core #0 when set to 1. this allows the execution
+    /* Unpin the execution .This option does not
+    * set the execution to be pinned to a specific number of cores when set to 1. this allows the execution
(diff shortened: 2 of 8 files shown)


kylophone:
Curious, what is the advantage to pinning to core 0?

Source/API/EbSvtAv1Enc.h
```diff
@@ -552,14 +552,14 @@ typedef struct EbSvtAv1EncConfiguration {
...
     * of multiple encodes on the CPU wihtout having to pin them to a specific mask
-    * 1: unpinned from core 0
-    * 0: pinned to core 0
+    * 1: unpinned
+    * 0: pinned
     *
-    * Default is 1. */
-    uint32_t unpin_lp1;
+    * If logical_processors is set to 1 default is 1 otherwise it is 0. */
+    uint32_t unpin;
```
tianjunwork, line 562: `unpin` new name is fine, but I think the old description is more accurate and better.

tianjunwork:
> Curious, what is the advantage to pinning to core 0?
when use `--lp 1`, all the threads is pinned to logical processor #0 which reduce thread context switching. Not only `--lp 1`, `--lp N` pin threads from 0 to N-1.

Source/API/EbSvtAv1Enc.h
ttrigui: The only thing I changed is that I made the unpin option to be generic to work on any number of cores not only when logical_processors is set to 1. I tried to change the description accordingly. Please let me know if you propose a better description.

ttrigui: The only thing I changed is that I made the unpin option to be generic to work on any number of cores not only when logical_processors is set to 1. I tried to change the description accordingly. Please let me know if you propose a better description.

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -254,7 +254,7 @@ void eb_set_thread_management_parameters(EbSvtAv1EncConfiguration *config_ptr)
     }
 #elif defined(__linux__)
     uint32_t num_logical_processors = get_num_processors();
-    if (config_ptr->logical_processors == 1 && config_ptr->unpin_lp1 == 1) {
+    if (config_ptr->logical_processors == 1 && config_ptr->unpin == 1) {
```
tianjunwork, line 257: Hi @ttrigui, from the logic here, `unpin` still only takes effect when `-lp 1`.

ttrigui: Hi @tianjunwork I did not change the logic inside the function call but I'm only calling eb_set_thread_management_parameters when unpin is set to 0 (default or if user specifies) or if logical_processors is set to one.

tianjunwork: Hi @ttrigui , welcome back. You commented `unpin option to be generic to work on any number of cores not only when logical_processors is set to 1`. So seems that we need to clean up all the `config_ptr->logical_processors == 1` check with `unpin` in this file and also in `eb_set_thread_management_parameters` function.

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -1672,7 +1702,8 @@ EB_API EbErrorType svt_av1_enc_init(EbComponentType *svt_enc_component)
     * Thread Handles
     ************************************/
     EbSvtAv1EncConfiguration   *config_ptr = &enc_handle_ptr->scs_instance_array[0]->scs_ptr->static_config;
-
+    config_ptr->unpin = (config_ptr->logical_processors == 1) ? 1 : config_ptr->unpin;
```
tianjunwork, line 1705: If we want `unpin` for any number of `logical_processors`, we need to clean up `config_ptr->logical_processors == 1` here.

Source/Lib/Encoder/Globals/EbEncHandle.c
ttrigui: Hi @tianjunwork thank you :). I just want to clarify that this task meant to add this option to any number of cores except to 1 core. So in order to not touch the current state (i.e always pin when number of logical cores is 1) I added those checks. The user can still set unpin to 0 or 1 regardless of the number of cores for that reason I had to force unpin to be 1 when number of logical cores is 1. I hope this clarifies a little. If you suggest further clean up based on this please let me know. Thank you for the feedback.

tianjunwork: I am not sure about what is the history of why `-lp 1` is special. The different behavior causes user to be confused.

ttrigui: @tianjunwork ok can we get this PR in for now ? I'll bring up your suggestion and we can have a second commit as a "cleanup" later?

tianjunwork: sure.

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -1672,7 +1702,8 @@ EB_API EbErrorType svt_av1_enc_init(EbComponentType *svt_enc_component)
...
+    if ((config_ptr->logical_processors == 1) || (config_ptr->unpin == 0))
```
tianjunwork, line 1706: With `-lp > 1 or -ss` and `-unpin 0`,  since eb_set_thread_management_parameters is not called anymore, threads are not setting affinity by `-lp -ss`. Not sure that's what `unpin` is for? It is the same as not using `-lp and -unpin`.
btw, `-lp and -ss` is broken on master, threads are following the affinity setting at the beginning, but are scheduled to use all the cores later on.

Source/Lib/Encoder/Globals/EbEncHandle.c
hassount: @tianjunwork @ttrigui 
- Agreed that ```config_ptr->logical_processors == 1``` checks should be removed from inside the ```eb_set_thread_management_parameters```
- The unpin check could just happen before the call of the ```eb_set_thread_management_parameters``` function

Source/Lib/Encoder/Globals/EbEncHandle.c
hassount: agreed, this line would force unpin to be always off to on if -lp =1 which should not be the case.

Source/Lib/Encoder/Globals/EbEncHandle.c
hassount: @ttrigui @tianjunwork thank you both for the work and diligence on this PR
Using unpin with lp would result in reducing the amount of resources (memory) used by the encoder while allowing it to work on any of the cores of the machine and not restrict it to specific cores. 
so -lp 4 with unpin 0 would restrict the encoder to work on cpu0-3 and reduce the resource allocation to only what's needed to using 4 cores. -lp 4 with -unpin 1, would reduce the allocation to what's needed for 4 cores but not restrict the encoder to run on cpu 0-3, in this case the encoder might end up using more than 4 cores due to the multi-threading nature of the encoder, but would at least allow for more multiple -lp4 encodes to run on the same machine without them being all restricted to run on cpu 0-3 or overflow the memory usage.

in practice, if we have a 72 core machine, we can run:
72 cores x -lp 1 -unpin 1
36 cores x -lp 2 -unpin 1
18 cores x -lp 4 -unpin 1
and so on, doing our best to allow the encoder to machine to be fully utilized without overflowing the memory.

without this change, in order to fully utilize the machine, we could only run the following:
72 jobs x -lp 1 -unpin-lp 1
1 job x -lp 72

Now as for -ss, I agree that -unpin 1 and -ss x as a combination would not make sense anymore and should probably be blocked or that when using -ss we overwrite -unpin to be always 0.

last but not least, agreeing with Jun that all of the above should be clearly described in the documentation. @ttrigui could you please add a section in the users guide about the use of -lp / -unpin, also in EbAppConfig.c for a good description to show when using --help along with an updated description in the configuration file.

tianjunwork:
I see. Thanks @hassount for the explanation about `--lp`. So the description of it should also be changed. It is not only constraint of logical processor number, but also constraint of resources (memory) which match with that number(-unpin 1).
How about below steps to this kind of interface that needs more detail:
1. add correct description of the interface in user guide.
2. add detailed explanation/usage in appendix(if needed). or have a separate doc as tuning guide.

Docs/svt-av1_encoder_user_guide.md
```diff
@@ -280,6 +280,20 @@ If only TargetSocket is set, threads run on all the logical processors of socket
...
+The combinational use of (`-unpin`)  with (`-lp`) results in memory reduction while allowing the execution to work on any of the cores and not restrict it to specific cores.
+
+This is an example on how to use them together.
+
+so -lp 4 with unpin 0 would restrict the encoder to work on cpu0-3 and reduce the resource allocation to only what's needed to using 4 cores. -lp 4 with -unpin 1, would reduce the allocation to what's needed for 4 cores but not restrict the encoder to run on cpu 0-3, in this case the encoder might end up using more than 4 cores due to the multi-threading nature of the encoder, but would at least allow for more multiple -lp4 encodes to run on the same machine without them being all restricted to run on cpu 0-3 or overflow the memory usage.
+
+Example: 72 core machine:
+72 cores x -lp 1 -unpin 1
+36 cores x -lp 2 -unpin 1
+18 cores x -lp 4 -unpin 1
```
tianjunwork, line 294: The example `72 cores x -lp 1 -unpin 1` is not easy to understand. needs more words to explain.
Also need empty line between them, otherwise it is just one line if you view the markdown file.

Docs/svt-av1_encoder_user_guide.md
```diff
@@ -280,6 +280,20 @@ If only TargetSocket is set, threads run on all the logical processors of socket
...
+
+(`-ss`) and (`-unpin 1`) is not a valid combination.(`-unpin `) is overwritten to 0 when (`-ss`) is used.
```
tianjunwork, line 296: `-unpin ` one unnecessary space.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -799,10 +799,17 @@ ConfigEntry config_entry_global_options[] = {
...
      UNPIN_TOKEN,
-     "allows the execution of multiple encodes on the CPU without having to pin them to a "
-     "specific mask( 0: OFF ,1: ON) ",
+    "Allows the execution to be pined/unpined to/from a specific number of cores \n"
+    "The combinational use of --unpin with --l results in memory reduction while allowing the execution to work on any of the cores and not restrict it to specific cores \n"
+    "--unpin is overwritten to 0 when --ss is set to 0 or 1. ( 0: OFF [default] ,1: ON) \n"
+    "Example: 72 core machine: \n"
+    "72 cores x - lp 1 - unpin 1 \n"
+    "36 cores x - lp 2 - unpin 1 \n"
+    "18 cores x - lp 4 - unpin 1 ",
```
tianjunwork, line 808: `- lp 1 - unpin 1` ==> `--lp 1 --unpin 1`

Source/App/EncApp/EbAppConfig.c
```diff
@@ -799,10 +799,17 @@ ConfigEntry config_entry_global_options[] = {
     {SINGLE_INPUT, THREAD_MGMNT, "number of logical processors to be used", set_logical_processors},
     {SINGLE_INPUT,
      UNPIN_TOKEN,
-     "allows the execution of multiple encodes on the CPU without having to pin them to a "
-     "specific mask( 0: OFF ,1: ON) ",
+    "Allows the execution to be pined/unpined to/from a specific number of cores \n"
+    "The combinational use of --unpin with --l results in memory reduction while allowing the execution to work on any of the cores and not restrict it to specific cores \n"
```
tianjunwork, line 803: `--l` ==> `--lp`

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -1676,8 +1670,7 @@ EB_API EbErrorType svt_av1_enc_init(EbComponentType *svt_enc_component)
     * Thread Handles
     ************************************/
     EbSvtAv1EncConfiguration   *config_ptr = &enc_handle_ptr->scs_instance_array[0]->scs_ptr->static_config;
-    config_ptr->unpin = (config_ptr->logical_processors == 1) ? 1 : config_ptr->unpin;
-    if ((config_ptr->logical_processors == 1) || (config_ptr->unpin == 0))
+    if (config_ptr->unpin == 0)
```
tianjunwork, line 1673: better to have indent for `eb_set_thread_management_parameters(config_ptr);`

Docs/svt-av1_encoder_user_guide.md
```diff
@@ -280,6 +280,23 @@ If only TargetSocket is set, threads run on all the logical processors of socket
...
+
+The combinational use of (`-unpin`)  with (`-lp`) results in memory reduction while allowing the execution to work on any of the cores and not restrict it to specific cores.
+
+This is an example on how to use them together.
+
+so -lp 4 with -unpin 0 would restrict the encoder to work on cpu0-3 and reduce the resource allocation to only what's needed to using 4 cores. -lp 4 with -unpin 1, would reduce the allocation to what's needed for 4 cores but not restrict the encoder to run on cpu 0-3, in this case the encoder might end up using more than 4 cores due to the multi-threading nature of the encoder, but would at least allow for more multiple -lp4 encodes to run on the same machine without them being all restricted to run on cpu 0-3 or overflow the memory usage.
+
+Example: 72 core machine:
+
+72 cores x --lp 1 --unpin 1 (In order to maximize the CPU utilization 72 jobs are run simultaneously with each job utilitizing 1 core without being pined to a specific core)
```
tianjunwork, line 293: Thanks for the details. Now it is easier to understand.
Is `72 jobs` better than `72 cores` here?

Source/App/EncApp/EbAppConfig.c
```diff
@@ -798,11 +798,18 @@ ConfigEntry config_entry_global_options[] = {
...
-     set_unpin_single_core_execution},
-    {SINGLE_INPUT, TARGET_SOCKET, "Specify  which socket the encoder runs on", set_target_socket},
+     UNPIN_TOKEN,
+    "Allows the execution to be pined/unpined to/from a specific number of cores \n"
+    "The combinational use of --unpin with --lp results in memory reduction while allowing the execution to work on any of the cores and not restrict it to specific cores \n"
+    "--unpin is overwritten to 0 when --ss is set to 0 or 1. ( 0: OFF [default] ,1: ON) \n"
+    "Example: 72 core machine: \n"
+    "72 cores x -- lp 1 -- unpin 1 \n"
+    "36 cores x -- lp 2 -- unpin 1 \n"
+    "18 cores x -- lp 4 -- unpin 1 ",
```
tianjunwork, line 808: also here, should we use `jobs` instead of `cores`?

Docs/svt-av1_encoder_user_guide.md
ttrigui: sure I'll change it to avoid confusion. thank you for the feedback.

PR #1279, opened 2020-05-03
Travis: Re-enable travis-ci for pull requests

1480c1:
It seems that it was disabled in https://github.com/OpenVisualCloud/SVT-AV1/commit/6bc9f229d21e4803961a63ce469ffb80c29add0a


tianjunwork:
Re-enable Travis-ci which is disabled by accident.
Now it only runs for branches not PR.

PR #1260, opened 2020-04-24
More cppcheck warning fixes

ePirat:
This supersedes PR #1236

PR #1283, opened 2020-05-06
Fix wrong parameter for assert, 0 is valid

tszumski:
For Debug build with command line below assert is called
cmdline: 176x144.yuv -w 176 -h 144 -bit-depth 8 -enc-mode 0 -n 40 -b out.av1

@ePirat could you confirm that this is fixed correctly? (assert is introduced in PR #1260 )

### DIFF
--- a/Source/Lib/Encoder/Codec/EbMdRateEstimation.c
+++ b/Source/Lib/Encoder/Codec/EbMdRateEstimation.c
@@ -828,7 +828,7 @@ static AOM_INLINE void sum_intra_stats(PictureControlSet *pcs_ptr, BlkStruct *bl
     }
     if (av1_is_directional_mode(get_uv_mode(uv_mode)) && av1_use_angle_delta(bsize, pcs_ptr->parent_pcs_ptr->scs_ptr->static_config.intra_angle_delta)) {
         assert((uv_mode - UV_V_PRED) < DIRECTIONAL_MODES);
-        assert((uv_mode - UV_V_PRED) > 0);
+        assert((uv_mode - UV_V_PRED) >= 0);
         update_cdf(fc->angle_delta_cdf[uv_mode - UV_V_PRED],
                    mbmi->block_mi.angle_delta[PLANE_TYPE_UV] + MAX_ANGLE_DELTA,
                    2 * MAX_ANGLE_DELTA + 1);


ePirat:
@tszumski I don't understand what you mean by "fixed locally"

tszumski:
If you mean " fixed correctly", then I mean "Do you agree that this assert should have ">=" (more or equal) instead of ">" (more)

ePirat:
I did not write this assert so I don't know.

ePirat:
It seems @tianjunwork who merged #1260 squashed all commits, loosing the info who actually made the change.

ePirat:
@tianjunwork For the future please avoid this especially for such a variety of changes where some parts are not authored by me…

ePirat:
The specific change was done by either @hassount or @kylophone, apparently.

1480c1:
The change was introduced in 45d84ac79058c39270239b1d21e59fff4e2af0c4 by @hassount

tianjunwork:
Hi @ePirat , `Squash and merge` is now the way merge for this repo, which helps view git history and do regression test.

jbkempf:
@tianjunwork This is an exceptionally bad idea, and goes against every best practices.

Rebasing and avoiding merge commits helps bisecting, true.

But squashing commits does exactly the opposite, and makes both git history more complex and bissecting worse.
If you squash multiple commits, you have mega-commits which impact a lot of code and it's impossible to see at first glance what it was.

The recommendations is to do small commits, and not mega-commits. And you should rebase-and-merge.

CC @hassount

silverbacknet:
Is there a defined policy on when and how to squash? After all, many of the commits were simple one-or-two line fixes to small errors, many in a row, which are always demanded to be squashed on this project. Group by author instead of the full patchset? Group by changetype?

jbkempf:
> Is there a defined policy on when and how to squash? After all, many of the commits were simple one-or-two line fixes to small errors, many in a row, which are always demanded to be squashed on this project. Group by author instead of the full patchset? Group by changetype?

One issue -> one commit
One feature -> one commit

If you need 2 commits for the same  issue, you can squash them, but you should ask the author about it.

Also, tbh, there is also a legal implication, because as committed, you assigned his copyright to @hassount without his approval or his employer approval! You should never squash 2 commits from different authors, unless you are sure that they have common copyright!

tianjunwork:
Hi @jbkempf I am sorry for squashing #1260 with commits are from three different authors. Will use `rebase and merge` with PR in that case.
Other than that, if the commits are from one author, `Squash and merge` is preferred which improves repository commit history quality and references back to a PR where the whole changes are tracked.

jbkempf:
>  Other than that, if the commits are from one author, Squash and merge is preferred which improves repository commit history quality and references back to a PR where the whole changes are tracked.

No, it does not. Where did that idea come from?

Squashing different changes together is never a good idea, unless it is done on purpose, because the author wants to show the different revisions, to show his work process.
If you believe, 2 commits should be done in one, because they are fixing the same issue or adding the same feature, you ask the author to rebase locally.

Squashing 2 unrelated commits together does not improve commit history quality at all! On the contrary, when reviewing the git log you cannot understand what is happening.
And when you read the git logs, you do not see the PR. The git will stay the same long after the particular platform you are working on is gone... You have no idea what will become of Github in 5 years, or if the project will not be transferred to another hosting platform. Only the git history stay, and therefore, only this git log should be made nice.

tianjunwork:
I am not sure about other repo, but for SVT-*, commits under one PR are usually just to fix one issue. If they are not, maintainer would suggest separate them into different PRs. Commits are not unrelated. There are cases that many commits come in a row just for one issue:
1. Working a draft PR
2. Address review comments for multiple times
3. Habit of some developers to have small incremental commits. Some of the commits in between don't have nice git log and message.
Based on above reasons, `Squash and merge` does improve repository commit history quality.

I understand your worry towards Github that we may lose track of the commits.
Once @hassount is back, we can have more discussion about maintenance. Your suggestion is valued!

jbkempf:
Of course not, that is absolutely not reasonable, and the example is the example PR, with 23 commits that became a master commit changing things in 60 files, with over 250 changes!

Most PR have more than one commit, since they are doing multiple things on the same theme. Example being fixing analyzer reports or code cleaning, or CI. It makes no sense to do one PR for each, because you cannot review and understand one without the other.

As for incremental commits, you need to do _commit hygiene_ on your side. It's called git rebase -i, squash or tig/gitk to check it. It's not the reviewer's role to do it. You can write C++ and encoders, but you cannot do a decent git commit?
Your git history is ugly because the commiters did not do the work properly, not because you did not use squash.
_You cannot change the commit log for someone else, see legal implications already mentioned above._

So, it should never be the default.
It can be used, at the request of the submitter, if he does not want to do the work himself, but making this the default is unreasonable.

hassount:
@jbkempf 
Thank you for the valuable feedback on this. So in order to summarize (and we should probably add this to the contributing document when we agree):

- Make commits concise and to the point (1 feature / issue = 1 commit )
- Authors would be responsible for breaking down the PR into commits that make sense (with proper commit messages)
- Use rebase and merge for all PRs to make sure that all commits are carried over to the master
- Maintainers would only use squash and merge with the permission of the authors.

Would the above read right to all ? 

@jbkempf one major convenience that "squash and merge" would give us but not rebase and merge, is an automatic link to the PR which has been a major convenience when tracking issues and reading the full discussion on the PR. Now with the understanding that PRs might not be something visible if we port the repo somewhere else, it is still extremely helpful. Do you know if there's a way to set a link to the PR number to be automatically added to the top commit after rebasing and merging (assuming we wouldn't count on the authors to do so) ?

jbkempf:
Yes for the first part.

For the second, no idea, maybe @ePirat knows.

ePirat:
I don't think its possible to have the PR number in the commit message except when squashing or doing merge commits, as then Github writes the commit message. But in case of fast forward merges it does not touch the original commits, afaik.

1480c1:
@hassount for the point of an automatic link to the PR, in the github's site, if you click on a commit, it will show the PR related to the commits, for example https://github.com/OpenVisualCloud/SVT-AV1/commit/a0524a941cb2cd6e4b6ef635e7317280f45ddfb0

image
Near the bottom of the blue banner, there is `master (#1296)`, and clicking the `#1296` brings you to https://github.com/OpenVisualCloud/SVT-AV1/pull/1296

hassount:
@1480c1 Great thank you, it's not as convenient as the other reference but it is there at least. @tianjunwork can we then follow the guidelines below from now on ?

- Make commits concise and to the point (1 feature / issue = 1 commit )
- Authors would be responsible for breaking down the PR into commits that make sense (with proper commit messages)
- Use rebase and merge for all PRs to make sure that all commits are carried over to the master
- Maintainers would only use squash and merge with the permission of the authors.

tianjunwork:
Since these are something that better to have folks know, better to be written into CONTRIBUTING.md.
There will be some PR with commits which title/message only make sense within the context of this PR.  So maintainer will ask the author's choice how to go ahead?
1. author to rebase and provide better commit message, or
2. maintainer to squash

hassount:
> Since these are something that better to have folks know, better to be written into CONTRIBUTING.md.
> There will be some PR with commits which title/message only make sense within the context of this PR. So maintainer will ask the author's choice how to go ahead?
> 
> 1. author to rebase and provide better commit message, or
> 2. maintainer to squash

Agreed, in addition, it might be even better to have that answered as part of a template PR, i.e, the contributor would choose if they would take care of the commits or they would let the maintainer squash them, I ll try to look into this over the next few days (any suggestions are welcome too :) )

PR #1287, opened 2020-05-07
Cleanup: Remove always false skip arguments

ePirat:
**UPDATE**: This PR was updated to remove the skip parameter, see discussion. This makes the issue detailed below obsolete.

There is a tautology in the `avc_style_luma_interpolation_filter_posj_ssse3` function:

```
    if (skip)
        avc_style_luma_interpolation_filter_horizontal_ssse3_intrin(ref_pic - src_stride,
                                                                    src_stride,
                                                                    temp_buf,
                                                                    pu_width,
                                                                    pu_width,
                                                                    (pu_height + 3),
                                                                    0,
                                                                    EB_FALSE,
                                                                    2);
    else
        avc_style_luma_interpolation_filter_horizontal_ssse3_intrin(
            ref_pic - src_stride,
            src_stride,
            temp_buf,
            pu_width,
            pu_width,
            skip ? (2 * pu_height + 3) : (pu_height + 3),
            0,
            EB_FALSE,
            2);
```

If `skip` is true, the first branch is taken, calls `avc_style_luma_interpolation_filter_horizontal_ssse3_intrin` with `(pu_height + 3)`.

If `skip` is false, the second branch is taken, calls
`avc_style_luma_interpolation_filter_horizontal_ssse3_intrin` with `(pu_height + 3)`.

So there is no difference between the two branches, making them useless.
Additionally `avc_style_luma_interpolation_filter_horizontal_ssse3_intrin` has a skip parameter, why is that not used here? Is it not applicable?
Which variant of the code is correct? Maybe it was actually supposed to do what was done in the second branch and call `avc_style_luma_interpolation_filter_horizontal_ssse3_intrin` with `(2 * pu_height + 3)` when `skip` is true?

### DIFF
--- a/Source/Lib/Common/ASM_SSE2/EbAvcStyleMcp_Intrinsic_SSE2.c
+++ b/Source/Lib/Common/ASM_SSE2/EbAvcStyleMcp_Intrinsic_SSE2.c
@@ -8,19 +8,10 @@
 #include "emmintrin.h"
 #include "common_dsp_rtcd.h"
 void avc_style_copy_sse2(EbByte ref_pic, uint32_t src_stride, EbByte dst, uint32_t dst_stride,
-                         uint32_t pu_width, uint32_t pu_height, EbByte temp_buf, EbBool skip,
+                         uint32_t pu_width, uint32_t pu_height, EbByte temp_buf,
                          uint32_t frac_pos) {
     (void)temp_buf;
     (void)frac_pos;
-    if (skip) {
-        //do the last row too.
-        eb_memcpy(
-            dst + (pu_height - 1) * dst_stride, ref_pic + (pu_height - 1) * src_stride, pu_width);
-
-        src_stride <<= 1;
-        dst_stride <<= 1;
-        pu_height >>= 1;
-    }
 
     picture_copy_kernel_sse2(ref_pic, src_stride, dst, dst_stride, pu_width, pu_height);
 }
--- a/Source/Lib/Common/ASM_SSE2/EbAvcStyleMcp_SSE2.h
+++ b/Source/Lib/Common/ASM_SSE2/EbAvcStyleMcp_SSE2.h
@@ -12,7 +12,7 @@ extern "C" {
 #endif
 
 void avc_style_copy_sse2(EbByte ref_pic, uint32_t src_stride, EbByte dst, uint32_t dst_stride,
-                         uint32_t pu_width, uint32_t pu_height, EbByte temp_buf, EbBool skip,
+                         uint32_t pu_width, uint32_t pu_height, EbByte temp_buf,
                          uint32_t frac_pos);
 
 #ifdef __cplusplus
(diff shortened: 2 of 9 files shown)


1480c1:
@ePirat can you squash the commits, since the first commit doesn't do anything in the end

ePirat:
I can do that before merge but as per the comment I first want to know if this is even the right thing to do here.

1480c1:
```c
    if (skip)
        AvcStyleLumaInterpolationFilterHorizontal_SSSE3_INTRIN(ref_pic - src_stride, src_stride, tempBuf, pu_width, pu_width, (pu_height + 3), 0, EB_FALSE, 2);
    else
        AvcStyleLumaInterpolationFilterHorizontal_SSSE3_INTRIN(ref_pic - src_stride, src_stride, tempBuf, pu_width, pu_width, skip ? (2 * pu_height + 3) : (pu_height + 3), 0, EB_FALSE, 2);
```
seems they had it there since `initial commit`

tianjunwork:
I am not sure, `skip ? (2 * pu_height + 3) : (pu_height + 3),` looks more reasonable. Adding others to review.

ePirat:
Note that this is not really a cleanup but possibly a bugfix if `skip ? (2 * pu_height + 3) : (pu_height + 3)` is what the code should do, as it's not doing that currently.

tianjunwork:
> Note that this is not really a cleanup but possibly a bugfix if `skip ? (2 * pu_height + 3) : (pu_height + 3)` is what the code should do, as it's not doing that currently.

yeah, agree, I will re-label it based on review from others.

joelsole:
Shouldn't the goal be to match the c implementation?

tianjunwork:
Thanks @joelsole for the hint.
Hi @spawlows, need some help here. To match with c code, how should we fix?

spawlows:
Actual commits are fine, and will not effect on output. Alternatively you can cleanup the skip parameter from all code, and during cleanup check that really skip is always false. C code ignore parameter skip.

avc_style_luma_interpolation_filter_helper_c()
'  /* Code with 'skip' true are not used. Cleanup should remove 'skip' parameter. */
    /* frac_pos and fractional_position are redundant as well, cleanup should also unify the two*/
    (void)skip;
    assert(!skip);'

tianjunwork:
Hi @ePirat , would you like to remove skip parameter from `avc_style_luma_interpolation_filter` all together in this PR? It is not used in `c` implementation. Better to make the simd code compliant.

ePirat:
@tianjunwork I can do that, yes

1480c1:
@ePirat any updates?

ePirat:
I will try to finish my WIP on this tomorrow. (Removing the skip parameter)

1480c1:
@ePirat can you remove the reverted commit if it's not necessary?

@hassount how soon can we review this? It would be nice to have this done since it's also one of the cppcheck issues

hassount:
@OmranAbdelkafi has this PR been tested by any chance ? 
@joelsole is this good to go after the explanation above  https://github.com/OpenVisualCloud/SVT-AV1/pull/1287#issuecomment-629145866

ePirat:
@1480c1 Oh sorry, forgot to squash the revert… Done.

1480c1:
@joelsole 🏓

PR #1288, opened 2020-05-07
EbAppMain, Ebthreads: use lowercase windows.h for case sensitive systems

1480c1:
On nix systems with mingw-w64, gcc will error out on `Windows.h` since they do not do case insensitive include search


tianjunwork:
Hi @joelsole @kylophone is the PR good to go?

PR #1294, opened 2020-05-11
Optimize performance reduce round() on double to int

spawlows:
Speedup encoder, remove round() on double.

For cmd line below:
```
-i 1920x1080_60fps_420_8bit.yuv -enc-mode [mode] -w 1920 -h 1080 -bit-depth 8 -fps 60 -b out.ivf
```

Encoding speedup:
 Mode | master[fps] | Changes[fps] | speedup[%]
-- | -- | --| --
M8 | 18,953 | 19,659|3.73
M7 | 11,672 | 11,986|2.69
M6 | 10,454 | 10,761|2.94
M5 | 5,615 | 5,669|0.96
M4 | 3,745 | 3,786|1.09
M3 | 2,483 | 2,511|1.13
M2 | 0,32 | 0,321|0.31
M1 | 0,051 | 0,052|1.96
M0 | 0,037 | 0,037|0.00

### DIFF
--- a/Source/Lib/Encoder/Codec/EbProductCodingLoop.c
+++ b/Source/Lib/Encoder/Codec/EbProductCodingLoop.c
@@ -31,6 +31,8 @@
 #include "EbCommonUtils.h"
 #include "EbResize.h"
 
+#define DIVIDE_AND_ROUND(x, y) (((x) + ((y) >> 1)) / (y))
+
 EbErrorType generate_md_stage_0_cand(SuperBlock *sb_ptr, ModeDecisionContext *context_ptr,
                                      uint32_t *         fast_candidate_total_count,
                                      PictureControlSet *pcs_ptr);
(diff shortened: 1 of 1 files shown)


tianjunwork:
Thanks for this optimization! Surprised to see data type casting bring that much speed penalty. Curious, is that because `float` is used? or just casting return value from double to integer.

spawlows:
> Thanks for this optimization! Surprised to see data type casting bring that much speed penalty. Curious, is that because `float` is used? or just casting return value from double to integer.

Float is one of thing, but this change remove calls function round(), just doing some trick on integers, and one division enough to get that same result.

EricLamVCD:
Obj-1 8bit and1 10bit clips: 
  | Speed Deviation  | BD-Rate   Deviation
-- | -- | --
M0 - 8bit | 0.80% | 0.00%
M4 - 8bit | 0.70% | 0.00%
M8 - 8bit | 2.40% | 0.00%
M0 - 10bit | 0.52% | 0.00%
M4 - 10bit | -0.15% | 0.00%
M8 - 10bit | 1.21% | 0.00%

PR #1296, opened 2020-05-12
Remove the macro for 31B

lijing0010:
The macro "MRP_31B_SUPPORT" is removed in EbDefinition.h (enabled by default), so remove it here.

When I submit the previous PR #1265 (31B hang related) for review, the macro is there.
When merging the PR #1265, the macro is removed by some other PR , so it will still hang for 31B


tianjunwork:
Hi @OmranAbdelkafi , the fix is already validated in #1265.  This one just needs a quick sanity check.

PR #1298, opened 2020-05-13
Travis CI: Move to GCC 10

EwoutH:
Yesterday GCC 10.1 was [migrated](https://tracker.debian.org/news/1143538/gcc-10-1010-1-migrated-to-testing/) to [Debian Testing](https://tracker.debian.org/pkg/gcc-10), which makes updating the CI fairly simple.

Use GCC 10 with the build jobs on Arm64, PowerPC and IBM Z, as well as with the Valgrind job.

The PowerPC jobs produces some new notes (light warnings) with GCC 10, other than that all runs pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,7 +104,7 @@ jobs:
       arch: s390x
     # Valgrind on Linux
     - name: Valgrind
-      env: build_type=debug
+      env: build_type=debug CC=gcc-10 CXX=g++-10
       addons:
         apt:
           update: true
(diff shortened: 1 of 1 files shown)


1480c1:
@EwoutH you might as well move the CC and CXX variable to the global envs

EwoutH:
Good call, done!

tianjunwork:
Moving it to global caused error in clang.

1480c1:
I guess CC=clang and CXX=clang++ needs to be added to the macOS jobs, however, I am currently working on removing them from travis-ci

EwoutH:
I reset the branch to 5f9c009, it should be ready to merge. Chris can then move it to a global variable once the macOS jobs are migrated.

PR #1299, opened 2020-05-14
ConfigureGitVersion: Remove --broken

1480c1:
git describe only supports `--broken` if git's version is greater than 2.13.7, else it will print out
```none
  Failure to get version from Git: error: unknown option `broken'
  usage: git describe [<options>] [<commit-ish>...]
     or: git describe [<options>] --dirty
...
<rest of git describe -h>
```

### DIFF
--- a/Source/Lib/Common/Codec/ConfigureGitVersion.cmake
+++ b/Source/Lib/Common/Codec/ConfigureGitVersion.cmake
@@ -16,11 +16,11 @@
 set(PACKAGE_VERSION_STRING "v${PACKAGE_VERSION_STRING}")
 
 find_package(Git QUIET)
-if((Git_FOUND) AND (EXISTS "${GIT_ROOT_DIR}/.git"))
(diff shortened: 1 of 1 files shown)


tianjunwork:
Thanks Chris for the quick fix. I am on Ubuntu 16.04 with git version 2.7.4 which doesn't have `--broken` option. So it will error out. #1220 fails to work.
Is  `--broken` necessary for all version?

ePirat:
We can probably get rid of --broken completely, it's not terribly important to have it.

tianjunwork:
Thanks @ePirat. 
@1480c1 let's remove it and simply the code.

1480c1:
> Thanks @ePirat.
> @1480c1 let's remove it and simply the code.

Done

tianjunwork:
Hi @joelsole @OmranAbdelkafi , removing `--broken` fix my error on Ubuntu 16.04 with git version 2.7.4. Please review. Thanks.

PR #1300, opened 2020-05-14
Actions: Add releases job; Remove from appveyor

1480c1:
actions doesn't yet support glob uploading release assets

Removed the MSVC release build and uploading of artifacts for Releases
from appveyor

Related to #873

@ePirat can you help to review this?

Currently will look like
image


ePirat:
Maybe the artifacts should be zipped? Or does the exes works fine without the .dll's?

ePirat:
Overall looks good to me, thanks a lot for this!

EwoutH:
Looks good!

@ePirat Right now the libraries are linked dynamically, so the dll's are needed.

What do you guys think of adding the version number in the release string? So `SvtAv1Dec-0.8.5.dll` etc. An advantage would be that it's very clear for user, but maybe a bit less practical for CI / machine applications.

It would also be very nice to add a .dmg for macOS on release.

ePirat:
I don't think you should do that, instead create a zip with the dll end exe and name that correctly. If anything, the dll should have the soversion in the name.

ePirat:
(not that it currently matters much as the ABI is anyway not stable right now, but at least once it is, it would be quite confusing to have different dll names even though the ABI is compatible)

EwoutH:
Good point, agreed.

How stable are the FFmpeg builds currently? It would be amazing if we can publish an FFmpeg build with SVT-AV1 on each release. Would there be any issues with licencing in that case?

ePirat:
The ffmpeg patch is a bit outdated it seems. While it still builds fine, some of the arguments to not map correctly (anymore?)

ePirat:
The rc option for example:

https://github.com/OpenVisualCloud/SVT-AV1/blob/7bdab42e3441620ef28abfb6269783590ea6c182/ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch#L579-L583

uses the values 0, 2, 3, but 2 and 3 are not even documented in the SVT-AV1 code and last time I tried it would just assert when using 3…

tianjunwork:
Hi @EwoutH ffmpeg plugin will be worked on/improved before the official release. E.g. clean up the options, fix conflicts. Now is not a good time to add in release build. But it is a good idea!

1480c1:
> How stable are the FFmpeg builds currently? It would be amazing if we can publish an FFmpeg build with SVT-AV1 on each release. Would there be any issues with licencing in that case?

https://ci.appveyor.com/project/tianjunwork/svt-av1/builds/32857368/job/nkyb8qjrdc1pg8lj/artifacts
We currently upload an FFmpeg artifact built using mingw-w64 for windows, but it's not actually published to releases and is supposed to be for "testing" rather than a full release

1480c1:
@ePirat @EwoutH @tianjunwork @hassount it seems that someone is requesting static lib files instead of the dynamic dll files that are currently being produced and uploaded. Do you guys have any thoughts?

I am fine with using static libs instead since I prefer using static libs and binaries since, based on my personal experience, it's easier to tell a windows user to just download the exe in comparison to telling them to download two files.

The only issue that I see is that some people like to be able to just download the dll to upgrade their current dll file in-place, which although I understand is a good use of using dlls, I'm not sure is such a good idea right now considering the exe often gets updates and bugfixes too

Also, @ePirat, based on asking some people who use the release binaries (Really small sample though), they prefer to be able to download the individual files instead of needing to download a zip and unzipping it

ePirat:
I really think zipping up the related files is better, else it clutters the download list quite a bit… And I don't think it's too much to ask to run unzip on the downloaded file…

Can you just build shared and static and include both in the release? (Make two zip's one for the shared and one for the static ones?)

1480c1:
> I really think zipping up the related files is better, else it clutters the download list quite a bit… And I don't think it's too much to ask to run unzip on the downloaded file…
> 
> Can you just build shared and static and include both in the release? (Make two zip's one for the shared and one for the static ones?)

Sure, we could do that, maybe something like `win64-static.zip` and `win64-shared.zip`? Should we separate the encoder and decoder so we have something like `win64-svtav1enc-static.zip`?

oviano:
Static libs would be my preference, and also consistent with what you are doing with SVT-HEVC.

If you stick with DLLs, then at least include the import libs too, which it looks like you did in the previous (before 0.8.2) release, but it seems it would be simplest just to provide a statically build EXE and a static LIB along with the source.

ePirat:
Oh indeed the import lib should be included as well, else the dlls are a bit useless…

1480c1:
This would have been a good case for having the build output being configurable

tianjunwork:
If it is not too much trouble, how about we release `static lib` + `standalone exe` + `shared library zip(exe+dll+dll's import lib)` for encoder, `standalone exe` for decoder?

1480c1:
> If it is not too much trouble, how about we release `static lib` + `standalone exe` + `shared library zip(exe+dll+dll's import lib)` for encoder, `standalone exe` for decoder?

We could try that, good thing we added a `BUILD_APPS` and `BUILD_ENC` and `BUILD_DEC` options earlier on

1480c1:
image
Currently, a release will look like this, with the zip file looking like 
image

@EwoutH @ePirat @oviano @hassount Any comments or concerns or suggestions?

oviano:
> image
> Currently, a release will look like this, with the zip file looking like
> image
> 
> @EwoutH @ePirat @oviano @hassount Any comments or concerns or suggestions?

Looks great - but presumably the source code would be there too, so that developers who want to link the libraries can get easy access to the required header files too?

1480c1:
> Looks great - but presumably the source code would be there too, so that developers who want to link the libraries can get easy access to the required header files too?

It would be available through the auto-generated tar.gz and zip file from GitHub

oviano:
> > Looks great - but presumably the source code would be there too, so that developers who want to link the libraries can get easy access to the required header files too?
> 
> It would be available through the auto-generated tar.gz and zip file from GitHub

Right, so that's done automatically.

It seems like you have covered all bases, looks great to me.

1480c1:
image
Here's how it would look in the end

1480c1:
@EwoutH @ePirat and @luigi311 can you guys make sure there are no problems with this right now? Anything I am forgetting to add or remove?

1480c1:
image
Seems I need to redo this one, it does not seem that this will work if using github's UI to make a release, only really suitable with `git tag`

.github/workflows/releases.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: 7z.exe a SvtAv1Enc-shared.zip ./shared/Release/SvtAv1Enc.dll ./shared/Release/SvtAv1Enc.lib ./shared/Release/SvtAv1EncApp.exe
+
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1.0.1
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: SVT-AV1 ${{ github.ref }}
+          draft: true
+          prerelease: true
```
luigi311, line 37:
```suggestion
```
Not needed if tags are created manually via github ui.

.github/workflows/releases.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actions/upload-release-asset@v1.0.2
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: static/Release/SvtAv1EncApp.exe
+          asset_name: SvtAv1EncApp.exe
+          asset_content_type: application/vnd.microsoft.portable-executable
+      - name: Upload static SvtAv1DecApp.exe
+        uses: actions/upload-release-asset@v1.0.2
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
```
luigi311, line 49: Since the create release is happening outside of the this process we might need to manually recreate the link to the latest release. You can not use /releases/tag/latest do to us only doing pre-releases now so there is no latest and that also wont work reliably in the future do to the possibility of create pre-releases. It can be manually recreated though by setting it equal to 
```
https://github.com/OpenVisualCloud/SVT-AV1/releases/tag/${{ github.ref }}
```
or recreating it more dynamically
```
https://github.com/${{ github.repository }}/releases/tags/${{ github.ref }}
```
to allow it to work on forked repositories which i think is the better solution

oviano:
I am just wondering if you were planning to re-release 0.8.4 with these additional files, or do I need to wait for 0.8.5 to get the static library or build it myself?

1480c1:
Mainly depends on if we get this in before 0.8.4

tianjunwork:
Hi @oviano , you mean re-release the most recent tag 0.8.3, right? Sure, we will re-upload the pre-built as what we discussed here in https://github.com/OpenVisualCloud/SVT-AV1/releases/tag/v0.8.3

oviano:
> Hi @oviano , you mean re-release the most recent tag 0.8.3, right? Sure, we will re-upload the pre-built as what we discussed here in https://github.com/OpenVisualCloud/SVT-AV1/releases/tag/v0.8.3

That's great (yes I meant v0.8.3, sorry!)

.github/workflows/releases.yml
1480c1: Done

.github/workflows/releases.yml
1480c1: The `upload_url` is more like `https://uploads.github.com/repos/1480c1/SVT-AV1/releases/<some numbers>/assets{?name,label}` so using a hardcoded URL would not work

luigi311: Ahh that's good to know i will have to try and remember this if i ever need to accomplish something similar.

tianjunwork:
Hi all reviewers, this PR is ready to review.
With this PR, a release job will be triggered when `publish a release` on **Github UI** (using `git tag` is not supported).
The assets are https://github.com/OpenVisualCloud/SVT-AV1/pull/1300#issuecomment-628855867 based on feedback from community.

1480c1:
If one of the releasers do use `git tag`, do mention it, I can add a functionality to create a draft release for it.

PR #1306, opened 2020-05-17
build.bat: Add static and shared options

1480c1:
Whichever one comes last will be the one that is built, Should help with people who want to choose their build type.

### DIFF
--- a/Build/windows/build.bat
+++ b/Build/windows/build.bat
@@ -11,11 +11,10 @@ if NOT -%1-==-- call :args %*
 if exist CMakeCache.txt del /f /s /q CMakeCache.txt 1>nul
 if exist CMakeFiles rmdir /s /q CMakeFiles 1>nul
 if NOT "%GENERATOR%"=="" set GENERATOR=-G"%GENERATOR%"
-if NOT "%unittest%"=="" set "cmake_eflags=%cmake_eflags% -DBUILD_TESTING=ON"
 if "%vs%"=="2019" (
-    cmake ../.. %GENERATOR% -A x64 -DCMAKE_INSTALL_PREFIX=%SYSTEMDRIVE%\svt-encoders -DCMAKE_CONFIGURATION_TYPES="Debug;Release" %cmake_eflags%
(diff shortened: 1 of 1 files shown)


Build/windows/build.bat
```diff
@@ -119,6 +126,6 @@ goto :args
 
 :help
     echo Batch file to build SVT-AV1 on Windows
-    echo Usage: generate.bat 2019^|2017^|2015^|clean [release|debug] [nobuild] [test]
+    echo Usage: generate.bat 2019^|2017^|2015^|clean [release|debug] [nobuild] [test] [shared|static]
```
tianjunwork, line 129: generate.bat ==> build.bat
2019^|2017^|2015^|clean ==>[2019^|2017^|2015^|clean]

1480c1: Seems I forgot to escape some stuff

ePirat:
Does it fix the buildtype as well? Last time I tried to use that it just errored out with a confusing error so I had to use cmake directly.

1480c1:
It at least does for me, although it doesn't properly do it with non msvc generators since that requires setting `-DCMAKE_BUILD_TYPE=<type>`, but it wouldn't be good to add that without checking the generator, but bat doesn't have a good way to do case switches

Build/windows/build.bat
```diff
@@ -4,24 +4,41 @@
...
 setlocal
 cd /d "%~dp0"
-set instdir=%CD%
 
+:: Set defaults to prevent inheriting
 set "build=y"
+:: Default is debug
+set "buildtype=Debug"
+:: Default is shared
+set "static=ON"
```
tianjunwork, line 13: `static=on` means building shared? not easy to read.

1480c1: Should I change this to `shared` instead?

1480c1: Changed it

tianjunwork:
Hi @ePirat , could you also review and approve? Thank you.

PR #1307, opened 2020-05-18
EB_API: Use a def file instead of __declspec(dllexport)

1480c1:
Both results in the same usable output, however, with the use of `__declspec(dllexport)` without having proper guards when compiling the same headers in the executable, CMake and by extension Visual Studio will generate a .lib file for SvtAv1EncApp and SvtAv1DecApp, which is not wanted.

The alternative is to trace the headers and add extra guards to prevent `__declspec(dllexport)` from appearing when compiling the executables

The def file is generated by running `gendef - SvtAv1Enc.dll > SvtAv1Enc.def` using gendef from mingw-w64 gotten through msys2. To add to the def file, just add the function name to the def file as long as the function name is not mangled like in C++.

https://github.com/Alexpux/mingw-w64/tree/master/mingw-w64-tools/gendef
https://docs.microsoft.com/en-us/cpp/build/reference/module-definition-dot-def-files?view=vs-2019
https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files?view=vs-2019
https://docs.microsoft.com/en-us/cpp/build/reference/exports?view=vs-2019
https://docs.microsoft.com/en-us/cpp/cpp/dllexport-dllimport?view=vs-2019
https://docs.microsoft.com/en-us/cpp/build/reference/export-exports-a-function?view=vs-2019

I can alternatively keep the current `EB_API` in the code but define it to nothing so that at least the functions are still easy to spot


ePirat:
Note that we need some sort of API exports file for Linux/macOS as well so maybe we should choose an approach that works across platforms so we dont end up with several files to maintain?

1480c1:
I believe we used to have something like a version-script file in this repo, and we still do in SVT-VP9 at least

1480c1:
Reading https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/BuildingWinDLL
It seems that we might be able to use `SvtAv1Enc_EXPORTS` and `SvtAv1Dec_EXPORTS` to determine when to use `EB_API` and maybe we can properly look at doing exported functions for both windows and *nix

ePirat:
Yes, we probably should use that generated header and annotations, we probably need a symbol list too, as for ASM symbols those will always be visible without a symbol export list used at link-time.

1480c1:
> `SvtAv1Enc_EXPORTS` and `SvtAv1Dec_EXPORTS`

Seems we can't use these defines since cmake only defines them for the source files that are listed directly under add_library, but we use object files, which do not reverse inherit defines and flags

1480c1:
I will try using the `GenerateExportHeader` thing in cmake now

ePirat:
> Seems we can't use these defines since cmake only defines them for the source files that are listed directly under add_library, but we use object files, which do not reverse inherit defines and flags

Uh what exactly do you mean by that? IIRC how it works is that CMake generates a header with the export macros which we can just include where needed?

1480c1:
```cmake
project(test)

add_library(testlib SHARED test.c)
```

CMake will add `testlib_EXPORTS` define when compiling test.c when compiling for MSVC shared dll libraries, but in the case where we have something like

```cmake
project(test)

add_library(objlib OBJECT test.c)

add_library(testlib SHARED dummy.c $<TARGET_OBJECTS:objlib>)
```
CMake will not add `testlib_EXPORTS` when compiling test.c, but will add it when compiling dummy.c, so we cannot use `testlib_EXPORTS` when compiling test.c, which might have the `__declspec(dllexport)` declaration inside it

tianjunwork:
Hi @1480c1 , Thanks for the PR which eliminates generating `SvtAv1EncApp.lib` when static targets are built on Windows. Now I agree with ePirat that it is better to have a solution what works cross-platform. Otherwise, it is easy for developer to forget updating def file when APIs are updated.
I would suggest close this PR. Maybe revisit when it is needed.
Thanks! It is a good experiment!

Issue #1309, opened 2020-05-20
Encoder crashes with enc_mode = 0 and rate_control_mode = 1


tkwf:
Command line:
-i input.yuv420 -w 1280 -h 720 -fps 30 -rc 1 -enc-mode 0 -n 10 -b output.ivf

PR #1310, opened 2020-05-20
Fix RC crash

OmranAbdelkafi:
## Description : 
RC only works with sb 64x64
## Related issue : 
Closes issue #1309 
## Performance : 
No impact


ccccheung:
I was able to reproduce issue #1309 on Windows and Linux using enc-modes 0-3. 

Using this PR, the issue no longer occurs.

PR #1340, opened 2020-06-16
replace memcpy by eb_memcpy for the library

ttrigui:
# Description
unifing all memcpy to use eb_memcpy for the library

# Author(s)
@ttrigui 

# Performance impact
- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [x] other
- [ ] N/A

# Merge method


1480c1:
Are all of the void pointer casts necessary? Does gcc or clang give any warnings without them?

ttrigui:
> Are all of the void pointer casts necessary? Does gcc or clang give any warnings without them?

VS on debug mode gives warnings.

1480c1:
I see

silverbacknet:
I feel like a better option would be to silence C4090 (I assume that's the warning) at eb_memcpy instead of spewing pointer casts all over the codebase.

ttrigui:
> C4090

yes it is C4090.
@hassount is that an option?

1480c1:
Can you instead try setting the second argument of eb_memcpy to `const void *` to match regular memcpy?

Source/Lib/Common/ASM_AVX512/convolve_2d_avx512.c
```diff
@@ -1120,20 +1120,20 @@ void eb_av1_convolve_2d_copy_sr_avx512(const uint8_t *src, int32_t src_stride, u
 
     if (w == 2) {
```
xuguangxin, line 1121: since we already did memcpy size check  in eb_memcpy, why we still need `if (w == x )` logical here?

Source/Lib/Common/ASM_SSE4_1/filterintra_sse4.c
```diff
@@ -42,7 +42,7 @@ void eb_av1_filter_intra_predictor_sse4_1(uint8_t *dst, ptrdiff_t stride, TxSize
     for (r = 1; r < bh + 1; r += 2) {
         for (c = 1; c < bw + 1; c += 4) {
             DECLARE_ALIGNED(16, uint8_t, p[8]);
-            memcpy(p, &buffer[r - 1][c - 1], 5 * sizeof(uint8_t));
+            eb_memcpy((void*)p, (void*)&buffer[r - 1][c - 1], 5 * sizeof(uint8_t));
```
xuguangxin, line 45: converting to void* is not needed, all pointers compatible with void*
see https://stackoverflow.com/questions/6534222/convert-into-void/6534258#6534258 for details

hassount:
> > C4090
> 
> yes it is C4090.
> @hassount is that an option?

Unless it is absolutely necessary. Preferably, we would find a way to address the warnings.

ttrigui:
> Can you instead try setting the second argument of eb_memcpy to `const void *` to match regular memcpy?

That should work. Working on changing that.

hassount:
@ttrigui can you please rebase the PR so that we can make sure the ffmpeg build is now passing ?

Source/Lib/Common/ASM_AVX512/convolve_2d_avx512.c
ttrigui: we can combine w == 2 and w ==4 but the rest have to stay as we are using different instructions

1480c1:
@ttrigui, can you please rebase? I am currently trying to fix more cppcheck issues, but since it might change some of the code you have in this branch, I was going to rebase my local branch on this branch


@hassount Is there anything else that would need to be done for this PR after rebasing?

hassount:
we can merge it as long as @OmranAbdelkafi give us a green light from a validation point of view

1480c1:
Btw @ttrigui, have you looked at the `eb_memcpy_app` inside `EbUtility.c` that has the same exact functions and everything?

Source/Lib/Common/ASM_AVX512/convolve_2d_avx512.c
xuguangxin: Is it generic enough? can we put them in eb_memcpy?
thanks

EricLamVCD:
Tested obj-1 8bit and 10bit clips: 
  | Speed Deviation  | BD-Rate   Deviation | Memory   Deviation
-- | -- | -- | --
M0 - 8bit | -3.84% | 0.04% | 0.02%
M4 - 8bit | -4.74% | 0.00% | 0.00%
M8 - 8bit | -3.80% | 0.00% | 0.03%
M0 - 10bit | 0.95% | 0.00% | 0.00%
M4 - 10bit | 1.02% | 0.00% | 0.01%
M8 - 10bit | 1.26% | 0.00% | -0.02%

The results were lossless.

Issue #1172, opened 2020-03-30
Crash on SIGINT


ePirat:
When terminating an encode with Ctrl-C (SIGINT) in a debug build with address sanitizer it crashes:

```
$ ./Bin/Release/SvtAv1EncApp -asm 0 -i ~/fireworks_3840x2160.yuv -n 450 -w 3840 -h 2160 -b ~/out.ivf                                                                 130 ↵
-------------------------------------------
SVT-AV1 Encoder
SVT [version]:	SVT-AV1 Encoder Lib v0.8.1
SVT [build]  :	GCC 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.29)	 64 bit
LIB Build date: Mar 30 2020 14:01:35
-------------------------------------------
Number of logical cores available: 4
Number of PPCS 72
[asm level on system : up to avx2]
[asm level selected : up to c]
------------------------------------------- 
SVT [config]: Main Profile	Tier (auto)	Level (auto)	
SVT [config]: EncoderMode 							: 8 
SVT [config]: EncoderBitDepth / EncoderColorFormat / CompressedTenBitFormat	: 8 / 1 / 0
SVT [config]: SourceWidth / SourceHeight					: 3840 / 2160 
SVT [config]: FrameRate / Gop Size						: 30 / 32 
SVT [config]: HierarchicalLevels  / PredStructure		: 4 / 2
SVT [config]: BRC Mode / QP  / LookaheadDistance / SceneChange			: CQP / 50 / 33 / 0 
------------------------------------------- 
warning: -asm will be deprecated soon, please use --asm
Encoding          ^CAddressSanitizer:DEADLYSIGNAL
=================================================================
==4237==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000028 (pc 0x00010a9877a2 bp 0x7ffee54adc30 sp 0x7ffee54adc00 T0)
==4237==The signal is caused by a READ memory access.
==4237==Hint: address points to the zero page.
    #0 0x10a9877a1 in eb_get_empty_object EbSystemResourceManager.c:598
    #1 0x10ae86d63 in eb_svt_enc_send_picture EbEncHandle.c:3472
    #2 0x10a76bff6 in process_input_buffer EbAppProcessCmd.c:952
    #3 0x10a767230 in main EbAppMain.c:212
    #4 0x7fff72ebe7fc in start (libdyld.dylib:x86_64+0x1a7fc)

[...]

PR #1358, opened 2020-06-30
fix crash on Ctrl+C [WIP]

xuguangxin:
# Description

sem_wait will interrupted by singals, like Ctrl-C.
In this case, eb_get_full_object/eb_get_empty_object wakeup with zero object in queue.
It will crash the application

    Test steps:
    1. run ./ffmpeg -i input.264 test.ivf -y
    2. wait sometime to make sure encode is started.
    3. press Ctrl+C

    Expected:
    No crash

# Issue

# Author(s)
Guangxin.Xu@intel.com

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/Source/Lib/Common/Codec/EbThreads.c
+++ b/Source/Lib/Common/Codec/EbThreads.c
@@ -244,8 +244,11 @@ EbErrorType eb_block_on_semaphore(EbHandle semaphore_handle)
             ? EB_ErrorSemaphoreUnresponsive
             : EB_ErrorNone;
 #else
-    return_error =
-        sem_wait((sem_t *)semaphore_handle) ? EB_ErrorSemaphoreUnresponsive : EB_ErrorNone;
+    int ret;
+    do {
+        ret = sem_wait((sem_t *)semaphore_handle);
+    } while(ret == -1 && errno == EINTR);
+    return_error = ret ? EB_ErrorSemaphoreUnresponsive : EB_ErrorNone;
 #endif
 
     return return_error;


xuguangxin:
@1480c1 is any way to simulate mac build env on linux?
The Pr failed on mac, but I do not have mac build env on hand

thanks

1480c1:
> @1480c1 is any way to simulate mac build env on linux?
> The Pr failed on mac, but I do not have mac build env on hand
> 
> thanks

afaik, you will need to run macOS in a virtual machine, which depending on how you do it, it will be a pain and has some legal gray areas most of the time.

in no specific ordering,

https://github.com/kholia/OSX-KVM
https://github.com/foxlet/macOS-Simple-KVM
https://www.howtogeek.com/289594/how-to-install-macos-sierra-in-virtualbox-on-windows-10/ (not necessarily windows 10, but parts of the Virtualbox might help if you are using that)


essentially, you may need to run a virtual hackintosh.


@ePirat do you know an easy way to debug those issues without having a macOS?

Source/Lib/Common/Codec/EbThreads.c
```diff
@@ -31,12 +31,16 @@
...
 #include <fcntl.h>
 #include <pthread.h>
-#include <semaphore.h>
 #include <unistd.h>
-#endif // _WIN32
 #ifdef __APPLE__
 #include <dispatch/dispatch.h>
-#endif
+typedef dispatch_semaphore_t sem_t;
+INLINE int sem_wait(sem_t* psem) { return (int)dispatch_semaphore_wait(*psem, DISPATCH_TIME_FOREVER); }
```
1480c1, line 38:
```suggestion
static inline int sem_wait(sem_t* psem) { return (int)dispatch_semaphore_wait(*psem, DISPATCH_TIME_FOREVER); }
```

static inline might be better to reduce the amount of generic global function names we might have.

xuguangxin: @1480c1 , thanks for suggestion. 
I will change it later.

xuguangxin:
@1480c1 , hackintosh is not allowed. 
Let me find a mac to fix this. 
thanks

xuguangxin:
@1480c1 please help review this again.
also could you help remove macos label on this?
It fix issues for linux alike system includes mac
thanks

Source/Lib/Common/Codec/EbThreads.c
```diff
@@ -31,12 +31,16 @@
...
 #include <fcntl.h>
 #include <pthread.h>
-#include <semaphore.h>
 #include <unistd.h>
-#endif // _WIN32
 #ifdef __APPLE__
 #include <dispatch/dispatch.h>
-#endif
+typedef void sem_t;
+static INLINE int sem_wait(sem_t* psem) { return (int)dispatch_semaphore_wait((dispatch_semaphore_t)psem, DISPATCH_TIME_FOREVER); }
```
1480c1, line 38:
```suggestion
static inline int sem_wait(sem_t* psem) { return (int)dispatch_semaphore_wait((dispatch_semaphore_t)psem, DISPATCH_TIME_FOREVER); }
```

since we are using c99, `inline` is properly handled, so no need to use the defined `INLINE` for the inline extension

1480c1:
For the actual macOS stuff, you may want to ping @kylophone or @ePirat since they are the main ones that I know of that use macOS and knows how to develop for it, and @ePirat did do the dispatch semaphore part.

Source/Lib/Common/Codec/EbThreads.c
ePirat: `inline` should be fine here, it is not in cross-platform code though as MSVC does not or only recently learned about inline. That said, in general it would be better to consistently use `inline` everywhere and just `#define` that properly when missing for some MSVC versions.

ePirat:
In general I am not that happy with this change as it makes the code quite a bit more confusing to read.
It seems a previous message got lost so I will just write it again:

Additionally the change for macOS seems useless as with the usage of a dispatch semaphore, signal interruption should be no issue.

xuguangxin:
@ePirat , the msg may dropped by github. thank  for write it again.
The apple's document is little confuse to me. not sure we need handle interruption or not.
https://developer.apple.com/documentation/dispatch/1453087-dispatch_semaphore_wait

`Decrement the counting semaphore. If the resulting value is less than zero, this function waits for a signal to occur before returning.`

ePirat:
Hi, I can not reproduce it on macOS anymore so I guess changing to dispatch semaphores accidentally fixed it and I did not notice.
Regarding the documentation, what they mean by signal is probably signaling the semaphore with `dispatch_semaphore_signal`, not a "normal" signal.

xuguangxin:
Hi @ePirat ,
thanks for check. Change this to linux only.
If we can't reproduce https://github.com/OpenVisualCloud/SVT-AV1/issues/1172, could you help close it?

hassount:
@xuguangxin @ePirat  what is the status of this ? could we merge it or do we still need to address the macos case ?

xuguangxin:
@hassount it's ready to merge. 
see
https://github.com/OpenVisualCloud/SVT-AV1/issues/1172#issuecomment-659965714

ePirat:
@hassount This is fine, macOS should already have been fixed accidentally by my patch to use of dispatch semaphores.

PR #998, opened 2020-02-01
Add combined test mode for two pass

jsunintel:

PR #1365, opened 2020-07-06
SvtAv1EncApp : combined 2 passes encode

xuguangxin:
# Description
SvtAv1EncApp : enable x264 style 2 pass encoder and two passes encode.
Now you can use following command line to encode 2 pass bitstream

    SvtAv1EncApp --pass 1   -i <input> -b <output>
    SvtAv1EncApp --pass 2  -i <input> -b <output>

    SvtAv1EncApp --passes 2   -i <input> -b <output>

    ffmpeg -i input.y4m -c:v libsvt_av1 -pass 1 pass1.ivf -y
    ffmpeg -i input.y4m -c:v libsvt_av1 -pass 2 pass2.ivf -y

# Issue

No

# Author(s)
jing.a.sun@intel.com
guangxin.xu@intel.com

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -17,6 +17,7 @@
 #include <io.h>
 #else
 #include <unistd.h>
+#include <sys/file.h>
 #endif
 
 /**********************************
@@ -251,6 +252,50 @@
 #if 1//ON_OFF_FEATURE_MRP
 #define MRP_LEVEL_TOKEN "--mrp-level"
 #endif
+
+#ifdef _WIN32
+static HANDLE get_file_handle(FILE* fp)
+{
+    return (HANDLE)_get_osfhandle(_fileno(fp));
+}
+#endif
+
+static EbBool fopen_and_lock(FILE** file, const char* name, EbBool write)
+{
+    if (!file || !name)
+        return EB_FALSE;
+
+    const char* mode = write ? "wb" : "rb";
+    FOPEN(*file, name, mode);
+    if (!*file)
+        return EB_FALSE;
+
+#ifdef _WIN32
+    HANDLE handle = get_file_handle(*file);
+    if (handle == INVALID_HANDLE_VALUE) return EB_FALSE;
+    if (LockFile(handle, 0, 0, MAXDWORD, MAXDWORD))
+        return EB_TRUE;
+#else
+    int fd = fileno(*file);
+    if (flock(fd, LOCK_EX | LOCK_NB) == 0)
+        return EB_TRUE;
+#endif
+    fprintf(stderr, "ERROR: locking %s failed, is it used by other encoder?\n", name);
+    return EB_FALSE;
+}
+
+static void unlock_and_fclose(FILE* file)
+{
+    if (!file)
+        return;
+
+#ifdef _WIN32
+    HANDLE handle = get_file_handle(file);
+    UnlockFile(handle, 0, 0, MAXDWORD, MAXDWORD);
+#endif
+    fclose(file);
+}
+
 /**********************************
  * Set Cfg Functions
  **********************************/
@@ -334,12 +379,12 @@ static void set_passes(const char* value, EbConfig *cfg) {
 }
 
 static void set_input_stat_file(const char *value, EbConfig *cfg) {
-    if (cfg->input_stat_file) { fclose(cfg->input_stat_file); }
-    FOPEN(cfg->input_stat_file, value, "rb");
+    if (cfg->input_stat_file) { unlock_and_fclose(cfg->input_stat_file); }
+    fopen_and_lock(&cfg->input_stat_file, value, EB_FALSE);
 };
 static void set_output_stat_file(const char *value, EbConfig *cfg) {
-    if (cfg->output_stat_file) { fclose(cfg->output_stat_file); }
-    FOPEN(cfg->output_stat_file, value, "wb");
+    if (cfg->output_stat_file) { unlock_and_fclose(cfg->output_stat_file); }
+    fopen_and_lock(&cfg->output_stat_file, value, EB_TRUE);
 };
 static void set_snd_pass_enc_mode(const char *value, EbConfig *cfg) {
     cfg->snd_pass_enc_mode = (uint8_t)strtoul(value, NULL, 0);
@@ -1875,11 +1920,11 @@ void eb_config_dtor(EbConfig *config_ptr) {
         config_ptr->stat_file = (FILE *)NULL;
     }
     if (config_ptr->input_stat_file) {
-        fclose(config_ptr->input_stat_file);
+        unlock_and_fclose(config_ptr->input_stat_file);
         config_ptr->input_stat_file = (FILE *)NULL;
     }
     if (config_ptr->output_stat_file) {
-        fclose(config_ptr->output_stat_file);
+        unlock_and_fclose(config_ptr->output_stat_file);
         config_ptr->output_stat_file = (FILE *)NULL;
     }
     free((void*)config_ptr->stats);
@@ -2186,16 +2231,14 @@ static EbErrorType verify_settings(EbConfig *config, uint32_t channel_number) {
 
     const char* stats = config->stats ? config->stats : "svtav1_2pass.log";
     if (pass == 1) {
-        FOPEN(config->output_stat_file, stats, "wb");
-        if (!config->output_stat_file) {
+        if (!fopen_and_lock(&config->output_stat_file, stats, EB_TRUE)) {
             fprintf(config->error_log_file,
                 "Error instance %u: can't open stats file %s for write \n",
                 channel_number + 1, stats);
             return EB_ErrorBadParameter;
         }
     } else if (pass == 2) {
-        FOPEN(config->input_stat_file, stats, "rb");
-        if (!config->input_stat_file) {
+        if (!fopen_and_lock(&config->input_stat_file, stats, EB_FALSE)) {
             fprintf(config->error_log_file,
                 "Error instance %u: can't read stats file %s for read\n",
                 channel_number + 1, stats);
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -2506,6 +2506,19 @@ uint32_t get_passes(int32_t argc, char *const argv[], EncodePass pass[MAX_ENCODE
     }
     if (check_two_pass_conflicts(argc, argv))
         return 0;
+
+    int preset = MAX_ENC_PRESET;
+    if (find_token(argc, argv, PRESET_TOKEN, config_string) == 0
+        || find_token(argc, argv, ENCMODE_TOKEN, config_string) == 0) {
+        preset = strtol(config_string, NULL, 0);
+    }
+    if (preset > 4) {
+        fprintf(stderr,
+            "\nWarn: --passes 2 for preset > 4 is not supported yet, force single pass\n\n");
+        pass[0] = ENCODE_SINGLE_PASS;
+        return 1;
+    }
+
     pass[0] = ENCODE_FIRST_PASS;
     pass[1] = ENCODE_LAST_PASS;
     return 2;
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -33,8 +33,13 @@
 #define OUTPUT_RECON_TOKEN "-o"
 #define ERROR_FILE_TOKEN "-errlog"
 #define QP_FILE_TOKEN "-qp-file"
+
+/* two pass token */
+#define PASS_TOKEN "--pass"
+#define TWO_PASS_STATS_TOKEN "--stats"
 #define INPUT_STAT_FILE_TOKEN "-input-stat-file"
 #define OUTPUT_STAT_FILE_TOKEN "-output-stat-file"
+
 #define STAT_FILE_TOKEN "-stat-file"
 #define INPUT_PREDSTRUCT_FILE_TOKEN "-pred-struct-file"
 #define WIDTH_TOKEN "-w"
@@ -307,6 +312,19 @@ static void set_cfg_qp_file(const char *value, EbConfig *cfg) {
     if (cfg->qp_file) { fclose(cfg->qp_file); }
     FOPEN(cfg->qp_file, value, "r");
 };
+
+static void set_pass(const char* value, EbConfig *cfg) {
+    cfg->pass = strtol(value, NULL, 0);
+}
+
+static void set_two_pass_stats(const char* value, EbConfig *cfg) {
+#ifndef _WIN32
+    cfg->stats = strdup(value);
+#else
+    cfg->stats = _strdup(value);
+#endif
+}
+
 static void set_input_stat_file(const char *value, EbConfig *cfg) {
     if (cfg->input_stat_file) { fclose(cfg->input_stat_file); }
     FOPEN(cfg->input_stat_file, value, "rb");
@@ -900,6 +918,8 @@ ConfigEntry config_entry_rc[] = {
     {SINGLE_INPUT, NULL, NULL, NULL}};
 ConfigEntry config_entry_2p[] = {
     // 2 pass
+    {SINGLE_INPUT, PASS_TOKEN, "Multipass bitrate control (1: first pass, generates stats file , 2: second pass, uses stats file)", set_pass},
+    {SINGLE_INPUT, TWO_PASS_STATS_TOKEN, "Filename for 2 pass stats(\"svtav1_2pass.log\" : [Default])", set_two_pass_stats},
     {SINGLE_INPUT, OUTPUT_STAT_FILE_TOKEN, "First pass stat file output", set_output_stat_file},
     {SINGLE_INPUT,
      INPUT_STAT_FILE_TOKEN,
@@ -1307,8 +1327,13 @@ ConfigEntry config_entry[] = {
     {SINGLE_INPUT, OUTPUT_RECON_TOKEN, "ReconFile", set_cfg_recon_file},
     {SINGLE_INPUT, QP_FILE_TOKEN, "QpFile", set_cfg_qp_file},
     {SINGLE_INPUT, STAT_FILE_TOKEN, "StatFile", set_cfg_stat_file},
+
+    // two pass
+    {SINGLE_INPUT, PASS_TOKEN, "Pass", set_pass},
+    {SINGLE_INPUT, TWO_PASS_STATS_TOKEN, "Two pass stat", set_two_pass_stats},
     {SINGLE_INPUT, INPUT_STAT_FILE_TOKEN, "InputStatFile", set_input_stat_file},
     {SINGLE_INPUT, OUTPUT_STAT_FILE_TOKEN, "OutputStatFile", set_output_stat_file},
+
     {SINGLE_INPUT, INPUT_PREDSTRUCT_FILE_TOKEN, "PredStructFile", set_pred_struct_file},
     // Picture Dimensions
     {SINGLE_INPUT, WIDTH_TOKEN, "SourceWidth", set_cfg_source_width},
@@ -1786,6 +1811,8 @@ void eb_config_ctor(EbConfig *config_ptr) {
     config_ptr->md_stage_2_3_cand_prune_th  = 15;
     config_ptr->md_stage_2_3_class_prune_th = 25;
 
+    config_ptr->pass = DEFAULT;
+
     return;
 }
 
@@ -1846,6 +1873,7 @@ void eb_config_dtor(EbConfig *config_ptr) {
         fclose(config_ptr->output_stat_file);
         config_ptr->output_stat_file = (FILE *)NULL;
     }
+    free((void*)config_ptr->stats);
     return;
 }
 
@@ -2118,6 +2146,54 @@ static EbErrorType verify_settings(EbConfig *config, uint32_t channel_number) {
         return_error = EB_ErrorBadParameter;
     }
 
+    if (config->input_stat_file && config->output_stat_file) {
+        fprintf(config->error_log_file,
+                "Error instance %u: do not set input_stat_file and output_stat_file at same time\n",
+                channel_number + 1);
+        return EB_ErrorBadParameter;
+    }
+    int pass = config->pass;
+
+    if (pass != 2 && pass != 1 && pass != DEFAULT) {
+        fprintf(config->error_log_file,
+                "Error instance %u: %d pass encode is not supported\n",
+                channel_number + 1, config->pass);
+        return EB_ErrorBadParameter;
+    }
+
+    if (pass != DEFAULT  && (config->input_stat_file || config->output_stat_file)) {
+        fprintf(config->error_log_file,
+                "Error instance %u: --pass can't work with -input-stat-file or -output-stat-file \n",
+                channel_number + 1);
+        return EB_ErrorBadParameter;
+    }
+
+    if ((pass != DEFAULT || config->input_stat_file || config->output_stat_file) && channel_number > 0) {
+        fprintf(config->error_log_file,
+                "Error instance %u: 2 pass encode for multi instance is not supported\n",
+                channel_number + 1);
+        return EB_ErrorBadParameter;
+    }
+
+    const char* stats = config->stats ? config->stats : "svtav1_2pass.log";
+    if (pass == 1) {
+        FOPEN(config->output_stat_file, stats, "wb");
+        if (!config->output_stat_file) {
+            fprintf(config->error_log_file,
+                "Error instance %u: can't open stats file %s for write \n",
+                channel_number + 1, stats);
+            return EB_ErrorBadParameter;
+        }
+    } else if (pass == 2) {
+        FOPEN(config->input_stat_file, stats, "rb");
+        if (!config->input_stat_file) {
+            fprintf(config->error_log_file,
+                "Error instance %u: can't read stats file %s for read\n",
+                channel_number + 1, stats);
+            return EB_ErrorBadParameter;
+        }
+    }
+
     return return_error;
 }
 
@@ -2182,6 +2258,10 @@ uint32_t get_help(int32_t argc, char *const argv[]) {
         const char *empty_string         = "";
         fprintf(stderr,
                 "Usage: SvtAv1EncApp <options> -b dst_filename -i src_filename\n");
+        fprintf(stderr, "\n%-25s\n", "Examples:");
+        fprintf(stderr, "\n%-25s", "Two passes encode:");
+        fprintf(stderr, "\n\t%s", "SvtAv1EncApp --pass 1 -i <input> -b <output>");
+        fprintf(stderr, "\n\t%s\n", "SvtAv1EncApp --pass 2 -i <input> -b <output>");
         fprintf(stderr, "\n%-25s\n", "Options:");
         while (config_entry_options[++options_token_index].token != NULL) {
             uint32_t check = check_long(
@@ -2643,6 +2723,8 @@ const char *handle_warnings(const char *token, char *print_message, uint8_t doub
     if (EB_STRCMP(token, PAETH_TOKEN) == 0) linked_token = PAETH_NEW_TOKEN;
     if (EB_STRCMP(token, SMOOTH_TOKEN) == 0) linked_token = SMOOTH_NEW_TOKEN;
     if (EB_STRCMP(token, PIC_BASED_RATE_EST_TOKEN) == 0) linked_token = PIC_BASED_RATE_EST_NEW_TOKEN;
+    if (EB_STRCMP(token, INPUT_STAT_FILE_TOKEN) == 0) linked_token = TWO_PASS_STATS_TOKEN;
+    if (EB_STRCMP(token, OUTPUT_STAT_FILE_TOKEN) == 0) linked_token = TWO_PASS_STATS_TOKEN;
 
     if (EB_STRLEN(linked_token, WARNING_LENGTH) > 1) {
         const char *message_str = " will be deprecated soon, please use ";
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -37,6 +37,7 @@
 /* two pass token */
 #define PASS_TOKEN "--pass"
 #define TWO_PASS_STATS_TOKEN "--stats"
+#define PASSES_TOKEN "--passes"
 #define INPUT_STAT_FILE_TOKEN "-input-stat-file"
 #define OUTPUT_STAT_FILE_TOKEN "-output-stat-file"
 
@@ -323,6 +324,13 @@ static void set_two_pass_stats(const char* value, EbConfig *cfg) {
 #else
     cfg->stats = _strdup(value);
 #endif
+}
+
+static void set_passes(const char* value, EbConfig *cfg) {
+    (void)value;
+    (void)cfg;
+    /* empty function, we will handle passes at higher level*/
+    return;
 }
 
 static void set_input_stat_file(const char *value, EbConfig *cfg) {
@@ -920,6 +928,7 @@ ConfigEntry config_entry_2p[] = {
     // 2 pass
     {SINGLE_INPUT, PASS_TOKEN, "Multipass bitrate control (1: first pass, generates stats file , 2: second pass, uses stats file)", set_pass},
     {SINGLE_INPUT, TWO_PASS_STATS_TOKEN, "Filename for 2 pass stats(\"svtav1_2pass.log\" : [Default])", set_two_pass_stats},
+    {SINGLE_INPUT, PASSES_TOKEN, "Number of passes (1: one pass encode, 2: two passes encode)", set_passes},
     {SINGLE_INPUT, OUTPUT_STAT_FILE_TOKEN, "First pass stat file output", set_output_stat_file},
     {SINGLE_INPUT,
      INPUT_STAT_FILE_TOKEN,
@@ -2260,8 +2269,10 @@ uint32_t get_help(int32_t argc, char *const argv[]) {
                 "Usage: SvtAv1EncApp <options> -b dst_filename -i src_filename\n");
         fprintf(stderr, "\n%-25s\n", "Examples:");
         fprintf(stderr, "\n%-25s", "Two passes encode:");
-        fprintf(stderr, "\n\t%s", "SvtAv1EncApp --pass 1 -i <input> -b <output>");
-        fprintf(stderr, "\n\t%s\n", "SvtAv1EncApp --pass 2 -i <input> -b <output>");
+        fprintf(stderr, "\n\t%s", "SvtAv1EncApp <--stats svtav1_2pass.log> --pass 1 -b dst_filename -i src_filename");
+        fprintf(stderr, "\n\t%s", "SvtAv1EncApp <--stats svtav1_2pass.log> --pass 2 -b dst_filename -i src_filename");
+        fprintf(stderr, "\n    Or a combined cli:");
+        fprintf(stderr, "\n\t%s\n", "SvtAv1EncApp <--stats svtav1_2pass.log> --passes 2 -b dst_filename -i src_filename");
         fprintf(stderr, "\n%-25s\n", "Options:");
         while (config_entry_options[++options_token_index].token != NULL) {
             uint32_t check = check_long(
@@ -2451,6 +2462,53 @@ uint32_t get_number_of_channels(int32_t argc, char *const argv[]) {
         }
     }
     return 1;
+}
+
+static EbBool check_two_pass_conflicts(int32_t argc, char *const argv[])
+{
+    char     config_string[COMMAND_LINE_MAX_SIZE];
+    const char* conflicts[] = {
+        PASS_TOKEN,
+        INPUT_STAT_FILE_TOKEN,
+        OUTPUT_STAT_FILE_TOKEN,
+        NULL,
+    };
+    int i = 0;
+    const char* token;
+    while ((token = conflicts[i])) {
+        if (find_token(argc, argv, token, config_string) == 0) {
+            fprintf(stderr,
+                "--passes 2 conflicts with %s\n", token);
+            return EB_TRUE;
+        }
+        i++;
+    }
+    return EB_FALSE;
+}
+
+uint32_t get_passes(int32_t argc, char *const argv[], EncodePass pass[MAX_ENCODE_PASS]) {
+    char     config_string[COMMAND_LINE_MAX_SIZE];
+    uint32_t passes;
+    if (find_token(argc, argv, PASSES_TOKEN, config_string) != 0) {
+        pass[0] = ENCODE_SINGLE_PASS;
+        return 1;
+    }
+    passes = strtol(config_string, NULL, 0);
+    if (passes == 0 || passes > 2) {
+        fprintf(stderr,
+            "Error: The number of passes has to be within the range [1,%u]\n",
+            (uint32_t)MAX_ENCODE_PASS);
+        return 0;
+    }
+    if (passes == 1) {
+        pass[0] = ENCODE_SINGLE_PASS;
+        return 1;
+    }
+    if (check_two_pass_conflicts(argc, argv))
+        return 0;
+    pass[0] = ENCODE_FIRST_PASS;
+    pass[1] = ENCODE_LAST_PASS;
+    return 2;
 }
 
 void mark_token_as_read(const char *token, char *cmd_copy[], int32_t *cmd_token_cnt) {
@@ -2762,11 +2820,12 @@ EbErrorType read_command_line(int32_t argc, char *const argv[], EbConfig **confi
 
     for (index = 0; index < MAX_CHANNEL_NUMBER; ++index)
         config_strings[index] = (char *)malloc(sizeof(char) * COMMAND_LINE_MAX_SIZE);
-    // Copy tokens (except for CHANNEL_NUMBER_TOKEN ) into a temp token buffer hosting all tokens that are passed through the command line
-    size_t len = EB_STRLEN(CHANNEL_NUMBER_TOKEN, COMMAND_LINE_MAX_SIZE);
+    // Copy tokens (except for CHANNEL_NUMBER_TOKEN and PASSES_TOKEN ) into a temp token buffer hosting all tokens that are passed through the command line
+    size_t len = COMMAND_LINE_MAX_SIZE;
     for (token_index = 0; token_index < argc; ++token_index) {
         if ((argv[token_index][0] == '-') &&
             strncmp(argv[token_index], CHANNEL_NUMBER_TOKEN, len) &&
+            strncmp(argv[token_index], PASSES_TOKEN, len) &&
             !is_negative_number(argv[token_index]))
             cmd_copy[cmd_token_cnt++] = argv[token_index];
     }
--- a/Source/App/EncApp/EbAppConfig.h
+++ b/Source/App/EncApp/EbAppConfig.h
@@ -30,6 +30,13 @@ the App.
 */
 typedef enum AppPortActiveType { APP_PortActive = 0, APP_PortInactive } AppPortActiveType;
 
+typedef enum EncodePass {
+    ENCODE_SINGLE_PASS, //single pass mode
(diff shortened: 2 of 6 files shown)


xuguangxin:
@jsunintel , this is a simplified version of https://github.com/OpenVisualCloud/SVT-AV1/pull/998. please help review it.
thanks

ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
```diff
@@ -53,6 +54,20 @@ index 7495f35faa..e0a466b968 100755
...
                                   require libopencv opencv2/core/core_c.h cvCreateImageHeader -lopencv_core -lopencv_imgproc; } ||
+diff --git a/fftools/ffmpeg_opt.c b/fftools/ffmpeg_opt.c
+index 9d1489ce01..799e7eb7a5 100644
+--- a/fftools/ffmpeg_opt.c
++++ b/fftools/ffmpeg_opt.c
+@@ -1817,7 +1817,8 @@ static OutputStream *new_video_stream(OptionsContext *o, AVFormatContext *oc, in
+                      ost->logfile_prefix ? ost->logfile_prefix :
+                                            DEFAULT_PASS_LOGFILENAME_PREFIX,
+                      i);
+-            if (!strcmp(ost->enc->name, "libx264")) {
```
xuguangxin, line 65: @jamrial instead of use strcmp here. Is it possible add some AV_CODEC_CAP_NEED_STATS. to indicate we need a stats file name?
thanks

jamrial: This libx264 specific branch exists because libx264 creates the log file itself. Every other external encoder exports the stats in some buffer that the encoder wrapper/plugin then copies to avctx->stats_out for pass1 (Which is then dumped into the log file by the code here in ffmpeg_opt.c), then for pass2 the contents of the log file are written into avctx->stats_in that the wrapper/plugin then passes to the external library, again using some buffer provided by it.

If you want the log file creation to be done by SVT-AV1 then your suggestion for a codec cap to signal it sounds ok, but by no means call fopen() from within the ffmpeg wrapper/plugin like you're doing below. Either have SVT-AV1 take a filename parameter to fopen() it itself internally, or change the design to instead export the output pass data and read the input pass data using dedicated buffer/s that the wrapper can then pass to/from stats_{in,out}. See for example how libaom and libvpx handle it.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -2038,6 +2112,10 @@ uint32_t get_help(int32_t argc, char *const argv[]) {
         const char *empty_string         = "";
         fprintf(stderr,
                 "Usage: SvtAv1EncApp <options> -b dst_filename -i src_filename\n");
+        fprintf(stderr, "\n%-25s\n", "Examples:");
+        fprintf(stderr, "\n%-25s", "Two pass with a bitrate of 3000kbps:");
+        fprintf(stderr, "\n\t%s", "SvtAv1EncApp --pass 1 --rc 1 --tbr 3000  -i <input> -b <output>");
```
hassount, line 2117: so the encoder would basically assume that there's a file with a specific name outputted to read the second pass data from ?

hassount:
@xuguangxin 
thank you for the PR, a few comments:
- this PR does not introduce a quality change, rather just an interface change, so I believe the quality box should not be checked for this PR
- within the library there's currently no support for a 2-pass rc so I recommend removing the printouts suggesting such because it is misleading as to what capabilities would the library have. 
- can we make sure that the current input / output / 2 pass commandlines is still valid after this change ?
- just to be on the same page, this does not replace #998 yet because it does not contain the 2 pass one cli like libaom yet right ?

ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
xuguangxin: Hi @jamrial ,
thanks for the replay. 
Currently, svt library interface does not support buffer as input. it need FILE*. See https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Source/API/EbSvtAv1Enc.h#L207
.  Is fopen not allowed in ffmpeg plugin?
thanks

Source/App/EncApp/EbAppConfig.c
xuguangxin: Yes. The default file name is [ svtav1_2pass.log ](https://github.com/OpenVisualCloud/SVT-AV1/pull/1365/files#diff-2b4d204fc9f0f34f82a4e454d6eea4f7R861). you can change it using --stats

xuguangxin:
@hassount thanks for suggestion.
> @xuguangxin
> thank you for the PR, a few comments:
> 
> * this PR does not introduce a quality change, rather just an interface change, so I believe the quality box should not be checked for this PR

sure I will uncheck it

> * within the library there's currently no support for a 2-pass rc so I recommend removing the printouts suggesting such because it is misleading as to what capabilities would the library have.

sure I will change it.

> * this PR does not seem to link the stat files to the existing infra structure so that the library would use such, only a command line change without an effect on the library right ?

yes, It's no need to change the library. unless @jamrial  said, fopen is not allowed in ffmpeg plugin 

> * just to be on the same page, this does not replace #998 yet because it does not contain the 2 pass one cli like libaom yet right ?

yes, it's x264 style 2 pass, ffmpeg uses this style too. Seems you prefer libaom style two passes. Let me add --passes comandline param like libaom. 

thanks

xuguangxin:
> yes, it's x264 style 2 pass, ffmpeg uses this style too. Seems you prefer libaom style two passes. Let me add --passes comandline param like libaom.
done, @hassount please help review.

1480c1:
I personally would prefer it if we could move the argument parsing stuff into the library so we can support something like [`x264-params`](https://trac.ffmpeg.org/wiki/Encode/H.264#Overwritingdefaultpresetsettings) in order to make maintenance simpler in the future since that way, we do not have to expose much in terms of the command line for FFmpeg other than the bare minimum, and the rest of the options related to bitstream optimization can be handled through something like `svt-params` or similar. Plus it could complement our current config file options

xuguangxin:
> I personally would prefer it if we could move the argument parsing stuff into the library so we can support something like [`x264-params`](https://trac.ffmpeg.org/wiki/Encode/H.264#Overwritingdefaultpresetsettings) in order to make maintenance simpler in the future since that way, we do not have to expose much in terms of the command line for FFmpeg other than the bare minimum, and the rest of the options related to bitstream optimization can be handled through something like `svt-params` or similar. Plus it could complement our current config file options

yes, it's a good idea. the command line parser need a big refact. please fill a issues, someone maybe interested on this.

thanks

xuguangxin:
@1480c1 do you know where the frame number printout?

If we use  `SvtAv1EncApp --passes 2   -i <input> -b <output>` to encode a 200 frames clip. [This ](https://github.com/OpenVisualCloud/SVT-AV1/pull/1365/commits/ad9b709d9ab1d74ba472d94cefc4ff9cfa87d1e1#diff-1fbcd93da2b5deda512119fe8e198918R207)line will report 400 frames at second pass.

1480c1:
frame number is printed out in EbAppProcessCmd.c in `process_output_stream_buffer`
there should be a line that has `++frame_count;`

xuguangxin:
> frame number is printed out in EbAppProcessCmd.c in `process_output_stream_buffer`
> there should be a line that has `++frame_count;`

seems somebody used a static variable to count frames. 
https://github.com/OpenVisualCloud/SVT-AV1/blob/d9522f82f4510b3d886b7b3c41d441a974d50f4b/Source/App/EncApp/EbAppProcessCmd.c#L1140

xuguangxin:
@hassount , 2 workaround suggested by you added please help review. 28e5a6e57a7ca9b20e94b85411fcd3d97c074cb4 f265c876cc8e51bb9a25455a0971aae988aaf918

following cases tested, and got bit match results

- SvtAv1EncApp -i input.y4m -b pass1.ivf  -input-stat-file stat.stat  --preset 8 -lp 1 && SvtAv1EncApp -i input.y4m -b pass2.ivf  -input-stat-file stat.stat  --preset 4 -lp 1
- SvtAv1EncApp -i input.y4m -b pass1.ivf  --pass 1--stats stat.stat  --preset 8 -lp 1 && SvtAv1EncApp -i input.y4m -b pass2.ivf  -pass 2 --stats stat.stat  --preset 4 -lp 1
- SvtAv1EncApp -i input.y4m -b pass2.ivf  --passes 2 --preset 4

ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
jamrial: Ideally that should be avoided, yes.

I see you implemented a new stats filename parameter much like libx264, so that should be good, and we can go ahead with your idea of adding a codec capability for this.

1480c1:
@xuguangxin please add a define or similar for `_WIN32` for `strdup` to be changed to `_strdup`
```
D:\a\SVT-AV1\SVT-AV1\Source\App\EncApp\EbAppConfig.c(295,18): error C2220: the following warning is treated as an error [D:\a\SVT-AV1\SVT-AV1\Build\Source\App\EncApp\SvtAv1EncApp.vcxproj]
D:\a\SVT-AV1\SVT-AV1\Source\App\EncApp\EbAppConfig.c(295,18): warning C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _strdup. See online help for details. [D:\a\SVT-AV1\SVT-AV1\Build\Source\App\EncApp\SvtAv1EncApp.vcxproj]
```

xuguangxin:
@1480c1 seems, we are passed windows build. 
Current build time is too long, we need wait more than 3 hours to see the build result.
Is any way to increase the build machine to reduce the build time?

thanks

ffmpeg_plugin/0001-Add-ability-for-ffmpeg-to-run-svt-av1.patch
xuguangxin: thanks for the feedback, hassount mentioned you are still working on upstream the ffmpeg patch. To reduce complicity,
let us revisit two pass encode in ffmpeg after it's upstreamed.
thanks

1480c1:
There could be a few things to optimize the msvc builds, one could be not having `BUILD_TESTING` for debug. For the mingw-w64 FFmpeg build, I'm not sure if the ccache is implemented properly, mainly because of github action's caching key matching.

For trying to use a caching system for msvc, we could try looking into sccache and see if we can use it along with using ninja as the build generator instead of sln so we can use `CMAKE_CXX_COMPILER_LAUNCHER=sccache` or similar

EricLamVCD:
Hi @xuguangxin , I tested obj-1 fast 8bit and obj-1 fast 10bit. M4 was lossless but M0 was not lossless. 

As an example, I verified the bdrate deviation with the niklas360p_60f_10bit clip for the MOD and REF.

Example MOD Command Line:
`(/usr/bin/time --verbose   ./SvtAv1EncApp --passes 2 -enc-mode 0 -w 640 -h 360 -bit-depth 10 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119  -lad 0 -lp 1  -i  /home/ebrisk/stream/10bit/niklas360p_60f_10bit.yuv  -b  bitstreams/svt_M0_niklas360p_60f_10bit_Q63.bin )  > bitstreams/svt_M0_niklas360p_60f_10bit_Q63.txt 2>&1 `

Example REF Command Line:
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 640 -h 360 -bit-depth 10 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  -i  /home/ebrisk/stream/10bit//niklas360p_60f_10bit.yuv  -b  bitstreams/svt_M8_niklas360p_60f_10bit_Q63.bin  --output-stat-file bitstreams/svt_M8_niklas360p_60f_10bit_Q63.file)  > bitstreams/svt_M0_niklas360p_60f_10bit_Q63.txt 2>&1 

(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 0 -w 640 -h 360 -bit-depth 10 -fps-num 30000 -fps-denom 1000 -q 63  -n  60  -intra-period  119  -lad 0 -lp 1  -i  /home/ebrisk/stream/10bit//niklas360p_60f_10bit.yuv  -b  bitstreams/svt_M0_niklas360p_60f_10bit_Q63.bin  --input-stat-file bitstreams/svt_M8_niklas360p_60f_10bit_Q63.file)  >> bitstreams/svt_M0_niklas360p_60f_10bit_Q63.txt 2>&1 `

xuguangxin:
@EricLamVCD , thanks for the test.
Could you run REF commandline more times, to make sure it's no round 2 round issues?
for the code,  M0 is no different than M4. If M4 can get a bit match result, M0 can get it too.
thanks

EricLamVCD:
@xuguangxin , I ran it for the second time. The results are relatively the same again for M0.

xuguangxin:
hmm, preset 3 will bit match too. The mismatch starts from preset 2. 
@hassount which encode tool added in preset 2? Do we have global or static variables in this tool?
tanks

hassount:
Could be the prediction structure tables guessing from what was changed in #998

xuguangxin:
@hassount you are right, --pass 2 preset 2 will change the reference model 

reference model for: SvtAv1EncApp -i input.y4m -b pass1.ivf **--pass 1**--stats stat.stat --preset 8 -lp 1
image

reference model for: SvtAv1EncApp -i input.y4m -b pass2.ivf **-pass 2** --stats stat.stat --preset 2 -lp 1
image

reference model for: SvtAv1EncApp -i input.y4m -b pass2.ivf **--passes 2** --preset 2
image

which one is right? the --pass 2 or --passes 2?

xuguangxin:
here is single pass preset 2 reference model: SvtAv1EncApp -i input.y4m -b pass_single.ivf  --preset 2 -lp 1

image

it match with  pass 2.

hassount:
Every preset has to use its own reference model, so the solution would be to change the allocation of the ref model tables to be done on the fly rather than being static tables like what’s done in #998

xuguangxin:
ok, thanks for suggestion. let me speed some time on this.

xuguangxin:
@EricLamVCD please try again.
@hassount please review https://github.com/OpenVisualCloud/SVT-AV1/pull/1365/commits/ab481e3b4b6d9aa71e09c20e66c0eab5071cf157
thanks

EricLamVCD:
Hi @xuguangxin , 
We did some tests and found the following behavior: 
1) We ran M4 for full set of obj-1 8bit and M0 for a efw clips with lad 0. The PR is lossless and everything is fine. 
2) Since 2 pass uses look ahead distance. We ran a test for M4 lad 16 and the PR is showing bdrate deviation as seen in the following table:
  | Speed Deviation  | BD-Rate   Deviation | Memory   Deviation
-- | -- | -- | --
M4 - 8bit | 0.03% | 0.48% | -0.16%
3) We ran the PR mod 3 times with lad 16 for M4 and found significant r2r for all 3 runs.

xuguangxin:
@EricLamVCD , thanks for the test. 

For the big r2r issue, it maybe some other static/global variables. For smaller one, need more check.
Could you help provide your commandline and clips for item 2, 3, 4,
thanks

EricLamVCD:
Here are the commands: 
**PR MOD Command**: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 0 -w 1920 -h 1080 -bit-depth 8 -fps-num 50000 -fps-denom 1000  -n  60 -q 43  -intra-period  119 --passes 2  -lad 16 -lp 1  -i  /home/ubuntu/stream/8bit/ducks_take_off_1080p50_60f.yuv  -b  bitstreams/svt_M0_ducks_take_off_1080p50_60f_Q43.bin )  > bitstreams/svt_M0_ducks_take_off_1080p50_60f_Q43.txt 2>&1 `

**PR REF command**:
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 1920 -h 1080 -bit-depth 8 -fps-num 50000 -fps-denom 1000  -n  60 -q 20  -intra-period  119 -lad 0 -lp 1  -i  /home/ubuntu/stream/8bit/ducks_take_off_1080p50_60f.yuv  -b  bitstreams/svt_M8_ducks_take_off_1080p50_60f_Q20.bin  --output-stat-file bitstreams/svt_M8_ducks_take_off_1080p50_60f_Q20.file)  > bitstreams/svt_M0_ducks_take_off_1080p50_60f_Q20.txt 2>&1 `
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 0 -w 1920 -h 1080 -bit-depth 8 -fps-num 50000 -fps-denom 1000 -q 20  -n  60  -intra-period  119  -lad 16 -lp 1  -i  /home/ubuntu/stream/8bit/ducks_take_off_1080p50_60f.yuv  -b  bitstreams/svt_M0_ducks_take_off_1080p50_60f_Q20.bin  --input-stat-file bitstreams/svt_M8_ducks_take_off_1080p50_60f_Q20.file)  >> bitstreams/svt_M0_ducks_take_off_1080p50_60f_Q20.txt 2>&1 `

xuguangxin:
Hi @EricLamVCD ,
seems you use a different qp for **PR MOD Command** & **PR REF command**. Is it the root cause?

hassount:
> Hi @EricLamVCD ,
> seems you use a different qp for **PR MOD Command** & **PR REF command**. Is it the root cause?

Actually I am wondering if the issue here is the stat file, in the combined cli, and assuming many jobs are running at the same time, is the default stat file being used by all of the encodes at the same time ? in which case, such would most likely create a R2R issue
to know whether it's the case or not, we would need to add the -stat token and specify a unique file name when using 2-pass in multiple encodes. 
on the code side, we probably need to do one of two things, either lock the stat file when opening it for reading or writing / or at least display a warning when the default stat file is used that on multiple encodes one would need to specify unique file names.

EricLamVCD:
> Hi @EricLamVCD ,
> seems you use a different qp for **PR MOD Command** & **PR REF command**. Is it the root cause?

@xuguangxin It is not the root cause. I just happened to copy the command with the wrong QP. Sorry about that.

EricLamVCD:
We tested by adding the --stat command line to #1365 to see if there was a difference. Here is a sample of the command line: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 0 -w 1920 -h 1080 -bit-depth 8 -fps-num 30000 -fps-denom 1001  -n  60 -q 20  -intra-period  119  --passes 2  -lad 16 -lp 1  -i  /home/ubuntu/stream/8bit/aspen_1080p_60f.yuv  --stats  bitstreams/svt_M0_aspen_1080p_60f_Q20.stat   -b  bitstreams/svt_M0_aspen_1080p_60f_Q20.bin )  > bitstreams/svt_M0_aspen_1080p_60f_Q20.txt 2>&1 `

We found that the bdrate deviation is closer a lot smaller compared by using the --stat parameter. Here are the results: 
Tested obj-1 8bit and 10bit clips: 
  | Speed Deviation  | BD-Rate   Deviation | Memory   Deviation
-- | -- | -- | --
M4 - 8bit | 0.52% | 0.00% | 7.15%
M4 - 10bit | 0.06% | -0.01% | 10.84%

The results are lossy. 

We think that the bd rate deviation may be occurring due to the existing R2R issue within master when enabling lad 16. We confirmed this by running 3 runs of M4 for both obj-1 8bit and 10bit clips.

xuguangxin:
> on the code side, we probably need to do one of two things, either lock the stat file when opening it for reading or writing / or at least display a warning when the default stat file is used that on multiple encodes one would need to specify unique file names.

done, tested on linux and windows
@hassount please help review it.
https://github.com/OpenVisualCloud/SVT-AV1/pull/1365/commits/5dab89a5a7b3142c39f4ba39b80863f926f375cd

EricLamVCD:
We tested the #1365 with lad 0 and found the following :
  | Speed Deviation  | BD-Rate   Deviation | Memory   Deviation
-- | -- | -- | --
M0 - 8bit | -0.04% | 0.00% | 6.37%
M2 - 8bit | 0.09% | 0.00% | 6.10%
M4 - 8bit | 0.57% | 0.00% | 5.70%
M0 - 10bit | 1.96% | 0.00% | 9.97%
M2 - 10bit | 2.16% | 0.00% | 10.16%
M4 - 10bit | 2.45% | 0.00% | 9.64%

We ran each encoding twice to check if there were any r2r which we did not find any.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -2038,6 +2166,12 @@ uint32_t get_help(int32_t argc, char *const argv[]) {
         const char *empty_string         = "";
         fprintf(stderr,
                 "Usage: SvtAv1EncApp <options> -b dst_filename -i src_filename\n");
+        fprintf(stderr, "\n%-25s\n", "Examples:");
+        fprintf(stderr, "\n%-25s", "Two passes encode:");
+        fprintf(stderr, "\n\t%s", "SvtAv1EncApp --pass 1 -i <input> -b <output>");
```
hassount, line 2171: I suggest adding the --stats output even being optional so that the user is aware that this is now done using a file.

Source/App/EncApp/EbAppConfig.c
```diff
@@ -2038,6 +2166,12 @@ uint32_t get_help(int32_t argc, char *const argv[]) {
...
+        fprintf(stderr, "\n\t%s", "SvtAv1EncApp --pass 2 -i <input> -b <output>");
+        fprintf(stderr, "\n\tOr combined one:");
```
hassount, line 2173: suggest changing to fprintf(stderr, "\n\tOr a combined cli:");

xuguangxin:
@hassount done.

@EricLamVCD for the speed and memory changes. Did you measure it before and after my patch?
I just rebased my patch to latest. it will not impact speed or memory.
thanks

ttrigui:
I run some tests to evaluate 2-pass versus 1-pass and I found the following:

- 2 command lines: no memory impact

- 1 combined command line: an 8% memory increase is recorded for my case ( using dark720p for preset 4)

The results are collected using /usr/bin/time --verbose on Linux.
The memory allocation the encoder outputs in debug mode "total allocated memory" shows the same results as 1-pass though.
Removing allocating and writing to the outputstream in packetization " output_stream_ptr->p_buffer = (uint8_t *)malloc(output_stream_ptr->n_alloc_len); " for the 1st pass did not impact the memory results
Trying to remove the new way of allocating the prediction structure did not help either
@xuguangxin any thoughts about what can cause the memory increase for the combined command?

xuguangxin:
@ttrigui I checked with valgrind. it's no leak for --passes 2. 
could you share your commandline and how you measure the memory?

thanks

ttrigui:
> @ttrigui I checked with valgrind. it's no leak for --passes 2.
> could you share your commandline and how you measure the memory?
> 
> thanks

Thanks @xuguangxin for verifying. You can use any input I believe they should give similar results. I have used a 720p clip.
The command I used it the following:

- 1-pass:  (/usr/bin/time --verbos ./SvtAv1EncApp -enc-mode 4 -q 20  -intra-period  119 -lad 0 -lp 1  -i  /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m  -b  bitstreams/svt_M4_dark720p_60f_Q20.bin)  > bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1 


- 2-pass 1 combined command: (/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -q 20  -intra-period  119 -lad 0 -lp 1  --passes 2   -i  /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m  --stats bitstreams/svt_M4_dark720p_60f_Q20.stat  -b  bitstreams/svt_M4_dark720p_60f_Q20.bin )  > bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1 

- 2-pass 2 commands: (/usr/bin/time --verbose ./run-svt-2p-m4.sh )
 run-svt-2p-m4.sh has the following commands ./SvtAv1EncApp -enc-mode 8 -q 20  -intra-period  119 -lad 0 -lp 1  -i  /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m  -b  bitstreams/svt_M8_dark720p_60f_Q20.bin  --output-stat-file bitstreams/svt_M8_dark720p_60f_Q20.file  > bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1 
./SvtAv1EncApp -enc-mode 4 -q 20  -n  60  -intra-period  119 -lad 0 -lp 1  -i  /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m  -b  bitstreams/svt_M4_dark720p_60f_Q20.bin  --input-stat-file bitstreams/svt_M8_dark720p_60f_Q20.file  >> bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1 

I hope this helps.

xuguangxin:
@ttrigui which tool you used to measure memory? Did you measure the virtual or RSS memory?
thanks

1480c1:
@xuguangxin can you rebase?

xuguangxin:
@ttrigui here is my result
2 passes
```
3454.41user 4.87system 52:33.91elapsed 109%CPU (0avgtext+0avgdata 1170252maxresident)k
0inputs+75784outputs (0major+3951819minor)pagefaults 0swaps

```

pass 2

```
2889.24user 2.91system 44:15.15elapsed 108%CPU (0avgtext+0avgdata 1150028maxresident)k
0inputs+37240outputs (0major+2350951minor)pagefaults 0swaps
```

the difference is 1170252/1150028 - 1 = 1%.

xuguangxin:
done, @1480c1

ttrigui:
> @ttrigui here is my result
> 2 passes
> 
> ```
> 3454.41user 4.87system 52:33.91elapsed 109%CPU (0avgtext+0avgdata 1170252maxresident)k
> 0inputs+75784outputs (0major+3951819minor)pagefaults 0swaps
> ```
> 
> pass 2
> 
> ```
> 2889.24user 2.91system 44:15.15elapsed 108%CPU (0avgtext+0avgdata 1150028maxresident)k
> 0inputs+37240outputs (0major+2350951minor)pagefaults 0swaps
> ```
> 
> the difference is 1170252/1150028 - 1 = 1%.

Thank you @xuguangxin. Have you used the command lines I shared with you? which machine did you use to evaluate the memory impact?

xuguangxin:
@ttrigui yes the commandline is same, I compared 2passes.log and 2pass.log. 
But I used debug version. Will it impact the result?
I can retest it use release if you think following comandline has no issue.

- $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --passes 2 --preset 4 &> 2passes.log
- $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --pass 1 --preset 8 &> 1pass.log
- $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --pass 2 --preset 4 &> 2pass.log

ttrigui:
> @ttrigui yes the commandline is same, I compared 2passes.log and 2pass.log.
> But I used debug version. Will it impact the result?
> I can retest it use release if you think following comandline has no issue.
> 
> * $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --passes 2 --preset 4 &> 2passes.log
> * $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --pass 1 --preset 8 &> 1pass.log
> * $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --pass 2 --preset 4 &> 2pass.log

thank you @xuguangxin for your reply.
We usually use release to evaluate speed and thus all the memory results reported are based on running release.
However it should not matter as this is a relative number, but I still prefer using release unless we are doing some debugging.

I believe that you are using the same executable in the tests you run. I assume that you are using executable generated after your PR. I actually used two versions of the code one before your PR to evaluate the 2 pass with 2 command lines and a version after your changes to evaluate the combined command. If you are using the same version of the code,can you please try testing before your changes and let me know your findings?

xuguangxin:
@ttrigui I am little confused. 
you said , 2 command lines is the same before and after my patches. 
My test shows, after my patch,  1 combined command line only get 1% higher than 2 command lines' pass 2.
Is this mean 1 combined command line's memory only 1% higher than 2 command lines pass 2 before my patches?

> 2 command lines: no memory impact
> 
> 1 combined command line: an 8% memory increase is recorded for my case ( using dark720p for preset 4)

ttrigui:
> @ttrigui I am little confused.
> you said , 2 command lines is the same before and after my patches.
> My test shows, after my patch, 1 combined command line only get 1% higher than 2 command lines' pass 2.
> Is this mean 1 combined command line's memory only 1% higher than 2 command lines pass 2 before my patches?
> 
> > 2 command lines: no memory impact
> > 1 combined command line: an 8% memory increase is recorded for my case ( using dark720p for preset 4)

@xuguangxin nope. In that comment my reference was 1-pass.
Ok Let's make the picture clear here. 
When we first tested the PR we run before your commit (2 commands) and after your commit (1 combined command). We have never tested 2 commands after your commit assuming that memory was not impacted for that case.However you actually draw my attention to it and that might be a better scenario to debug if the memory is impacted after your commit when testing with 2 commands.

EricLamVCD:
> @ttrigui I am little confused.
> you said , 2 command lines is the same before and after my patches.
> My test shows, after my patch, 1 combined command line only get 1% higher than 2 command lines' pass 2.
> Is this mean 1 combined command line's memory only 1% higher than 2 command lines pass 2 before my patches?

Hi @xuguangxin , how did you run your test and what were the command lines? I was able to run your previously mentioned methods using 1 encode command at a time. However, I am unable to run multiple encodings with mulitple different QP's using your method. Although, I was able to run the new 2pass 1 command line method if I specified unique stat files for each encoding I ran. 

For my testing, I ran a few 360p clips from the obj-1-8bit fast test set for 5 qp's (20,32,43,55,63) for M4 using this PR. 

**Here are the following results from my testing:** 
1. I ran using --pass 1 and --pass 2 command line params using my output method.  Both the first pass and second pass failed to run.  
Here are the sample command lines: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  --pass 1  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  -b  bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.bin )  > bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000 -q 63  -n  60  -intra-period  119 -lad 0 -lp 1  --pass 2   -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin ) >> bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `

2. I ran using --stats, --pass 1 and --pass 2 command line params using my output method. Only the first pass ran (M8) while M4 did not run at all. There weren't any errors which indicate what the problem was. 
Here are the sample command lines: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  --pass 1  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  --stats bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.bin   -b  bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.bin )  > bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000 -q 63  -n  60  -intra-period  119 -lad 0 -lp 1  --pass 2   -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  --stats bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.stat   -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin ) >> bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `

3. I tested using xuguangxin's method for multiple qp's but I recevied an error.
xuguangxin's  commands:
> * $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --pass 1 --preset 8 &> 1pass.log
> * $ /usr/bin/time SvtAv1EncApp -lad 0 -q 20 -intra-period 119 -lp 1 -i ~/bits/SVT/ducks_take_off_1080p50_60f.y4m -b two_passes.ivf --pass 2 --preset 4 &> 2pass.log

Here are the sample command lines: 
`/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  --pass 1  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  -b  bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.bin &> bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.log`
`/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000 -q 63  -n  60  -intra-period  119 -lad 0 -lp 1  --pass 2   -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin &> bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.log 2>&1 `

Here is the following error in the text file: 
`ERROR: locking svtav1_2pass.log failed, is it used by other encoder?
Error instance 1: can't open stats file svtav1_2pass.log for write `

4. I ran using the old 2 pass method and the runs succeeded. 
Here are the sample command lines: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  -b  bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.bin  --output-stat-file bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.file)  > bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000 -q 63  -n  60  -intra-period  119  -lad 0 -lp 1  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin  --input-stat-file bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.file)  >> bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `

5. I ran using the new 1 command line 2 pass method. Only one encoding was able to run when trying to run multiple encodings in parallel. The other files had an error present in the text files. 
Here is a sample command line: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  --passes 2  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin )  > bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `

Here is the error output: 
`Error instance 1: can't open stats file svtav1_2pass.log for write 
Error in configuration, could not begin encoding! ... `

6. Using new 2pass 1 command line method with --stats param, I was able to run multiple encodings in parallel. 
Here is a sample command line: 
(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  --passes 2  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  --stats bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.stat   -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin )  > bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1

xuguangxin:
@EricLamVCD my comamdline did  not provide --stats file, so it will open a default for us in current directory.  We add some check to make sure only one instance can open the default stats file.
If you want run multi 2 pass encode, maybe you can:
1. use --stats. or.
2. put run encode in different directory.

xuguangxin:
@ttrigui I tried your 2 pass 2 commands. It's about 0.03% difference before and after my patch.
So the memory increment is pretty small.

- 2-pass 2 commands: (/usr/bin/time --verbose ./run-svt-2p-m4.sh )
run-svt-2p-m4.sh has the following commands ./SvtAv1EncApp -enc-mode 8 -q 20 -intra-period 119 -lad 0 -lp 1 -i /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m -b bitstreams/svt_M8_dark720p_60f_Q20.bin --output-stat-file bitstreams/svt_M8_dark720p_60f_Q20.file > bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1
./SvtAv1EncApp -enc-mode 4 -q 20 -n 60 -intra-period 119 -lad 0 -lp 1 -i /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m -b bitstreams/svt_M4_dark720p_60f_Q20.bin --input-stat-file bitstreams/svt_M8_dark720p_60f_Q20.file >> bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1
 

image

ttrigui:
> @ttrigui I tried your 2 pass 2 commands. It's about 0.03% difference before and after my patch.
> So the memory increment is pretty small.
> 
> * 2-pass 2 commands: (/usr/bin/time --verbose ./run-svt-2p-m4.sh )
>   run-svt-2p-m4.sh has the following commands ./SvtAv1EncApp -enc-mode 8 -q 20 -intra-period 119 -lad 0 -lp 1 -i /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m -b bitstreams/svt_M8_dark720p_60f_Q20.bin --output-stat-file bitstreams/svt_M8_dark720p_60f_Q20.file > bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1
>   ./SvtAv1EncApp -enc-mode 4 -q 20 -n 60 -intra-period 119 -lad 0 -lp 1 -i /home/ubuntu/stream/y4m/720p/dark/dark720p_60f.y4m -b bitstreams/svt_M4_dark720p_60f_Q20.bin --input-stat-file bitstreams/svt_M8_dark720p_60f_Q20.file >> bitstreams/svt_M4_dark720p_60f_Q20.txt 2>&1
> 
> image

thank you @xuguangxin can you please share the platform/machine you are using for testing?

EricLamVCD:
> @EricLamVCD my comamdline did not provide --stats file, so it will open a default for us in current directory. We add some check to make sure only one instance can open the default stats file.
> If you want run multi 2 pass encode, maybe you can:
> 
> 1. use --stats. or.
> 2. put run encode in different directory.

Hi @xuguangxin , I had a flaw with my 2nd test result because my stats file naming was incorrect. I ran the the tests again to see if the run would succeed. 

1. For this first test, I ran with the `--stats` param with the same name as the 1st and 2nd pass. However, only the first pass was able to run. Here are the command lines: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  --pass 1  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  --stats bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.stat  -b  bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.bin )  > bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000 -q 63  -n  60  -intra-period  119 -lad 0 -lp 1  --pass 2   -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  --stats bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.stat  -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin ) >> bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `

2. @hassount mentioned that we could combine the 1st pass and 2nd pass commands with &&. So, I did that and was able to run the runs successfully with out any issue. Here is a sample command: 
`(/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 8 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000  -n  60 -q 63  -intra-period  119 -lad 0 -lp 1  --pass 1  -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  --stats bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.stat  -b  bitstreams/svt_M8_thaloundeskmtg360p_60f_Q63.bin )  > bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1  && (/usr/bin/time --verbose   ./SvtAv1EncApp -enc-mode 4 -w 640 -h 360 -bit-depth 8 -fps-num 30000 -fps-denom 1000 -q 63  -n  60  -intra-period  119 -lad 0 -lp 1  --pass 2   -i  /home/ubuntu/stream/8bit/thaloundeskmtg360p_60f.yuv  --stats bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.stat  -b  bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.bin ) >> bitstreams/svt_M4_thaloundeskmtg360p_60f_Q63.txt 2>&1 `

Going forward with these results, I will be testing the PR again using the --stats parameter.

xuguangxin:
@ttrigui , i tested on a 104 cores xeon.

@EricLamVCD thanks

xuguangxin:
thanks every one help on this.

PR #1402, opened 2020-08-07
macOS: use upload and download artifact to make sure it uses the same binaries

1480c1:
# Description

make sure we are using the correct binaries for the unit tests in the case where we do not have the same environment as the one that built it, use parallel instead of xargs instead to handle multi-process, and don't clean the directory so we can save time recompiling

# Issue

# Author(s)
@1480c1

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -101,40 +101,33 @@ jobs:
   prep-tests:
     runs-on: [self-hosted, macos]
     name: Build unit tests
+    env:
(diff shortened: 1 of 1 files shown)


hassount:
is #1400 still needed ?

1480c1:
yes since that one affects the normal macOS ci builds while this one affects just the unit tests

PR #1469, opened 2020-09-03
2passes memory interface

xuguangxin:
# Description
Add memory interface for 2passes stats in and out.
Most of encoders like libaom will use memory as input/output for 2 passes stats.
This patch will make SvtAv1 follow the same tradition. 

# Issue

# Author(s)
guangxin.xu@intel.com

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -221,8 +221,8 @@ typedef struct EbSvtAv1EncConfiguration {
     *
     * Default is 0.*/
     EbBool use_qp_file;
-    /* Input stats file */
-    FILE *input_stat_file;
+    /* input buffer for the second pass */
+    SvtAv1FixedBuf rc_twopass_stats_in;
     /* output stats file */
     FILE *output_stat_file;
     /* Enable picture QP scaling between hierarchical levels
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -63,6 +63,18 @@ typedef enum {
     SUPERRES_MODES
 } SUPERRES_MODE;
 
+typedef enum {
+    SVT_AV1_STREAM_INFO_START = 1,
+
+    // The output is SvtAv1FixedBuf*
+    // Two use this, you need:
+    // 1. set the EbSvtAv1EncConfiguration.rc_firstpass_stats_out to EB_TRUE
+    // 2. call this when you got EB_BUFFERFLAG_EOS
+    SVT_AV1_STREAM_INFO_FIRST_PASS_STATS_OUT = SVT_AV1_STREAM_INFO_START,
+
+    SVT_AV1_STREAM_INFO_END,
+} SVT_AV1_STREAM_INFO_ID;
+
 /*!\brief Generic fixed size buffer structure
  *
  * This structure is able to hold a reference to any fixed size buffer.
@@ -223,8 +235,16 @@ typedef struct EbSvtAv1EncConfiguration {
     EbBool use_qp_file;
     /* input buffer for the second pass */
     SvtAv1FixedBuf rc_twopass_stats_in;
-    /* output stats file */
-    FILE *output_stat_file;
+    /* generate first pass stats output.
+    * when you set this to EB_TRUE, and you got the EB_BUFFERFLAG_EOS,
+    * you can get the encoder stats using:
+    *
+    * SvtAv1FixedBuf first_pass_stat;
+    * EbErrorType ret = svt_av1_enc_get_stream_info(component_handle,
+    *     SVT_AV1_STREAM_INFO_FIRST_PASS_STATS_OUT, &first_pass_stat);
+    *
+    * Default is 0.*/
+    EbBool rc_firstpass_stats_out;
     /* Enable picture QP scaling between hierarchical levels
     *
     * Default is null.*/
@@ -780,6 +800,16 @@ EB_API void svt_av1_enc_release_out_buffer(EbBufferHeaderType **p_buffer);
      * @ *p_buffer           Output buffer. */
 EB_API EbErrorType svt_av1_get_recon(EbComponentType *   svt_enc_component,
                                     EbBufferHeaderType *p_buffer);
+
+/* OPTIONAL: get stream information
+     *
+     * Parameter:
+     * @ *svt_enc_component  Encoder handler.
+     * @ *stream_info_id SVT_AV1_STREAM_INFO_ID.
+     * @ *info         output, the type depends on id */
+EB_API EbErrorType svt_av1_enc_get_stream_info(EbComponentType *    svt_enc_component,
+                                    uint32_t stream_info_id, void* info);
+
 
 /* STEP 6: Deinitialize encoder library.
      *
--- a/Source/API/EbSvtAv1Enc.h
+++ b/Source/API/EbSvtAv1Enc.h
@@ -63,6 +63,15 @@ typedef enum {
     SUPERRES_MODES
 } SUPERRES_MODE;
 
+/*!\brief Generic fixed size buffer structure
+ *
+ * This structure is able to hold a reference to any fixed size buffer.
+ */
+typedef struct SvtAv1FixedBuf {
+    void *buf;       /**< Pointer to the data. Does NOT own the data! */
+    uint64_t sz;       /**< Length of the buffer, in chars */
+} SvtAv1FixedBuf; /**< alias for struct aom_fixed_buf */
+
 // Will contain the EbEncApi which will live in the EncHandle class
 // Only modifiable during config-time.
 typedef struct EbSvtAv1EncConfiguration {
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -1686,6 +1686,8 @@ void eb_config_dtor(EbConfig *config_ptr) {
         config_ptr->stat_file = (FILE *)NULL;
     }
     if (config_ptr->input_stat_file) {
+        free(config_ptr->rc_twopass_stats_in.buf);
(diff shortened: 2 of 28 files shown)


xuguangxin:
@hassount @anaghdin please help review.

thanks

Source/API/EbSvtAv1Enc.h
```diff
@@ -63,6 +63,24 @@ typedef enum {
...
+    SVT_AV1_STREAM_INFO_END,
+} SVT_AV1_STREAM_INFO_ID;
+
+/*!\brief Generic fixed size buffer structure
+ *
+ * This structure is able to hold a reference to any fixed size buffer.
+ */
+typedef struct SvtAv1FixedBuf {
+    void *buf;       /**< Pointer to the data. Does NOT own the data! */
+    uint64_t sz;       /**< Length of the buffer, in chars */
```
1480c1, line 81:
```suggestion
    size_t sz;       /**< Length of the buffer, in chars */
```

maybe better to use `size_t` since it's usually used to represent objects etc size, even though it might be the same size as uint64_t, but it will also match aom's

Source/App/EncApp/EbAppConfig.c
```diff
@@ -1887,6 +1888,83 @@ static int32_t read_config_file(EbConfig *config, char *config_path, uint32_t in
...
+EbBool load_twopass_stats_in(EbConfig *config)
+{
+#ifdef _WIN32
+    int fd = _fileno(config->input_stat_file);
+    struct _stat file_stat;
+    int ret = _fstat(fd, &file_stat);
+#else
+    int fd = fileno(config->input_stat_file);
+    struct stat file_stat;
+    int ret = fstat(fd, &file_stat);
```
1480c1, line 1900: if we don't use the fd variable past this, maybe it might be better to just fold it into the fstat function?

Source/App/EncApp/EbAppConfig.c
```diff
@@ -1887,6 +1888,83 @@ static int32_t read_config_file(EbConfig *config, char *config_path, uint32_t in
...
+            return EB_FALSE;
+        }
+    }
+    return config->rc_twopass_stats_in.buf != NULL;
+}
+
+EbErrorType open_two_passes_stats(EbConfig *config, EncodePass pass,
+    const SvtAv1FixedBuf* rc_twopass_stats_in, uint32_t channel_number)
+{
+    if (pass == ENCODE_SINGLE_PASS) {
```
1480c1, line 1919: maybe a switch case might be better to make it obvious the only condition is the `pass` variable?

Source/App/EncApp/EbAppConfig.c
```diff
@@ -1887,6 +1888,83 @@ static int32_t read_config_file(EbConfig *config, char *config_path, uint32_t in
...
+                    channel_number + 1, stats);
+                return EB_ErrorBadParameter;
+            }
+
+        }
+    } else if (pass == ENCODE_FIRST_PASS) {
+        // for combined two passes,
+        // we only ouptut first pass stats when user explicitly set the --stats
+        if (config->stats) {
+            if (!fopen_and_lock(&config->output_stat_file, config->stats, EB_TRUE)) {
```
1480c1, line 1948: maybe combine the two ifs might look better

Source/App/EncApp/EbAppConfig.h
```diff
@@ -175,12 +175,12 @@ typedef struct EbConfig {
     int           pass;
     const char*   stats;
     FILE *        input_stat_file;
+    SvtAv1FixedBuf rc_twopass_stats_in;
     FILE *        output_stat_file;
+    EbBool        rc_firstpass_stats_out;
```
1480c1, line 180: Maybe put these two new ones before rc_twopass_stats_in or after input_pred_struct_filename so we don't interrupt a chain of pointers? An alternative place is at the end of the struct

Source/Lib/Encoder/Codec/EbEntropyCodingProcess.c
```diff
@@ -310,6 +310,8 @@ static EbBool update_entropy_coding_rows(PictureControlSet *pcs_ptr, uint32_t *r
 
     return process_next_row;
 }
+
+#if !TWOPASS_RC
```
1480c1, line 314: You may want to consider just flat out deleting the ifed out code if it is ifed out as I believe we are looking to clean up the excessive macros we have

Source/API/EbSvtAv1Enc.h
xuguangxin: will this conflict with current API, if we search the EbSvtAv1Enc.h, we have no size_t things.

Source/App/EncApp/EbAppConfig.c
xuguangxin: you mean use fd instead of FILE*?

Source/App/EncApp/EbAppConfig.c
xuguangxin: we can't do this. we need `config->rc_firstpass_stats_out = EB_TRUE;` line

Source/App/EncApp/EbAppConfig.h
xuguangxin: It's better to group same function together. else user need find many place for 2 pass things. I can move rc_twopass_stats_in after output_stat_file.

Source/Lib/Encoder/Codec/EbEntropyCodingProcess.c
xuguangxin: It's better to delete it with others together.

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -4174,4 +4172,20 @@ void eb_output_recon_buffer_header_destroyer(    EbPtr p)
...
+
+EB_API EbErrorType svt_av1_enc_get_stream_info(EbComponentType *    svt_enc_component,
+                                    uint32_t stream_info_id, void* info)
+{
+    if (stream_info_id >= SVT_AV1_STREAM_INFO_END || stream_info_id < SVT_AV1_STREAM_INFO_START) {
+        return EB_ErrorBadParameter;
+    }
+    EbEncHandle         *enc_handle = (EbEncHandle*)svt_enc_component->p_component_private;
+    if (stream_info_id == SVT_AV1_STREAM_INFO_FIRST_PASS_STATS_OUT) {
+        EncodeContext*      context = enc_handle->scs_instance_array[0]->encode_context_ptr;
```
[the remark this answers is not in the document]
xuguangxin: We only have one stream info id here, in future it may have multiple. They will use enc_handle too.

Source/Lib/Encoder/Codec/firstpass.c
```diff
@@ -44,14 +44,28 @@
...
                                     uint64_t frame_number) {
+    FirstPassStatsOut* stats_out = &scs_ptr->encode_context_ptr->stats_out;
     eb_block_on_mutex(scs_ptr->encode_context_ptr->stat_file_mutex);
-    int32_t fseek_return_value = fseek(scs_ptr->static_config.output_stat_file,
-                                       (long)frame_number * sizeof(FIRSTPASS_STATS),
-                                       SEEK_SET);
-    if (fseek_return_value != 0) SVT_LOG("Error in fseek  returnVal %i\n", fseek_return_value);
-    fwrite(stats, sizeof(FIRSTPASS_STATS), (size_t)1, scs_ptr->static_config.output_stat_file);
+    if (realloc_stats_out(stats_out, frame_number) != EB_ErrorNone) {
+        SVT_LOG("realloc_stats_out failed\n");
```
hassount, line 65: shouldn't this be an SVT_FATAL / SVT_ERROR ?

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -4174,4 +4172,20 @@ void eb_output_recon_buffer_header_destroyer(    EbPtr p)
     EB_FREE(obj->p_buffer);
     EB_FREE(obj);
 }
+
+EB_API EbErrorType svt_av1_enc_get_stream_info(EbComponentType *    svt_enc_component,
```
hassount, line 4176: Can you please add a comment to this API implementation

Source/Lib/Encoder/Codec/firstpass.c
```diff
@@ -44,14 +44,28 @@
 
 #define NCOUNT_INTRA_THRESH 8192
 #define NCOUNT_INTRA_FACTOR 3
+
+static EbErrorType realloc_stats_out(FirstPassStatsOut* out, uint64_t frame_number) {
+    if (frame_number < out->size)
+        return EB_ErrorNone;
+    if (frame_number >= out->capability) {
+        size_t capability = frame_number > 1024 ? frame_number * 3 / 2 : 1024;
```
hassount, line 52: I suggest reducing the size of the buffers to 100 and create a #define with that number so that it can be changed easily

Source/App/EncApp/EbAppConfig.c
1480c1: no, I mean
```c
int ret = fstat(fileno(config->input_stat_file), &file_stat);
```

since we do not use the `fd` variable past using it in fstat

Source/App/EncApp/EbAppConfig.c
1480c1: Oh, I guess didn't see the extra part at the end

hassount:
@xuguangxin this PR is causing a crash in a 2pass scenario when encoding long clips:
 can you please check it out ? take any yuv and encode it with the command line below (it doesn't matter if it's not a 128x128)

-i a.yuv -w 128 -h 128 --passes 2 --preset 3

PR #1476, opened 2020-09-04
Run-time flags cleanup

hassount:
# Description
Clean up run-time flags that are not applicable any more
update users guide and sample.cfg at the same time

# Issue

# Author(s)

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/Config/Sample.cfg
+++ b/Config/Sample.cfg
@@ -36,7 +36,7 @@ Tier                            : 0                         # Tier
 HierarchicalLevels              : 4                         # Set hierarchical levels
 PredStructure                   : 2                         # Set prediction structure( 0: low delay P, 1: low delay B, 2: random access [default] )
 PredStructFile                  : Config/PredStruct_level4.cfg        # Prediction structure file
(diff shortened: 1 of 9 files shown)


1480c1:
@hassount would it be possible to start limiting the switch (only has 0 or 1) variables like `unpin` to something like `EbBool` or just use plain int and instead do something like 

```c
static void set_unpin_execution(const char *value, EbConfig *cfg) {
    cfg->unpin = !!strtoul(value, NULL, 0);
};
```

for certain boolean only variables?

OmranAbdelkafi:
Tested obj-1-fast 8bit for both preset 3 and preset 7 : lossless PR.

OmranAbdelkafi (approved):

hassount:
> @hassount would it be possible to start limiting the switch (only has 0 or 1) variables like `unpin` to something like `EbBool` or just use plain unsigned and instead do something like
> 
> ```c
> static void set_unpin_execution(const char *value, EbConfig *cfg) {
>     cfg->unpin = !!strtoul(value, NULL, 0);
> };
> ```
> 
> for certain boolean only variables?
> 
> not necessarily in this PR, but sometime in the future

probably needs a major round of cleanup

Issue #1428, opened 2020-08-20
c and avx2 mismatch


OmranAbdelkafi:
## Description 
Compared obj-1-fast (8bit and 10bit) using c vs default(AVX2) for M0 and M4

## Sample Command line 
 ./SvtAv1EncApp -enc-mode 0 -w 1920 -h 1080 -bit-depth 8 -fps-num 30000 -fps-denom 1001  -n  60 -q 20  -intra-period  119  -asm c -enable-tpl-la 1  -lad 16 -lp 1  -i  aspen_1080p_60f.yuv  -b  aspen_1080p_60f_Q20.bin

## Branch used : 
Head branch from PR #1418

## Results
Bitstreams mismatch between the two different encodings.

EricLamVCD:
Tested thaloundeskmtg 10bit 360p clip for M5: 
  | Speed Deviation  | BD-Rate   Deviation | Memory   Deviation
-- | -- | -- | --
asm-sse4_1 vs asm-c - M5 - 10bit | 0.07% | 0.00% | 0.01%
asm-sse3 vs asm-c  - M5 - 10bit | 0.00% | 0.00% | 0.00%
asm-avx2 vs asm-c  - M5 - 10bit | 0.06% | 0.00% | 0.02%

EricLamVCD:
Tested thaloundeskmtg 10bit 360p clip: 
  | Speed Deviation  | BD-Rate   Deviation | Memory   Deviation
-- | -- | -- | --
asm-sse4_1 vs asm-c - M5 - 10bit | 58.90% | 0.00% | 0.05%
asm-sse3 vs asm-c  - M5 - 10bit | 0.38% | 0.00% | 0.02%
asm-avx2 vs asm-c  - M5 - 10bit | 1577.44% | 0.01% | 0.24%

ttrigui:
@spawlows can you please look into this. It seems that the mismatch is between AVX2 and C. SSE3 and SSE4 are fine.

spawlows:
I looking on this. On this moment I confirm issue, and this is not related with sad_loop

spawlows:
Fixed #1485

OmranAbdelkafi:
closed with PR#1485

PR #1485, opened 2020-09-09
Fix mismatch C/AVX2 kernel svt_av1_apply_temporal_filter_planewise()

spawlows:
# Description
Calculation double/float in some cases can return different value.

# Issue
#1428 

# Author(s)
Slawomir Pawlowski

# Performance impact

- [x] quality
- [ ] memory
- [x] speed
- [x] 8 bit
- [x] 10 bit
- [ ] N/A

# Test set
- [x] other
- [ ] N/A

# Merge method

### DIFF
--- a/Source/Lib/Encoder/ASM_AVX2/EbTemporalFiltering_AVX2.c
+++ b/Source/Lib/Encoder/ASM_AVX2/EbTemporalFiltering_AVX2.c
@@ -142,11 +142,11 @@ static void apply_temporal_filter_planewise(
 
     uint32_t acc_5x5_sse[BH][BW];
     // Larger noise -> larger filtering weight.
-    const float n_decay           = (float)decay_control * (0.7f + logf((float)sigma + 1.0f));
-    const float n_decay_qr_inv    = 1.0f / (2 * n_decay * n_decay);
-    const float block_balacne_inv = 1.0f / (TF_WINDOW_BLOCK_BALANCE_WEIGHT + 1);
-    const float distance_threshold_inv =
-        1.0f / (float)AOMMAX(context_ptr->min_frame_size * TF_SEARCH_DISTANCE_THRESHOLD, 1);
+    const double n_decay                = (double)decay_control * (0.7 + log1p(sigma));
+    const double n_decay_qr_inv         = 1.0 / (2 * n_decay * n_decay);
+    const double block_balacne_inv      = 1.0 / (TF_WINDOW_BLOCK_BALANCE_WEIGHT + 1);
+    const double distance_threshold_inv = 1.0 /
+        (double)AOMMAX(context_ptr->min_frame_size * TF_SEARCH_DISTANCE_THRESHOLD, 1);
 
     uint16_t *frame_sse =
         (plane == PLANE_TYPE_Y) ? luma_sq_error : chroma_sq_error;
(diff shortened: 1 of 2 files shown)


hassount:
LGTM, Pending review from @ttrigui  and testing results from @OmranAbdelkafi

ttrigui:
My understanding was that we decided to use float for the c code instead of double to match the AVX2 code. Is there a reason why we are also changing the AVX2 functions?

spawlows:
Hi @ttrigui 
I modified AVX2 kernel to use double variables in some places because there are some mathematical tricks like multiply by inverse of digit instead of division(this kind of operation can return different results in some cases when float is used), but changes in C are still needed to match bitstreams because expf()[float] is less expensive than exp()[double].
  
Compared to results that I shared with you previously there is smaller quality drop with similar performance.
 
I also found new optimization that reduce AVX2 implementation time from about 2.1% to 0.9% (please look at line 263 or 540; there was power function that uses float when input was on integers).

To summarize this changes in C and AVX2 produce less quality drop and is faster.

ttrigui:
> Hi @ttrigui
> I modified AVX2 kernel to use double variables in some places because there are some mathematical tricks like multiply by inverse of digit instead of division(this kind of operation can return different results in some cases when float is used), but changes in C are still needed to match bitstreams because expf()[float] is less expensive than exp()[double].
> 
> Compared to results that I shared with you previously there is smaller quality drop with similar performance.
> 
> I also found new optimization that reduce AVX2 implementation time from about 2.1% to 0.9% (please look at line 263 or 540; there was power function that uses float when input was on integers).
> 
> To summarize this changes in C and AVX2 produce less quality drop and is faster.

I see. So to fix the mismatch both C and AVX2 code had to be modified.
I'm glad that you changed the powf. I did profiling and I noticed that it is consuming many cycles.
We will then have to test AVX2 before and after PR to evaluate the impact, C before and after PR to evaluate the impact going from double to float and C vs AVX2 after the PR to make sure we do not have any more mismatches.
@OmranAbdelkafi does the testing make sense to you?

paulsgh:
Tested objective 1 fast 8bit and 10bit:

Note: this PR is not lossless.
**M4 - 10bit: BDRate** = 0.002825%
**M8 - 10bit: BDRate** = 0.003405%

� | Speed Deviation  | BD-Rate   Deviation | Memory Deviation
-- | -- | -- | --
M4 - 8bit| 0.14% | -0.01% | -0.01%
M8 - 8bit| 0.20% | -0.01% | 0.00%
M4 - 10bit| 0.03% | **0.00%*** | -0.01%
M8 - 10bit| 0.11% | **0.00%*** | 0.00%

paulsgh:
Tested objective 1 fast 8bit and 10bit:
Note: this PR is not lossless.

� | Speed Deviation  | BD-Rate   Deviation | Memory Deviation
-- | -- | -- | --
M0 - 8bit| 0.59% | 0.00365% | 0.00%
M0 - 10bit| 0.31% | -0.00275% | -0.01%

hassount:
@ttrigui / @OmranAbdelkafi is the BDR difference is as expected ?

OmranAbdelkafi:
Compared avx12 vs c using obj-1-fast 8bit test-set within presets 4 and 8 : Lossless comparison. @hassount Approved PR on my side.

hassount:
@ttrigui @OmranAbdelkafi thank you both, merging.

PR #1544, opened 2020-10-21
GitLab: add gitlab-ci yaml file

1480c1:
# Description

This currently adds a GitLab ci yaml file with stylecheck, cppcheck, and a gcc 10 job as a base to work off along with caching for cppcheck and ccache.

an example can be seen here https://gitlab.com/1480c1/SVT-AV1/-/pipelines/206029956

It is currently using docker containers as that would allow us to pre-install tools without needing to have the install step run on every job. The main issues I currently see is that I am not yet sure what should be done for the windows job since they do not offer container support for windows (maybe it might be okay to use a docker container with mingw-w64, but they do offer windows runners now, so I might try using that). For the macOS jobs, it seems we will need to install their runner on our current self-hosted macOS machines and add them to the repo.

Currently, those container's Dockerfile and actual containers are stored at https://gitlab.com/randomrandomderp/svt-av1-containers, I am currently debating if it would be better to request another repo under AOMediaCodec for storing CI containers or try to see if it would be possible to merge and maintain those files under the main SVT-AV1 repo, however, both seem hard to do for me at this moment.

@luigi311 @ePirat, would it be possible to ask you guys to review this, as you guys are the ones I mainly know that has some familiarity with gitlab's ci?

/cc @tmatth 

# Issue

https://github.com/AOMediaCodec/SVT-AV1/issues/1531

# Author(s)
@1480c1

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method


1480c1:
Going to convert to draft while I'm still working on this

tmatth:
> It is currently using docker containers as that would allow us to pre-install tools without needing to have the install step run on every job. The main issues I currently see is that I am not yet sure what should be done for the windows job since they do not offer container support for windows (maybe it might be okay to use a docker container with mingw-w64, but they do offer windows runners now, so I might try using that). 

FYI, for VLC our CI is doing windows builds with docker+mingw, see 
https://code.videolan.org/videolan/vlc/-/blob/master/extras/ci/gitlab-ci.yml

I suspect that it is the fastest option.

> For the macOS jobs, it seems we will need to install their runner on our current self-hosted macOS machines and add them to the repo.

Yes.

> 
> Currently, those container's Dockerfile and actual containers are stored at https://gitlab.com/randomrandomderp/svt-av1-containers, I am currently debating if it would be better to request another repo under AOMediaCodec for storing CI containers or try to see if it would be possible to merge and maintain those files under the main SVT-AV1 repo, however, both seem hard to do for me at this moment.

The current approach seems ok to me, or migrating that repo to AOMediaCodec. I don't see much benefit of having the Dockerfiles in the SVT-AV1 repo if they are already working externally. 

We keep ours separate as well:
https://code.videolan.org/videolan/docker-images

1480c1:
Currently working on adding general dockerfiles with tools and compilers https://gitlab.com/AOMediaCodec/ci-resources/-/merge_requests/1

1480c1:
@hassount I think this is as far as I think this will be ready for this week

1480c1:
it currently looks like https://gitlab.com/1480c1/SVT-AV1/-/pipelines/213055966

hassount:
@1480c1 can we rebase this PR to make sure all is good now ?

EwoutH:
@1480c1 Thanks for the huge effort! A few comments, mainly from a resource/efficiency standpoint:
 - I like the staging, it's something that's absent from our current setup.
 - A fast-failure option would be very nice, unfortunatly that isn't implemented in Gitlab CI yet (see https://gitlab.com/gitlab-org/gitlab/-/issues/23605)
 - Consider running a subset of tests on pushes to other branches than master. This way, regular pushes get a CI-light, and opened PRs and pushes to master are tested on the full stack. 
   -  I don't think we have to test so many versions of GCC so extensively.
   - Were is the number of 50 shards based on for the end to end tests and the unit tests? Would 5, 10 or 25 also suffice? Or do we need even more?
- Can we let the number of days we store artifacts depend on which kind of event it is? For a regular push, maybe a week is enough. Push to master, a month, and release maybe even indefinitely.
   - Is it necessary to publish the sanitizer artifacts on each push?
   - We could consider not packaging the e2e and unittests in the GCC-10 release artifact, that would save about 60 MB (about half the size) per artifact.

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
+stages:
+  - compile
+  - test
+
+style check:
+  stage: .pre
```
ePirat, line 6: It seems this stage is not listed in `stages`? Is that intended?

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
...
+  image: alpine
+  before_script:
+    - apk add --no-cache git
```
ePirat, line 9: This should probably be added to the docker image to prevent doing this over and over again for every build.

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
...
+          ret=1
+        fi
+        if printf '%s' "$msg" | head -1 | grep -q '\.$'; then
+          echo "Malformed commit message in $i, trailing period in subject line"
+          ret=1
+        fi
+      done <<EOF
+      $(git rev-list HEAD ^FETCH_HEAD)
+      EOF
+      exit ${ret:-0}
```
ePirat, line 48: It would probably be simpler to put this in a .sh file instead of putting it all in the CI yaml

ePirat: That would make it easy to run it locally too to check before pushing.

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
...
+
+cppcheck:
+  stage: compile
+  image: registry.gitlab.com/aomediacodec/aom-testing/ubuntu1804
+  cache:
+    key: cppcheck
+    paths:
+      - .cppcheck
```
ePirat, line 56: Does it make a difference to cache this?

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
...
+    CXXFLAGS: -Werror -Wshadow $MY_CXXFLAGS
+    CCACHE_DIR: $CI_PROJECT_DIR/.ccache
+  cache:
+    key: ${CI_JOB_NAME}
+    paths:
+      - .ccache
+    policy: pull-push
+  script:
+    - ccache -s
+    - eval cmake -GNinja -B Build -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE:=Release} -DBUILD_SHARED_LIBS=OFF $MY_CMAKE_FLAGS
```
ePirat, line 85: Whats the reason for the use of eval here?

ePirat: IMO it would be easier to read to put CMAKE_BUILD_TYPE in `variables` and use it unconditionally here.

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
...
+    - cmake --build Build --parallel 4 --config $CMAKE_BUILD_TYPE --target install
```
ePirat, line 86: Instead of hardcoding `--parallel 4` it should probably be queried from the number of processor cores or threads.

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
...
+
+valgrind compile:
+  extends: .linux-compiler-base
+  image: registry.gitlab.com/aomediacodec/aom-testing/ubuntu1804
+  variables:
+    MY_CFLAGS: -Wno-error -g
+    MY_CXXFLAGS: -Wno-error -g
+    LDFLAGS: -static -static-libgcc
+    MY_CMAKE_FLAGS: -DCMAKE_OUTPUT_DIRECTORY=$PWD/valgrind
+    CMAKE_BUILD_TYPE: Debug
```
ePirat, line 116: Does the use of `MY_CFLAGS` and such actually works with having them expanded in the extended template? As I remember a Gitlab issue about this that it was impossible to do this.

ePirat: Nitpick: `MY_` should probably be changed to `EXTRA_` which would be much clearer imo that those are added to the base ones.

.gitlab-ci.yml
```diff
@@ -0,0 +1,269 @@
...
+    LSAN_OPTIONS: "verbosity=2:color=always:log_threads=1:report_objects=1"
+    ASAN_OPTIONS: "verbosity=2:color=always:print_cmdline=1:strict_string_checks=1"
+    MSAN_OPTIONS: "verbosity=2:color=always"
+    TSAN_OPTIONS: "verbosity=2:color=always:suppressions=./.github/workflows/sanitizers-known-warnings.txt"
+  script:
+    - |
+      command="$PWD/$SANITIZER/SvtAv1EncApp -i Chimera-Aerial_480x264_2997fps_10bit_420_150frames.y4m ${svt_asm:+--asm $svt_asm} -n 60 --preset 8 -b output.ivf"
+      case $SANITIZER in
+      address|memory) $command ;;
+      *) gdb -q -ex "handle SIG32 nostop" -ex r -ex bt -ex q --return-child-result --args $command ;;
```
ePirat, line 223: Do you know the reason for doing this for the thread sanitizer?

ePirat:
Hi, sorry for the late reply! In general this looks good, made some comments inline about some things that were unclear to me or that I think could be improved.

Why does this PR contains a number of changes outside the gitlab-ci.yaml?

.gitlab-ci.yml
1480c1: https://docs.gitlab.com/ee/ci/yaml/#pre-and-post

gitlab has a `.pre` and `.post` stage available regardless of `stages`, I did not list them in the `stages` in case someone decides to try to move them around

.gitlab-ci.yml
1480c1: I guess it could, I will try adding a simple alpine image with git preloaded to the aom-testing repo

.gitlab-ci.yml
1480c1: I guess we could start putting these in a script under the `test` folder

.gitlab-ci.yml
ePirat: 👍 Oh great, forgot about this "new" feature! Absolutely fine then.

.gitlab-ci.yml
1480c1: iirc, it made a substantial difference, (`8 minutes 50 seconds` -> `2 minutes 9 seconds`)

.gitlab-ci.yml
1480c1: the use of `eval` is so the `MY_CMAKE_FLAGS` variable get's evaluated twice as based on some rough testing, if I used something like
```yaml
    MY_CMAKE_FLAGS: -DCMAKE_OUTPUT_DIRECTORY=$PWD/valgrind
```
it would instead put it in the `/valgrind` folder instead of evaling the `$PWD`, at least if I remember correctly on that

.gitlab-ci.yml
1480c1: I can probably remove the `--parallel` flag entirely and let ninja decide

.gitlab-ci.yml
1480c1: `EXTRA_` does sound better, for the part about expanding, refer to the eval comment

.gitlab-ci.yml
ePirat: Oh thats great, for whatever reason last time I tried it seemed to make no difference with cppcheck but maybe I was just doing something wrong...

.gitlab-ci.yml
1480c1: I think the original reason was to try to debug what was causing https://gitlab.com/1480c1/SVT-AV1/-/jobs/807626714, with gdb stopping on SIG32, but after adding gdb and the handle part, it seemed to work fine https://gitlab.com/1480c1/SVT-AV1/-/jobs/846471684

After some skimming, I thought it was due to docker lacking privileges or ptrace caps, but I honestly still don't know why.

I excluded-address and memory since it seems they cannot run under gdb

1480c1:
> Why does this PR contains a number of changes outside the gitlab-ci.yaml?

Those changes are to make it so we can compile the unit tests statically so I can compile the gcc 10 job artifacts statically along with the unit tests and upload the artifacts in case I needed to test them locally on machines that might not have the same shared libraries available (in case we decide to use a different container just for testing the run time for example). I do want to eventually remove the `-Wno-error` on the gcc 10 case but due to the unit tests containing multiple warnings, it seems it will be a while before we can do that.

.gitlab-ci.yml
ePirat: Ah indeed, interesting trick to get this working!

1480c1:
> * I like the staging, it's something that's absent from our current setup.
> * A fast-failure option would be very nice, unfortunatly that isn't implemented in Gitlab CI yet (see https://gitlab.com/gitlab-org/gitlab/-/issues/23605)

This is definitely a feature I was looking for, but noticed was missing

> * Consider running a subset of tests on pushes to other branches than master. This way, regular pushes get a CI-light, and opened PRs and pushes to master are tested on the full stack.

I do not think I yet have enough familiarity with gitlab's yaml to be able to do this just yet, but I definitely will try to see if I can do it

>   * I don't think we have to test so many versions of GCC so extensively.

I personally agree with this, but for the first sets of tests, we decided to go ahead and try to mimic what we do in github actions for the compile and unit tests before removing or adding too many things. I personally think we really only need to test our lowest supported version, the default on a very common distro, and the highest available, but that can be changed in time.

>   * Were is the number of 50 shards based on for the end to end tests and the unit tests? Would 5, 10 or 25 also suffice? Or do we need even more?

The number 50 was set because that's the max gitlab supports for the `parallel:` keyword and I choose the max because using the current number `8` make the unit tests take a long time as I think they are using single core vm's as compared to github's 2 core vm's.

Plus, based on what I understand from https://docs.gitlab.com/ee/user/gitlab_com/index.html#gitlab-cicd, we can have up to 500 jobs in one pipeline (correct me if I'm wrong)

> * Can we let the number of days we store artifacts depend on which kind of event it is? For a regular push, maybe a week is enough. Push to master, a month, and release maybe even indefinitely.

I still haven't found a good way to use variables inside the yaml config as we did with github actions' `${{ env.var }}` syntax, but I will try to dig around more to see if it's possible to do something similar. It would be nice to be able to do something like this for the release uploading jobs

>   * Is it necessary to publish the sanitizer artifacts on each push?

If you are talking about the compiled sanitizer-enabled binaries, I think yes as that seems to be the main way to transfer data between jobs/stages, the fact that you can download it seems to be a side-effect at worst. I do think I can set their expiration to as low as possible though since we theoretically only need them in the same pipeline.

>   * We could consider not packaging the e2e and unittests in the GCC-10 release artifact, that would save about 60 MB (about half the size) per artifact.

Maybe I could split off the gcc 10 job into building and uploading normal binaries as artifacts that we would then use for normal testing and for others to quickly try and another job solely for the purpose of building the unit tests, as I am not aware if you can "label" the artifacts uploaded in a single job

.gitlab-ci.yml
1480c1: Seem's they decide to add a lot of "new" features that not even yaml verifiers can catch up https://github.com/SchemaStore/schemastore/issues/1341

.gitlab-ci.yml
1480c1: I just think it might be the difference between running cppcheck in a single-core VM and on a machine that has more liberties in terms of processing power

1480c1: Also, maybe because you didn't create the directory before running it? I noticed that cppcheck itself will not create it, but will use it if it can (aka if the directory exists and you have permission)

.gitlab-ci.yml
luigi311: If you are referring too variables on extended stuff overwritting the extended template you can get around that using anchors. I had to do that on my gitlab ci for svt-av1 here https://gitlab.com/Luigi311/encoders-docker/-/blob/master/.gitlab-ci.yml#L137

.gitlab-ci.yml
luigi311: Haha this reminded me when we did the same thing for github actions and then ended up just reverting it back to the ci file.

luigi311:
> Consider running a subset of tests on pushes to other branches than master. This way, regular pushes get a CI-light, and opened PRs and pushes to master are tested on the full stack.
I do not think I yet have enough familiarity with gitlab's yaml to be able to do this just yet, but I definitely will try to see if I can do it

This could possibly be done using an if statement to check the current branch to see if its the same as the default branch and if it is you can set a variable to 1 such as full_ci=1 and then on all the tests you want to be exclusive to the master branch check if full_ci=1. I am not sure how that logic works on forks though.
https://gitlab.com/Luigi311/encoders-docker/-/blob/master/.gitlab-ci.yml#L18

.gitlab-ci.yml
1480c1: https://gitlab.com/AOMediaCodec/aom-testing/-/merge_requests/3

.gitlab-ci.yml
1480c1: Moved to a new file called `test/stylecheck.sh`

.gitlab-ci.yml
1480c1: btw @luigi311 I'm pretty sure those global variables are inherited by default unless you specify the same variable inside the jobs

1480c1:
@ePirat @luigi311 @Ewouth currently looks like https://gitlab.com/1480c1/SVT-AV1/-/pipelines/216950015

right now other things to consider are

- `Consider running a subset of tests on pushes to other branches than master. This way, regular pushes get a CI-light, and opened PRs and pushes to master are tested on the full stack.`
- potentially reducing the number of compile tests we do

Source/App/EncApp/EbAppConfig.c
```diff
@@ -1389,7 +1389,7 @@ ConfigEntry config_entry[] = {
  * Constructor
  **********************************/
 EbConfig * svt_config_ctor(EncodePass pass) {
-    EbConfig *config_ptr = (EbConfig *)calloc(1, sizeof(EbConfig));
+    EbConfig *config_ptr = (EbConfig *)calloc(1, sizeof(*config_ptr));
```
hassount, line 1392: why do we need to do this change ? infact I believe that the need here is to be of size EbConfig

1480c1: no, this was there when I was trying to debug an issue, reverted locally

PR #1586, opened 2020-11-19
Fix a few issues with the stylecheck script

ePirat:
# Description
In interactive use (with a tty) git would use a pager for the grep output, causing the script to not work as expected locally.
Additionally the check for git was done after it was used, making it not very useful. For the EOF check, another test condition is added that will prevent a hard failure of the script when git diff would return a path to a folder or empty string.

# Issue

Might help with the failure mentioned at https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/1#note_451236568

# Author(s)
Marvin Scholz

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/test/stylecheck.sh
+++ b/test/stylecheck.sh
@@ -8,13 +8,13 @@ if ! type git > /dev/null 2>&1; then
 fi
 
 echo "Checking for tabs" >&2
-git grep -InP --heading "\t" -- . ':!third_party/**/*' && ret=1
+git --no-pager grep -InP --heading "\t" -- . ':!third_party/**/*' && ret=1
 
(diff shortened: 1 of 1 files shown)


1480c1:
Maybe we should also export `GIT_TERMINAL_PROMPT=0` to make git not even consider a terminal

1480c1:
nice to have

hassount:
> nice to have

Alright, I ll merge this and then we can have that in another PR

PR #1594, opened 2020-11-23
EbAppConfig: Rewrite find_token to fix assertion

ePirat:
# Description
The way find_token was written would lead to it unconditionally
using the next argument following the token. In cases where the
token is the last argument, this could lead to it passing NULL
to strcpy_s which is invalid, triggering an assertion on Windows
debug builds.

This was not reproducible on other platforms as the replacement
strcpy_s implementations does not seem to follow the Windows
implementation for that and only returns an error for these
cases, which is ignored here.

# Issue

# Author(s)
Marvin Scholz

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/Source/App/EncApp/EbAppConfig.c
+++ b/Source/App/EncApp/EbAppConfig.c
@@ -1624,19 +1624,41 @@ static void parse_config_file(EbConfig *config, char *buffer, int32_t size) {
     return;
 }
 
-/******************************************
-* Find Token
-******************************************/
+/**
+ * @brief Find token and its argument
+ * @param argc      Argument count
+ * @param argv      Argument array
+ * @param token     The token to look for
+ * @param configStr Output buffer to write the argument to or NULL
+ * @return 0 if found, non-zero otherwise
+ *
+ * @note The configStr buffer must be at least
+ *       COMMAND_LINE_MAX_SIZE bytes big.
+ *       The argv must contain an additional NULL
+ *       element to terminate it, so that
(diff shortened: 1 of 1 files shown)


Source/App/EncApp/EbAppConfig.c
```diff
@@ -1624,19 +1624,41 @@ static void parse_config_file(EbConfig *config, char *buffer, int32_t size) {
...
+ * @param configStr Output buffer to write the argument to or NULL
+ * @return 0 if found, non-zero otherwise
+ *
+ * @note The configStr buffer must be at least
+ *       COMMAND_LINE_MAX_SIZE bytes big.
+ *       The argv must contain an additional NULL
+ *       element to terminate it, so that
+ *       argv[argc] == NULL.
+ */
 static int32_t find_token(int32_t argc, char *const argv[], char const *token, char *configStr) {
```
1480c1, line 1630: In the case where the only outcome is -1 or 0, it might be better to just use `int` to signal that we don't need a specific length

ePirat: Yes but that change was out of scope of this PR for me, as this was a pre-existing API choice. Personally I think it should be changed to a boolean (as it can only express the token is found or not, anyway) and adjust all callers accordingly.

I can do a follow-up PR with such a change, if desired.

Issue #1588, opened 2020-11-21
get_cpu_flags broken on many CPU models


startrekdude:
Currently `get_cpu_flags` is implemented like:

```c
if (check_4thgen_intel_core_features()) {
    flags |= CPU_FLAGS_MMX | CPU_FLAGS_SSE | CPU_FLAGS_SSE2 | CPU_FLAGS_SSE3 | CPU_FLAGS_SSSE3 |
             CPU_FLAGS_SSE4_1 | CPU_FLAGS_SSE4_2 | CPU_FLAGS_AVX | CPU_FLAGS_AVX2;
}

if (can_use_intel_avx512()) {
    flags |= CPU_FLAGS_AVX512F | CPU_FLAGS_AVX512DQ | CPU_FLAGS_AVX512CD | CPU_FLAGS_AVX512BW |
              CPU_FLAGS_AVX512VL;
}

return flags;
```

This doesn't always give accurate results because it's possible (and not uncommon) for a CPU to have SSE2, etc, but not necessarily all the way up to AVX2. This causes a very significant performance reduction on certain CPUs.

silverbacknet:
check_4thgen_intel_core_features() *only* checks for full AVX2 support; if that's present then all lower flags are enabled. If not, no flags are enabled at all.

hassount:
Currently, all of our testing and validation resources are using AVX2 capable machines and such is currently stated in the readme file: https://github.com/AOMediaCodec/SVT-AV1#hardware 

That said, this item is open to contributions and we can try to help as much as possible in validating and reviewing the fixes.

startrekdude:
Fixed with #1600

PR #1596, opened 2020-11-23
Use cpu_features to get cpu instruction set extension information

startrekdude:
# Description

Fixes `get_cpu_flags` returning incorrect results for processors without AVX2 by replacing custom code with code based on [cpu_features](https://github.com/google/cpu_features). Correctly identifies supported instruction set extensions on every possible processor.

[cpu_features](https://github.com/google/cpu_features) is licensed under the (compatible) Apache license and also uses CMake as a build system.

# Issue

Addresses #1588 

# Author(s)

Sam Haskins
cpu_features by Google

# Performance impact

- [ ] quality
- [ ] memory
- [x] speed
- [ ] 8 bit
- [ ] 10 bit
- [ ] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method


1480c1:
Can you either copy the PIC related stuff from `cpu_features`'s cmake to our main cmake and replace the `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` line or alternativly set `set(CMAKE_POSITION_INDEPENDENT_CODE ON)
` inside `cpu_features`?

startrekdude:
I fixed the PIC issue and the cppcheck failure:
```
Checking /home/runner/work/SVT-AV1/SVT-AV1/Source/Lib/Common/Codec/common_dsp_rtcd.c: ARCH_X86_64=1;_FORTIFY_SOURCE=2...
Error: error: #error "Including cpuinfo_x86.h from a non-x86 target." (preprocessorErrorDirective)
```
is a false positive most likely due to cppcheck not #define-ing the architecture in quite the same way as a real compiler.

Not sure why it's failing every time it needs to go inside FFmpeg, if anyone could shine some light on that it would be very helpful.

Thanks @1480c1 for the -fPIC hint!

ePirat:
Are you sure this is fine, License wise? It seems the cpu_features is licensed as Apache-2.0 License so I am not sure you can just copy their code to SVT-AV1.

startrekdude:
That last commit fixed the rest of the build checks except for the cppcheck false positive I mentioned earlier and the ones that are failing for the other open pull requests too.

Regarding licensing: while I'm not an expert, I believe it should be possible to include cpu_features as a dependency as the [Apache 2.0 License](https://choosealicense.com/licenses/apache-2.0/) allows linking from other code, assuming attribution is given.

ePirat:
The issue is less what the Apache license allows but the fact that integrating this into SVT-AV1 will make part of it licensed under the Apache license, leading to issues like the incompatibility of the Apache 2.0 license with the GPL 2 (https://opensource.stackexchange.com/questions/5664/linking-from-lgpl-2-1-software-to-apache-2-0-library)

hassount:
@startrekdude thank you for the contribution. From previous legal discussions, the Apache 2.0 license does create a conflict license especially when linking libraries to ffmpeg, so we would not be able to use the Apache 2.0 licensed code. Would you be able to use the current functions to derive the information needed ?

startrekdude:
I can do it with the BSD-licensed [cpuinfo](https://github.com/pytorch/cpuinfo). Would that be acceptable?

hassount:
> I can do it with the BSD-licensed [cpuinfo](https://github.com/pytorch/cpuinfo). Would that be acceptable?

that should be ok, but let me get a legal advise just to make sure

hassount:
> > I can do it with the BSD-licensed [cpuinfo](https://github.com/pytorch/cpuinfo). Would that be acceptable?
> 
> that should be ok, but let me get a legal advise just to make sure

Legal has approved the use of the BSD cpuinfo, you can go ahead

startrekdude:
Alright, thanks! Will do.

PR #1611, opened 2020-11-28
Apply clang format

hassount:
# Description
Apply clang format on files that would not impact current dev projects

# Issue

# Author(s)
@hassount 

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method


1480c1:
After we completely run clang-format, should I try to add a job that uses https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/clang-format-diff.py to run something like `git diff -U0 --no-color FETCH_HEAD | python3 <(curl -s https://raw.githubusercontent.com/llvm/llvm-project/master/clang/tools/clang-format/clang-format-diff.py) -p1` to show any needed changes that can also be applied automatically by piping the output to `git apply`

hassount:
> After we completely run clang-format, should I try to add a job that uses https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/clang-format-diff.py to run something like `git diff -U0 --no-color FETCH_HEAD | python3 <(curl -s https://raw.githubusercontent.com/llvm/llvm-project/master/clang/tools/clang-format/clang-format-diff.py) -p1` to show any needed changes that can also be applied automatically by piping the output to `git apply`

sure, that would be great :)

Source/Lib/Encoder/Codec/EbInitialRateControlProcess.c
```diff
@@ -189,17 +194,17 @@ InitialRateControlReorderEntry *determine_picture_offset_in_queue(
...
-                  encode_context_ptr
-                      ->initial_rate_control_reorder_queue
-                          [encode_context_ptr->initial_rate_control_reorder_queue_head_index]
-                      ->picture_number);
+    queue_entry_index = (int32_t)(
+        pcs_ptr->picture_number -
+        encode_context_ptr
+            ->initial_rate_control_reorder_queue
+                [encode_context_ptr->initial_rate_control_reorder_queue_head_index]
+            ->picture_number);
```
chkngit, line 202: What I don't like is why the tool splits encode_context_ptr ->initial_rate_control_reorder_queue  into 2 lines!!
makes reading the code a bit difficult

hassount: I agree, but it is due to the fact that our variable names are too long :/ so the tool needs to split it to be able to fit within the line limit. Infact some of the files such as EbEncHandle.c are untouchable 

we would need to make an effort to reduce the length of our variable names so we avoid having scenarios such as the above.

@chkngit would you like me to revert back this file for now ?

chkngit: how come? even if we keep that variable not split, I see room in the first line to fit both sections? 
if this is just the tool, then dont worry about

hassount: Gotcha, @1480c1 maybe something to revisit we implement the CI

PR #1630, opened 2020-12-07
CMake: only apply fortify source if optimization is applied and not 0

1480c1:
# Description

cmake on centos errors out with `-Werror` since their header contains an error if `-O*` is not specified and fortify source is set

# Issue

# Author(s)
@1480c1

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,7 @@ else()
     if(MINGW)
         list(APPEND flags_to_test -mxsave -fno-asynchronous-unwind-tables)
     else()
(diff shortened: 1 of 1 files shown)


1480c1:
@hassount this PR is needed for moving the gcc 4.8.5 build to centos as centos with `-Werror` will error out since their headers error out if we use fortify source without any optimization

PR #1639, opened 2020-12-13
Only define export headers when building shared library.

tongyuantongyu:
# Description
Export header is used to export specific symbols in compiled shared library, so should only be defined when building shared library. Currently SVT-AV1's API will be exported in downstream project's library which is bad.

# Issue

# Author(s)
@tongyuantongyu

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method


Source/API/EbSvtAv1.h
```diff
@@ -25,11 +25,17 @@ extern "C" {
...
-#else
+#elif defined(__GNUC__) && __GNUC__ >= 4
 #define EB_API __attribute__ ((visibility ("default")))
-#endif
+#else
+#define EB_API
+#endif // if defined(_WIN32)
+#else
+#define EB_API
+#endif // if defined(EB_BUILDING_SHARED_LIBS)
```
1480c1, line 38: Could we move this to a "default path", something like ifndef then define to empty? Since the only two conditions in which `EB_API` should change is already accounted for.

1480c1 (commented):
I think it would be better to wait until https://github.com/AOMediaCodec/libavif/pull/447 is merged before we continue

1480c1 (approved):

PR #1652, opened 2020-12-22
CMake: Add SANITIZER  option and change how flags are added

1480c1:
# Description

Currently supports the Address, Memory, Thread, Undefined, and Integer sanitizer options for clang, however, undefined and integer will have runtime errors, so those might be something worth looking into. Currently doesn't test if the compiler itself accepts the options as trying to coordinate between checking the compilation and linking flags at the same time seems to be a bit too hard for cmake.

Activatable by adding `-DSANITIZER=Address` on the cmake line or through `./Build/linux/build.sh --sanitizer=address`

/cc @ePirat 

# Issue

# Author(s)
@1480c1

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method


CMakeLists.txt
```diff
@@ -79,6 +79,11 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_OUTPUT_DIRECTORY})
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_OUTPUT_DIRECTORY})
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_OUTPUT_DIRECTORY})
 
+if(NOT MSVC)
+    set(SANITIZER "" CACHE STRING "Sanitizer to enable. Currently known values are Address, Memory, Thread, Undefined, and Integer")
+    set_property(CACHE SANITIZER PROPERTY STRINGS Address Memory Thread Undefined Integer)
+endif()
```
ePirat, line 85: It should probably warn or even error using the sanitizer option wit MSVC.

hassount:
@chkngit FYI: 
```
./Build/linux/build.sh --sanitizer=address
```

CMakeLists.txt
```diff
@@ -280,6 +305,35 @@ include_directories(.)
 set(CMAKE_THREAD_PREFER_PTHREADS ON)
 find_package(Threads)
 
+macro(add_both_and_linker_flags)
```
ePirat, line 308: The name of that macro is very confusing. What's "both"?

CMakeLists.txt
```diff
@@ -280,6 +305,35 @@ include_directories(.)
...
+elseif(SANITIZER STREQUAL "memory")
+    add_both_and_linker_flags(-fsanitize=memory -fsanitize-memory-track-origins)
+elseif(SANITIZER STREQUAL "thread")
+    add_both_and_linker_flags(-fsanitize=thread)
+elseif(SANITIZER STREQUAL "undefined")
+    add_both_and_linker_flags(-fsanitize=undefined)
+elseif(SANITIZER STREQUAL "integer")
+    add_both_and_linker_flags(-fsanitize=integer)
+elseif(DEFINED SANITIZER)
+    message(WARNING "Unknown sanitizer: ${SANITIZER}")
```
ePirat, line 330: IMO this should be an error.

CMakeLists.txt
1480c1: I wasn't sure if I wanted to expand `both` to `c_cxx`

ePirat: IMO add_clike_and_ld_flags would make it a bit more obvious to understand what it does

1480c1: will switch to that, thanks

CMakeLists.txt
ePirat: Given that some sanitizer probably need additional flags to work it might be best to only support a definitive list of sanitizers. If someone needs it they can always not use that and manually set the needed flags.

1480c1:
@hassount please review this one when you have time

PR #1666, opened 2021-01-08
Add metadata OBU encoding support

MaxValenzuela:
# Description
This is a first effort to add metadata OBU encoding support.

I have previously contributed a similar API for libaom (commit 1) so users can feel familiar with the usage to attach a metadata stream to a to-be-encoded frame.

Please review this approach as I am not totally familiarized with the packetization process and the last changes in packetization_kernel were to get the metadata OBU in the correct position so it relates to the frame it corresponds inside the stream, but maybe you can provide feedback. Also, had to increase the value in packetization_reorder_entry_ctor from 16 to 160 to be able to write the metadata as 16 wasn't enough. This number is arbitrary.

Testing this commit with ffmpeg generates a stream with metadata OBUs in same order and format of what we get from libaom.
The metadata writing calls right now are in the interest of adding HDR10 (MDCV/CLL) /HDR10+ (ITU-T.35) metadata.

Let me know any suggestion to improve this contribution so it follows your expectation the best.

Also, any suggestion on appropriate tests to create or run for this are very welcome

# Issue

# Author(s)
@MaxValenzuela

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x ] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/Source/API/EbSvtAv1Metadata.h
+++ b/Source/API/EbSvtAv1Metadata.h
@@ -1,0 +1,118 @@
+/*
+* Copyright (c) 2021, Alliance for Open Media. All rights reserved
+*
+* This source code is subject to the terms of the BSD 2 Clause License and
+* the Alliance for Open Media Patent License 1.0. If the BSD 2 Clause License
+* was not distributed with this source code in the LICENSE file, you can
+* obtain it at https://www.aomedia.org/license/software-license. If the Alliance for Open
+* Media Patent License 1.0 was not distributed with this source code in the
+* PATENTS file, you can obtain it at https://www.aomedia.org/license/patent-license.
+*/
+
+#ifndef EbSvtAv1Metadata_h
+#define EbSvtAv1Metadata_h
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+#include <stdio.h>
+#include <string.h> //for 'memcpy'
+#include "EbSvtAv1.h"
+
+struct EbBufferHeaderType;
+
+typedef enum EbAv1MetadataType {
+    EB_AV1_METADATA_TYPE_AOM_RESERVED_0 = 0,
+    EB_AV1_METADATA_TYPE_HDR_CLL        = 1,
+    EB_AV1_METADATA_TYPE_HDR_MDCV       = 2,
+    EB_AV1_METADATA_TYPE_SCALABILITY    = 3,
+    EB_AV1_METADATA_TYPE_ITUT_T35       = 4,
+    EB_AV1_METADATA_TYPE_TIMECODE       = 5,
+} EbAv1MetadataType;
+
+/*!\brief Metadata payload. */
+typedef struct SvtMetadata {
+    uint32_t type; /**< Metadata type */
+    uint8_t *payload; /**< Metadata payload data */
+    size_t   sz; /**< Metadata payload size */
+} SvtMetadataT;
+
+/*!\brief Array of aom_metadata structs for an image. */
+typedef struct SvtMetadataArray {
+    size_t sz;                     /* Number of metadata structs in the list */
+    SvtMetadataT **metadata_array; /* Array of metadata structs */
+} SvtMetadataArrayT;
+
+/*!\brief Allocate memory for SvtMetadataT struct.
+ *
+ * Allocates storage for the metadata payload, sets its type and copies the
+ * payload data into the SvtMetadataT struct. A metadata payload buffer of size
+ * sz is allocated and sz bytes are copied from data into the payload buffer.
+ *
+ * \param[in]    type         Metadata type
+ * \param[in]    data         Metadata data pointer
+ * \param[in]    sz           Metadata size
+ *
+ * \return Returns the newly allocated SvtMetadataT struct. If data is NULL,
+ * sz is 0, or memory allocation fails, it returns NULL.
+ */
+EB_API SvtMetadataT *svt_metadata_alloc(const uint32_t type, const uint8_t *data, size_t sz);
+
+/*!\brief Free metadata struct.
+ *
+ * Free metadata struct and its buffer.
+ *
+ * \param[in]    ptr       Metadata struct pointer
+ */
+EB_API void svt_metadata_free(void *ptr);
+
+/*!\brief Alloc memory for SvtMetadataArrayT struct.
+ *
+ * Allocate memory for SvtMetadataArrayT struct.
+ * If sz is 0 the SvtMetadataArrayT struct's internal buffer list will be
+ * NULL, but the SvtMetadataArrayT struct itself will still be allocated.
+ * Returns a pointer to the allocated struct or NULL on failure.
+ *
+ * \param[in]    sz       Size of internal metadata list buffer
+ */
+EB_API SvtMetadataArrayT *svt_metadata_array_alloc(const size_t sz);
+
+/*!\brief Free metadata array struct.
+ *
+ * Free metadata array struct and all metadata structs inside.
+ *
+ * \param[in]    arr       Metadata array struct pointer
+ */
+EB_API void svt_metadata_array_free(void *arr);
+
+/*!\brief Add metadata to image.
+ *
+ * Adds metadata to EbBufferHeaderType.
+ * Function makes a copy of the provided data parameter.
+ * Metadata insertion point is controlled by insert_flag.
+ *
+ * \param[in]    buffer       Buffer descriptor
+ * \param[in]    type         Metadata type
+ * \param[in]    data         Metadata contents
+ * \param[in]    sz           Metadata contents size
+ *
+ * \return Returns 0 on success. If buffer or data is NULL, sz is 0, or memory
+ * allocation fails, it returns -1.
+ */
+EB_API int svt_add_metadata(struct EbBufferHeaderType *buffer, const uint32_t type, const uint8_t *data, const size_t sz);
+
+/*!\brief Return metadata size.
+ *
+ * Returns the metadata size of the selected metadata given by type
+ *
+ * \param[in]    metadata       Metadata array struct pointer
+ * \param[in]    type           Metadata type descriptor
+ */
+EB_API size_t svt_metadata_size(SvtMetadataArrayT *metadata, const EbAv1MetadataType type);
+
+#ifdef __cplusplus
+}
+#endif // __cplusplus
+
+#endif // EbSvtAv1Metadata_h
--- a/Source/Lib/Encoder/Codec/EbEntropyCoding.c
+++ b/Source/Lib/Encoder/Codec/EbEntropyCoding.c
@@ -4113,6 +4113,22 @@ static void add_trailing_bits(struct AomWriteBitBuffer *wb) {
         svt_aom_wb_write_bit(wb, 1);
     }
 }
+
+// writes the type and payload of the provided metadata to the address dst as a metadata OBU
+static uint32_t write_obu_metadata(SvtMetadataT *metadata, uint8_t *const dst) {
+    if (!metadata || !metadata->payload)
+        return 0;
+    struct AomWriteBitBuffer wb   = {dst, 0};
+    uint32_t                 size = 0;
+    svt_aom_wb_write_literal(&wb, metadata->type, 8);
+    for (size_t i = 0; i < metadata->sz; ++i) {
+        svt_aom_wb_write_literal(&wb, metadata->payload[i], 8);
+    }
+    add_trailing_bits(&wb);
+    size = svt_aom_wb_bytes_written(&wb);
+    return size;
+}
+
 static void write_bitstream_level(BitstreamLevel bl, struct AomWriteBitBuffer *wb) {
     uint8_t seq_level_idx = major_minor_to_seq_level_idx(bl);
     assert(is_valid_seq_level_idx(seq_level_idx));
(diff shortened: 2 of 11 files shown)


Source/API/EbSvtAv1Metadata.h
```diff
@@ -0,0 +1,121 @@
...
+struct EbBufferHeaderType;
+
+typedef enum {
+    OBU_METADATA_TYPE_AOM_RESERVED_0 = 0,
+    OBU_METADATA_TYPE_HDR_CLL = 1,
+    OBU_METADATA_TYPE_HDR_MDCV = 2,
+    OBU_METADATA_TYPE_SCALABILITY = 3,
+    OBU_METADATA_TYPE_ITUT_T35 = 4,
+    OBU_METADATA_TYPE_TIMECODE = 5,
+} OBU_METADATA_TYPE;
```
1480c1, line 27: careful with unprefixed or generic prefixed enums since they add to the global namespace and can't be blocked or undone

hassount:
@MaxValenzuela thank you for the contribution. Would you be able to address the build issues of the PR before we dig into reviewing the code ?
Also, can you please sign the appropriate CLA if you / your company has not done so already ? https://github.com/AOMediaCodec/SVT-AV1#contributor-agreement

MaxValenzuela:
@hassount Hi, I've been having some issues running the tests locally, but I'm getting the hang of it. I'll push a new version when the issues are addressed and I can test them out.
As for the CLA, I have previously contributed on behalf of my company to libaom and I see the link takes you to the same AOM one. Let me know if or how I should provide you with proof about the signature.

hassount:
> @hassount Hi, I've been having some issues running the tests locally, but I'm getting the hang of it. I'll push a new version when the issues are addressed and I can test them out.
> As for the CLA, I have previously contributed on behalf of my company to libaom and I see the link takes you to the same AOM one. Let me know if or how I should provide you with proof about the signature.

Thank you @MaxValenzuela no need to sign another CLA.

Source/API/EbSvtAv1Metadata.h
```diff
@@ -0,0 +1,115 @@
+#ifndef EbSvtAv1Metadata_h
```
hassount, line 1: Can you please add a license header to this file and all new files, if most of the code is already being ported from libaom, then please feel free to port the libaom license header.

Source/Lib/Encoder/Codec/EbEntropyCoding.c
```diff
@@ -4113,6 +4113,21 @@ static void add_trailing_bits(struct AomWriteBitBuffer *wb) {
         svt_aom_wb_write_bit(wb, 1);
     }
 }
+
+static uint32_t write_obu_metadata(SvtMetadataT *metadata, uint8_t *const dst) {
```
hassount, line 4117: can you please add a comment for this function to explain its purpose and dependencies ?

Source/Lib/Encoder/Codec/EbEntropyCoding.c
```diff
@@ -4229,6 +4244,41 @@ static uint32_t write_frame_header_obu(SequenceControlSet *     scs_ptr,
     return total_size;
 }
 
+EbErrorType write_metadata_av1(Bitstream *bitstream_ptr, SvtMetadataArrayT *metadata,
```
hassount, line 4247: can you please add a comment for this function to explain its purpose and dependencies ?

Source/Lib/Encoder/Codec/EbPacketizationReorderQueue.c
```diff
@@ -22,6 +22,6 @@ EbErrorType packetization_reorder_entry_ctor(PacketizationReorderEntry* entry_pt
     entry_ptr->dctor          = packetization_reorder_entry_dctor;
     entry_ptr->picture_number = picture_number;
     //16 should enough for show existing frame
-    EB_NEW(entry_ptr->bitstream_ptr, bitstream_ctor, 16);
+    EB_NEW(entry_ptr->bitstream_ptr, bitstream_ctor, 160);
```
hassount, line 25: this might be a significant byte overhead especially for low resolution low bitrate content. how many extra bytes do we actually need and for what metadata? can we  increase the allocation only when needed i.e. test on the option being enabled to start with.

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -3598,6 +3598,9 @@ static void copy_input_buffer(
     dst->qp = src->qp;
     dst->pic_type = src->pic_type;
 
+    // Copy the metadata array
+    dst->metadata = src->metadata;
```
hassount, line 3602: if this is added to the API and the buffer is filled at the application level, then every field needs to be copied separately and not the pointer redirected, because the pointer maybe recycled for the upcoming pictures.

Source/Lib/Encoder/Globals/EbEncHandle.c
```diff
@@ -3644,6 +3647,11 @@ static void copy_output_recon_buffer(
     dst->dts = src->dts;
     dst->flags = src->flags;
     dst->pic_type = src->pic_type;
+
+    // Copy the metadata array
+    dst->metadata = src->metadata;
```
hassount, line 3652: if this is added to the API and the buffer is filled at the application level, then every field needs to be copied separately and not the pointer redirected, because the pointer maybe recycled for the upcoming pictures.

Source/Lib/Encoder/Globals/EbMetadataHandle.c
```diff
@@ -0,0 +1,132 @@
+#include <stdlib.h>
```
hassount, line 1: Thank you for not adding this to EbEncHandle.c :) 
can you please add a license header to this ?

Source/Lib/Encoder/Globals/EbMetadataHandle.c
```diff
@@ -0,0 +1,132 @@
...
+    }
+    buffer->metadata->metadata_array = metadata_array;
+    buffer->metadata->metadata_array[buffer->metadata->sz] = metadata;
+    buffer->metadata->sz++;
+    return 0;
+}
+
+EB_API void print_metadata_array(SvtMetadataArrayT *metadata) {
+    if(!metadata){
+        printf("Metadata Struct is NULL\n");
```
hassount, line 83: can we use macros from EbLog.h in order to unify printfs and be able to mute them when needed ?

Source/Lib/Encoder/Globals/EbMetadataHandle.c
```diff
@@ -0,0 +1,132 @@
...
+            if (current_metadata && current_metadata->payload && current_metadata->type == type) {
+                sz = current_metadata->sz + 1 //obu type
+                    + 1 //trailing byte
+                    + 1 //header size
+                    + 1; //length field size
+            }
+        }
+    }
+    return sz;
+}
```
hassount, line 132: you might need an empty line at the end otherwise the style check would not pass.

Source/Lib/Encoder/Codec/EbPacketizationReorderQueue.c
MaxValenzuela: An idea is that whenever the temporal entry has metadata, we can check the metadata size to see if 16 is enough to hold it and realloc to add the metadata size if needed.
The case where we need something something bigger is for dynamic tone mapping metadata (p.e. HDR10+) where every frame has its own related metadata.  
This check would be needed everytime dynamic metadata is about to be encoded since we can't predict the exact size or which frames will include it, if we want to keep it to the minimum.

Source/Lib/Encoder/Globals/EbMetadataHandle.c
MaxValenzuela: I think that was left out as it was being used for debugging, we can either remove it or just add the SVT_LOG as you mention if it can be helpful for the user using the API

MaxValenzuela:
Thank you for the review @hassount ! 
There is still an MSVC compiling issue on the checks, should I force push so there is no non-compiling commits on history? I see your guidelines mention to avoid it when addressing comments, but there is still that check issue.

hassount:
> Thank you for the review @hassount !
> There is still an MSVC compiling issue on the checks, should I force push so there is no non-compiling commits on history? I see your guidelines mention to avoid it when addressing comments, but there is still that check issue.

Yes, if you have to rebase there's no way around it :/ we need to have all tests passing, I ll review again once they are.

Source/API/EbSvtAv1Metadata.h
hassount: ping on this

MaxValenzuela:
Since I'm adding a new pointer to the _EbBufferHeaderType_ to contain the metadata arrays, I think any test that makes use of the _EbBufferHeaderType_ struct and doesn't get updated to have the metadata pointer set to NULL in its initialization will be having issues. 

This is the same case for FFmpeg tests, we would need to modify their libsvtav1.c class which calls the SVT encoder with the same changes applied to EbAppProcessCmd.c in the first commit of this PR. 

Let me know if this stage is OK for a review from your side or if I should go further into this (updating many test cases) or what would be a good way to proceed so if this gets accepted in the near future, we can update FFmpeg at the same time to avoid compatibility issues.

Thank you!

Source/Lib/Encoder/Codec/EbPacketizationProcess.c
```diff
@@ -725,12 +738,42 @@ void *packetization_kernel(void *input_ptr) {
...
+
+        if (frm_hdr->show_frame) {
+            // Add HDR10+ dynamic metadata when show frame flag is enabled
+            write_metadata_av1(pcs_ptr->bitstream_ptr,
+                               pcs_ptr->parent_pcs_ptr->input_ptr->metadata,
+                               EB_AV1_METADATA_TYPE_ITUT_T35);
+            svt_metadata_array_free(&pcs_ptr->parent_pcs_ptr->input_ptr->metadata);
+        } else {
+            // Copy metadata pointer to the queue entry related to current frame number
+            uint64_t                        currentPictureNumber = pcs_ptr->picture_number;
```
hassount, line 763: currentPictureNumber should be snake_case

Source/Lib/Encoder/Codec/EbPacketizationProcess.c
```diff
@@ -740,9 +783,26 @@ void *packetization_kernel(void *input_ptr) {
                     output_stream_ptr);
 
         if (pcs_ptr->parent_pcs_ptr->has_show_existing) {
+            uint64_t                        nextPictureNumber = pcs_ptr->picture_number + 1;
```
hassount, line 786: snake_case

Source/Lib/Encoder/Codec/EbPacketizationReorderQueue.c
MaxValenzuela: My idea for this is reflected on the last commit of the latest PR, let me know what you think.

Source/Lib/Encoder/Globals/EbEncHandle.c
MaxValenzuela: Added the copy_metadata_buffer() function to attend these.

Source/Lib/Encoder/Globals/EbMetadataHandle.c
MaxValenzuela: Decided to remove this one as it was not important for the current PR, it can be added on a next PR if it's found useful, but was mostly for debugging.

hassount:
> Since I'm adding a new pointer to the _EbBufferHeaderType_ to contain the metadata arrays, I think any test that makes use of the _EbBufferHeaderType_ struct and doesn't get updated to have the metadata pointer set to NULL in its initialization will be having issues.
> 
> This is the same case for FFmpeg tests, we would need to modify their libsvtav1.c class which calls the SVT encoder with the same changes applied to EbAppProcessCmd.c in the first commit of this PR.
> 
> Let me know if this stage is OK for a review from your side or if I should go further into this (updating many test cases) or what would be a good way to proceed so if this gets accepted in the near future, we can update FFmpeg at the same time to avoid compatibility issues.
> 
> Thank you!

Yes please try to wrap this up so that we can merge it before we move the repo to gitlab on Wednesday

MaxValenzuela:
As expected, the only failing test is ffmpeg as their libsvtav1.c would need to be updated to set the metadata pointer to a valid value.
I will push a new version that forces the metadata pointer on the buffer coming from API users to NULL so it is compatible and the test passes.

hassount:
> ll push a new version that forces the metadata pointer on the buffer coming from API users to NULL so it is compatible and the

yes, infact we need to try to make sure it's backward compatible and that proper null checks are done before accepting this.

MaxValenzuela:
> yes, infact we need to try to make sure it's backward compatible and that proper null checks are done before accepting this.

null checks should be working but I'm not sure how we can handle backward compatibility with libsvtav1.c since we are adding a new metadata pointer to EbBufferHeaderType that would be uninitialized.
All that really needs to be done on their side is setting that new metadata pointer member to NULL as I did here on EbAppProcessCmd.c to prevent the pointer to be uninitialized. 

I don't know if you have any recommendation for this or how API changes are usually handled.

silverbacknet:
FFmpeg doesn't has a good external library versioning system, unfortunately. Going forward you can set an SVTAV1_BUILD similar to how X264_BUILD, and refuse to build or work around missing features, but that doesn't help with someone using an older ffmpeg with a newer library. That's not a solvable problem until a functioning API versioning is put in place. The version macros in place in EbSvtAv1.h don't actually call out the API, just the release number, and the API changes separately.

Maybe time to just add that to the header?

hassount:
> he version macros in place in EbSvtAv1.h don't actually call out the API, just the release number, and the API changes separately.

can we check on the library version basically to force the pointer to be Null in the library if <v0.8.8 and it would work otherwise, we just have to do some tracking on when this would go to ffmpeg (and gstreamer potentially) to enable it. basically using the library versioning as the API versioning.

MaxValenzuela:
> can we check on the library version basically to force the pointer to be Null in the library if <v0.8.8 and it would work otherwise, we just have to do some tracking on when this would go to ffmpeg (and gstreamer potentially) to enable it. basically using the library versioning as the API versioning.

Do you mean ffmpeg should check the SVTAV1 version it's using to define if it should use the metadata pointer? Got a little confused there...
I wouldn't understand the usage of SVT checking its own API version to make the decision unless I'm missing something

hassount:
> > can we check on the library version basically to force the pointer to be Null in the library if <v0.8.8 and it would work otherwise, we just have to do some tracking on when this would go to ffmpeg (and gstreamer potentially) to enable it. basically using the library versioning as the API versioning.
> 
> Do you mean ffmpeg should check the SVTAV1 version it's using to define if it should use the metadata pointer? Got a little confused there...
> I wouldn't understand the usage of SVT checking its own API version to make the decision unless I'm missing something

it would be basically shutting off the feature before a certain revision until both ffmpeg and svt-av1 are updated but thinking more about it, we might as well set it to Null until ffmpeg is updated.

Source/Lib/Encoder/Codec/EbPacketizationReorderQueue.c
hassount: sorry I seem to have missed this. anyways you seems to have addressed this already using the realloc. @1480c1 did we have an issue a while back with realloc (I may be not remembering correctly).

1480c1: Yes, we had a cppcheck issue of not checking the return of realloc to see if it was null or not

PR #1687, opened 2021-02-09
CRF: Add --crf option and clarify if running crf or cqp

1480c1:
# Description

Adds `--crf` option which is equal to `--qp x --rc 0 --enable-tpl-la 1` and update the output shown to show CRF instead of CQP

FFmpeg changes will probably take a bit longer

# Issue
Fixes https://github.com/AOMediaCodec/SVT-AV1/issues/1539
Fixes https://github.com/AOMediaCodec/SVT-AV1/issues/1675

# Author(s)
@1480c1

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Test set
- [ ] other
- [x] N/A

# Merge method

### DIFF
--- a/Docs/svt-av1_encoder_user_guide.md
+++ b/Docs/svt-av1_encoder_user_guide.md
@@ -92,7 +92,11 @@ The intra period defines the interval of frames after which you insert an Intra 
 
 `--rc integer` **[Optional]**
 
-This token sets the bitrate control encoding mode [1: Variable Bitrate, 0: Constant QP]. When `--rc` is set to 1, it is best to match the `--lookahead` (lookahead distance described in the next section) parameter to the `--keyint`. When `--rc` is set to 0, a qp value is expected with the use of the `-q` command line option otherwise a default value is assigned (25).
+This token sets the bitrate control encoding mode [1: Variable Bitrate, 0: Constant QP OR Constant Rate Factor]. When `--rc` is set to 1, it is best to match the `--lookahead` (lookahead distance described in the next section) parameter to the `--keyint`.
+
+When `--rc` is set to 0 and `--enable-tpl-la` is set to 0, a qp value is expected with the use of the `-q` or `--qp` command line option.\
+If `--rc` is set to 0 and `--enable-tpl-la` is set to 1, a crf value is expected with the use of the `--crf` command line option.\
+If a qp/crf values is not specified, a default value is assigned (50).
 
 For example, the following command encodes 100 frames of the YUV video sequence into the bin bit stream file. The picture is 1920 luma pixels wide and 1080 pixels high using the `Sample.cfg` configuration. The QP equals 30 and the md5 checksum is not included in the bit stream.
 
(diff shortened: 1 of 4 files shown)


hassount:
thnx @1480c1 can you please update the help menu and the encoder guide to state what the rc 0 option mean under what circumstances ?

Docs/svt-av1_encoder_user_guide.md
```diff
@@ -92,7 +92,11 @@ The intra period defines the interval of frames after which you insert an Intra
 
 `--rc integer` **[Optional]**
 
-This token sets the bitrate control encoding mode [1: Variable Bitrate, 0: Constant QP]. When `--rc` is set to 1, it is best to match the `--lookahead` (lookahead distance described in the next section) parameter to the `--keyint`. When `--rc` is set to 0, a qp value is expected with the use of the `-q` command line option otherwise a default value is assigned (25).
+This token sets the bitrate control encoding mode [1: Variable Bitrate, 0: Constant QP OR Constant Rate Factor]. When `--rc` is set to 1, it is best to match the `--lookahead` (lookahead distance described in the next section) parameter to the `--keyint`.
+
+When `--rc` is set to 0 and `--enable-tpl-la` is set to 0, a qp value is expected with the use of the `-q` or `--qp` command line option.\
+If `--rc` is set to 0 and `--enable-tpl-la` is set to 1, a crf value is expected with the use of the `--crf` command line option.\
+If a qp/crf values is not specified, a default value is assigned (50).
```
hassount, line 99: I think more emphasis of the behavior should be expressed here:
with --rc set to 0, if --crf is used, then enable-tpl-la is forced to 1, however when -q --qp is used, then the encoder will work in a CRF mode if --enable-tpl-la is set to 1 and in a CQP mode (fixed qp offsets regardless of the content) when --enable-tpl-la is set to 0

Docs/svt-av1_encoder_user_guide.md
```diff
@@ -104,7 +108,7 @@ It should be noted that not all the encoder parameters present in the `Sample.cf
 
 Here are some sample encode command lines
 
-#### 1 pass fixed QP at maximum speed from 24fps yuv 1920x1080 input
+#### 1 pass CRF at maximum speed from 24fps yuv 1920x1080 input
 `SvtAv1EncApp -i input.yuv -w 1920 -h 1080 --fps 24 --rc 0 -q 30 --preset 8 -b output.ivf`
```
hassount, line 112: can we change -q to --crf in this case?

Docs/svt-av1_encoder_user_guide.md
```diff
@@ -164,9 +168,10 @@ The encoder parameters present in the `Sample.cfg` file are listed in this table
 | **Configuration file parameter** | **Command line** | **Range** | **Default** | **Description** |
 | --- | --- | --- | --- | --- |
 | **RateControlMode** | --rc | [0 - 2] | 0 | 0 = CQP , 1 = VBR , 2 = CVBR |
-| **QP** | -q | [0 - 63] | 50 | Quantization parameter used when RateControl is set to 0 |
+| **QP** | -q | [0 - 63] | 50 | Quantization parameter used when RateControl is set to 0 and EnableTPLModel is set to 0 |
+| **CRF** | --crf | [0 - 63] | 50 | Rate control parameter used when RateControl is set to 0 and EnableTPLModel is set to 1 |
```
hassount, line 172: and forces EnableTPLModel to be set to 1

Docs/svt-av1_encoder_user_guide.md
```diff
@@ -164,9 +168,10 @@ The encoder parameters present in the `Sample.cfg` file are listed in this table
 | **Configuration file parameter** | **Command line** | **Range** | **Default** | **Description** |
 | --- | --- | --- | --- | --- |
 | **RateControlMode** | --rc | [0 - 2] | 0 | 0 = CQP , 1 = VBR , 2 = CVBR |
-| **QP** | -q | [0 - 63] | 50 | Quantization parameter used when RateControl is set to 0 |
+| **QP** | -q | [0 - 63] | 50 | Quantization parameter used when RateControl is set to 0 and EnableTPLModel is set to 0 |
```
hassount, line 171: we should add that this would also represent the CRF value if --enable-tpl-la is set to 1

1480c1: Should I make it so `--qp` will force `RateControlMode` to 0 and `EnableTPLModel` to 0?

hassount: hmm, but it would break backward compatibility :/ I think we should keep it this way for now and maybe update it a couple of revisions ahead. We could maybe add a little message if someone is using --qp with --enable-tpl-la 1 that they should use --crf as the option would be deprecated moving forward?

1480c1: Unless we add another variable just to signal if the `qp` member was set using `--crf` or `--qp`, I do not think we can reliably print a warning on just having `--qp` and `--enable-tpl-la 1`

